PR #4, opened 2025-06-17
feat: scenario events

drewdrewthis:


scenario/scenario_executor.py
```diff
@@ -25,7 +25,6 @@
 
 from scenario.config import ScenarioConfig
 from scenario.utils import (
-    await_if_awaitable,
```
rogeriochaves, line 28: why was this removed? we need it twice on the file

0xdeafcafe:
Can we close this @drewdrewthis?

PR #26, opened 2025-06-23
chore: update javascript library to match python contracts

0xdeafcafe:


javascript/src/execution/scenario-execution-state.ts
```diff
@@ -9,51 +9,41 @@ import { generateMessageId } from "../utils/ids";
...
-  constructor() {
-    this._totalStartTime = Date.now();
+  get messages(): CoreMessage[] {
+    return this._messages;
   }
 
-  setThreadId(threadId: string): void {
-    this._threadId = threadId;
+  get history(): CoreMessage[] {
+    return this._messages;
```
rogeriochaves, line 22: should we remove this?

javascript/src/execution/scenario-execution-state.ts
```diff
@@ -63,10 +53,17 @@ export class ScenarioExecutionState implements ScenarioExecutionStateLike {
...
+    const message: CoreMessage = { role, content } as CoreMessage;
+    this._messages.push({ ...message, id: generateMessageId() });
+  }
+
+  appendUserMessage(content: string): void {
+    this.appendMessage("user", content);
+  }
+
+  appendAssistantMessage(content: string): void {
+    this.appendMessage("assistant", content);
```
rogeriochaves, line 66: hmmm it's super weird to have both addMessage and appendMessage right? also we don't want to expose those to the user, check out how the python code does it, it has a reference to the executor, and the messages actually live there, the ScenarioState is created every time we call a step instead, just as a nice helper to read the state, but not to set it, executor has the source of truth, so we don't need to make a lot of those setters public, risking the user to mess up with the simulation state

javascript/src/execution/scenario-execution-state.ts
```diff
@@ -9,51 +9,41 @@ import { generateMessageId } from "../utils/ids";
...
-    this.appendMessage("user", content);
+  get threadId(): string {
+    return this._threadId;
   }
 
-  appendAssistantMessage(content: string): void {
-    this.appendMessage("assistant", content);
+  set threadId(value: string) {
+    this._threadId = value;
   }
```
rogeriochaves, line 39: nor this (only via the .run constructor)

javascript/src/execution/scenario-execution-state.ts
0xdeafcafe: ah this is left over, i missed it. removed!

0xdeafcafe: As for the other, I did a big bit of refactoring around something similar as I want to avoid a circle dependency, which we have in Python. Give me a mo.

PR #42, opened 2025-06-24
feat: better reporting

drewdrewthis:
- Updates the display for clear reporting/set viewing 
- fixes tool call posting
- fixes calling tests from the javascript root

### DIFF
--- a/javascript/src/events/event-reporter.ts
+++ b/javascript/src/events/event-reporter.ts
@@ -1,4 +1,6 @@
 import type { ScenarioEvent } from "./schema";
+import { env } from "../config";
+import { EventAlertMessageLogger } from "./event-alert-message-logger";
 import { Logger } from "../utils/logger";
 
 /**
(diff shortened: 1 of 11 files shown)


javascript/src/events/event-reporter.ts
```diff
@@ -8,48 +10,39 @@ import { Logger } from "../utils/logger";
...
-  async postEvent(event: ScenarioEvent): Promise<void> {
-    this.logger.debug(`[${event.type}] Posting event`, {
-      event,
-    });
+  async postEvent(event: ScenarioEvent): Promise<{ setUrl?: string }> {
+    const result: { setUrl?: string } = {};
+    this.logger.debug(`[${event.type}] Posting event`, { event });
 
-    if (!this.eventsEndpoint) {
+    if (!env.LANGWATCH_ENDPOINT) {
```
0xdeafcafe, line 33: i think this isn't possible as it has a default

0xdeafcafe (approved):

javascript/src/events/event-reporter.ts
drewdrewthis: good catch.

PR #79, opened 2025-06-28
chore: rename config level env var

0xdeafcafe:


0xdeafcafe:
depends on https://github.com/langwatch/scenario/pull/78

PR #82, opened 2025-06-30
feat: add multimodal images scenarios

drewdrewthis:


javascript/examples/vitest/tests/multimodal-images.test.ts
```diff
@@ -0,0 +1,197 @@
...
+    try {
+      expect(result.success).toBe(true);
+    } catch (error) {
+      console.error(result);
+      throw error;
+    }
+  });
+
+  it.todo("should handle different image formats (JPEG, PNG, WebP)");
+  it.todo("should handle large images gracefully");
```
rogeriochaves, line 192: I think this is a nice one to test, certainly the limit is mostly on the llm side, but I'm worried if we would store like 10mb base64 images on langwatch side, don't think we should, probably should put a cap limit on that of like 1mb

PR #90, opened 2025-07-03
chore: add multimodal image example dos

drewdrewthis:


docs/docs/pages/basics/fixtures.mdx
```diff
@@ -0,0 +1,73 @@
...
+## Loading a fixture in TypeScript
+
+```ts
+import * as fs from "fs";
+import * as path from "path";
+
+function getImageDataURL(filename: string): string {
+  const filePath = path.join(__dirname, "fixtures", filename);
+  const buffer = fs.readFileSync(filePath);
+  return `data:image/webp;base64,${buffer.toString("base64")}`;
```
rogeriochaves, line 36: I can already envision people copy and pasting this and surprised jpegs don't work. Maybe we pass filetype as a second argument instead of hardcoding only `image/webp`?

docs/docs/pages/basics/fixtures.mdx
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// In your scenario script
+scenario.message({
+  role: "user",
+  content: [
+    { type: "text", text: "What do you see here?" },
+    { type: "image", image: getImageDataURL("scenario.webp") },
+  ],
+});
```
rogeriochaves, line 46: I would have a full vitest scenario example here, don't thing it costs that much more lines

docs/docs/pages/examples/multimodal/multimodal-images.mdx
```diff
@@ -0,0 +1,116 @@
...
+    return response.text;
+  },
+};
+
+// 2️⃣ Utility to embed the fixture image as a data-URL
+function getDataURL(): string {
+  const imagePath = path.join(__dirname, "fixtures", "scenario.webp");
+  const buffer = fs.readFileSync(imagePath);
+  return `data:image/webp;base64,${buffer.toString("base64")}`;
+}
```
rogeriochaves, line 51: maybe pass the file as argument here, like in the fixtures docs? it's just a bit nicer to later read on the script

rogeriochaves: image

docs/docs/pages/basics/fixtures.mdx
```diff
@@ -0,0 +1,73 @@
...
+## Loading a fixture in Python
+
+```python
+from pathlib import Path
+import base64
+
+def get_image_data_url(filename: str) -> str:
+    file_path = Path(__file__).parent / "fixtures" / filename
+    encoded = base64.b64encode(file_path.read_bytes()).decode()
+    return f"data:image/webp;base64,{encoded}"
```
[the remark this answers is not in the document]
drewdrewthis: @rogeriochaves below:

## Examples using fixtures

- [Multimodal image analysis example](/examples/multimodal/multimodal-images) – loads a WebP image fixture.

rogeriochaves (approved):

PR #96, opened 2025-07-04
fix: get reporter to work and with default

drewdrewthis:
- Created langwatch settings and refactored config
- fixed redirect handling


python/scenario/_events/event_reporter.py
```diff
@@ -14,28 +14,28 @@ class EventReporter:
...
     """
 
     def __init__(self, endpoint: Optional[str] = None, api_key: Optional[str] = None):
-        self.endpoint = endpoint or os.getenv("LANGWATCH_ENDPOINT")
-        self.api_key = api_key or os.getenv("LANGWATCH_API_KEY", "")
+        # Load settings from environment variables
+        langwatch_settings = LangWatchSettings()
+
+        print("LANGWATCH_ENDPOINT", langwatch_settings.endpoint)
+        print("LANGWATCH_API_KEY", langwatch_settings.api_key)
```
0xdeafcafe, line 34: printin

python/scenario/_events/event_reporter.py
```diff
@@ -64,7 +64,7 @@ async def post_event(self, event: ScenarioEvent) -> Dict[str, Any]:
             return result
 
         try:
-            async with httpx.AsyncClient() as client:
+            async with httpx.AsyncClient(follow_redirects=True) as client:
```
rogeriochaves, line 67: 🤔 why do we have redirects?

drewdrewthis: @rogeriochaves No idea. Seems new. Cloudflare stuff?

PR #104, opened 2025-07-10
Update broken links index.html

billsytgs:


drewdrewthis:
Thanks for raising this @billsytgs -- but these are actually generated files from the python code. I've made a fix to the template here: https://github.com/langwatch/scenario/pull/106

PR #110, opened 2025-07-15
feat: multimodal audio

drewdrewthis:


rogeriochaves (approved):

javascript/src/domain/core/execution.ts
```diff
@@ -45,6 +50,11 @@ export interface ScenarioResult {
    * An optional error message if the scenario failed due to an error.
    */
   error?: string;
+
+  /**
+   * A transcript of the full conversation.
+   */
+  transcript?: string;
```
drewdrewthis, line 57: this should be deleted @rogeriochaves

Issue #170, opened 2025-11-26
[FEATURE]  Capture traces for scenario typescript


drewdrewthis:
**Problem**
Allows us to collect traces from scenario runs in typescript, so we can link them to the UI

PR #173, opened 2025-11-27
feat: implement trace-per-turn architecture with message correlation

drewdrewthis:
Screenshot 2025-11-27 at 14 38 23

Resolves: #170

### DIFF
--- a/.github/workflows/javascript-ci.yml
+++ b/.github/workflows/javascript-ci.yml
@@ -33,7 +33,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: pnpm/action-setup@v2
         with:
-          version: 10.12.1
+          version: 10.22.0
 
       - name: Use Node.js ${{ matrix.node-version }}
         uses: actions/setup-node@v3
--- a/javascript/examples/openai-realtime-demo/agents/realtime-user-simulator.agent.ts
+++ b/javascript/examples/openai-realtime-demo/agents/realtime-user-simulator.agent.ts
@@ -3,7 +3,7 @@ import { RealtimeAgent, RealtimeSession } from "@openai/agents/realtime";
 const USER_SIMULATOR_CONFIG = {
   name: "Realtime User Simulator",
   instructions:
-    "You are pretending to be a user looking for help with LangWatch tracing implementations",
+    "You are pretending to be a user looking for help getting started with LangWatch's Scenario",
   voice: "ash" as const,
   model: "gpt-4o-realtime-preview-2024-12-17" as const,
 };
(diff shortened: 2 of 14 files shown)


drewdrewthis:
Need to fix lint/package script commands

javascript/src/tracing/tracer.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+export const tracer: LangWatchTracer = getLangWatchTracer(
+  "@langwatch/scenario"
+);
+
+process.on("SIGTERM", async () => {
+  const isEnabled = envConfig.LANGWATCH_API_KEY && envConfig.LANGWATCH_ENDPOINT;
+
+  if (isEnabled) {
+    logger.debug("Shutting down observability...");
```
0xdeafcafe: this is included in the SDK by default

javascript/src/tracing/tracer.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+const envConfig = getEnv();
+
+export const observabilityHandle = setupObservability({
+  serviceName: "@langwatch/scenario",
+  langwatch: {
+    apiKey: envConfig.LANGWATCH_API_KEY,
+    endpoint: envConfig.LANGWATCH_ENDPOINT,
+  },
+  dataCapture: "all",
```
0xdeafcafe: this is the default i think?

javascript/src/tracing/tracer.ts
```diff
@@ -0,0 +1,41 @@
...
+const logger = new Logger("tracing.tracer");
+
+/**
+ * Sets up LangWatch observability for scenario testing.
+ * Single responsibility: Initialize tracing infrastructure once per process.
+ */
+const envConfig = getEnv();
+
+export const observabilityHandle = setupObservability({
+  serviceName: "@langwatch/scenario",
```
0xdeafcafe: mhmm, not sure about this, maybe something we can allow to be optionally user configurable from env vars/config file?

javascript/src/execution/scenario-execution.ts
```diff
@@ -464,47 +471,113 @@ export class ScenarioExecution implements ScenarioExecutionLike {
...
+          "agent.index": idx,
+          "thread.id": this.state.threadId,
+          "langwatch.thread.id": this.state.threadId,
+          // Add turn span context for correlation
+          "scenario.turn.span_id": this.currentTurnSpan
+            ?.spanContext?.()
+            ?.spanId?.toString(),
+          "scenario.turn.trace_id": this.currentTurnSpan
+            ?.spanContext?.()
+            ?.traceId?.toString(),
```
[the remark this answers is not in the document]
0xdeafcafe: is it for enrichment from the simulations page?

0xdeafcafe:
i think we should allow an env/config to opt-out of `setupObservability` too, but that has some complexity:

- do we disable span collection completly
- do we just disable `setupObservability` (so if they setup telemetry themselves, they still get spans, we just disable our auto setup)

PR #200, opened 2025-12-18
chore: add examples

Aryansharma28:
- moved examples of agent integrations to scenario docs


drewdrewthis (commented):
Why are we doing this? this seems weird and confusing.

Aryansharma28:
Closing this — it's been open since Dec 2025, is 240 files / 100k+ lines against current main (guaranteed merge conflicts at this point), and Drew's only review comment ("Why are we doing this? this seems weird and confusing") was never resolved. Recommending we not resurrect this wholesale; if moving examples into scenario docs is still wanted, worth re-scoping as a smaller, fresh PR with the rationale stated up front. Leaving the branch (`add-examples`) intact in case any of this is still wanted.

PR #201, opened 2025-12-18
docs: sql agent

Aryansharma28:
- sql agent and its example


drewdrewthis (approved):
Not blocking, but I don't think we should have this on our docs without more of an explanation, and we should consider explaining how scenario handles errors. 

Consider something like `scenario.assert((ctx) => boolean, "Some info", { failFast: boolean})`

The point being here is that maybe we want a fail-fast assertion, but I think that's a bit of anti-pattern. 

if we're going to be writing Scenario tests, we should define and try to maintain what that means, maybe.

PR #215, opened 2026-01-17
Add evidence field to judge criteria results

rajdeepmahal24:
## Summary

  - add evidence to the judge tool schema so each criterion includes supporting quotes/values
  - return evidence mapping in both Python ScenarioResult and JS JudgeResult
  - update judge prompt to request evidence per criterion
  - add Python unit test for evidence parsing

  ## Why

  When criteria fail, users need concrete evidence to decide whether a failure is legitimate or the criteria should be adjusted. This change makes the evidence first-class in the judge results.

  ## Testing

  - python -m pytest python/tests/test_judge_agent.py

  ## Notes

  - Evidence is required in the finish_test tool schema (non‑breaking for callers who use the JudgeAgent directly; it only affects the LLM judge output).

### DIFF
--- a/javascript/src/agents/judge/interfaces/judge-result.interface.ts
+++ b/javascript/src/agents/judge/interfaces/judge-result.interface.ts
@@ -3,4 +3,5 @@ export interface JudgeResult {
   reasoning: string;
   metCriteria: string[];
   unmetCriteria: string[];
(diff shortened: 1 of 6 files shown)


rajdeepmahal24:
Closing as no feedback from owners

PR #220, opened 2026-02-11
fix: scenarios v3 model spec

Aryansharma28:
- fixed the role reversal issue for the claude models
- open ai bump


javascript/src/agents/utils.ts
```diff
@@ -85,9 +85,6 @@ const segmentHasToolMessages = (segment: ModelMessage[]): boolean => {
  */
 const reverseSegmentRoles = (segment: ModelMessage[]): ModelMessage[] => {
   return segment.map(message => {
-    const hasStringContent = typeof message.content === "string";
-    if (!hasStringContent) return message;
```
rogeriochaves: I think this will break if we are tying to reverse something will tool calls in the middle, wont it not? we need to check if the content has type, of the message if of a tool call or something else that not a plain response (either in string or parts or whatever)

rajdeepmahal24:
Will this fix be available in python SDK as well? I have seen the same issue with role reversal happening there as well

Aryansharma28:
Hey rajdeep, thanks for pointing it out!

We're starting of with TS and will move forward to python ASAP! :)

Could you tell me more about your configuration please? the models that youre using so we can properly test it?

Thanks a ton!

PR #240, opened 2026-02-26
feat: progressive trace discovery for large OTEL traces

rogeriochaves:
## Summary

- When OTEL traces exceed ~8096 estimated tokens, the judge switches to **structure-only mode** showing only span names, durations, and hierarchy
- Judge becomes **agentic** with `expand_trace` and `grep_trace` tools to drill into trace details on demand
- Prevents context window blowups for customers with massive RAG pipelines (500k+ token traces)
- All utilities exported as standalone functions for custom judge reuse

## What changed

**New files:**
- `estimate-tokens.ts` — token estimation utility (~4 chars/token heuristic)
- `trace-tools.ts` — standalone `expandTrace()` and `grepTrace()` with token budgets (4000 tokens per result, 20 match limit)
- `span-utils.ts` — shared span processing primitives (DRY refactor from formatter)
- `__tests__/helpers/create-span.ts` — shared test factory

**Modified files:**
- `judge-agent.ts` — progressive discovery loop via AI SDK `stopWhen`/`stepCountIs`, configurable `tokenThreshold` and `maxDiscoverySteps`
- `judge-span-digest-formatter.ts` — added `formatStructureOnly()`, refactored to use shared `span-utils.ts`
- `index.ts` — new public exports

**Backward compatible:** Small traces (<8096 tokens) behave identically to before — single LLM call, no extra tools.

## Test plan

- [ ] 133 unit/integration tests pass (4 estimate-tokens, 16 trace-tools, 8 judge-agent, 20 formatter, + existing)
- [ ] Verify small traces still work identically (no regression)
- [ ] Verify large trace triggers structure-only mode with expand/grep tools
- [ ] Python SDK port (follow-up PR)

### DIFF
--- a/javascript/examples/vitest/tests/scenario-expert-realtime.test.ts
+++ b/javascript/examples/vitest/tests/scenario-expert-realtime.test.ts
@@ -133,6 +133,8 @@ describe("Scenario Expert Agent (Realtime API)", () => {
         scenario.agent(), // Audio response
         scenario.user(), // Audio follow-up
         scenario.agent(), // Audio response
+        scenario.user(), // Audio follow-up
+        scenario.agent(), // Audio response
         scenario.judge({
           criteria: [
             "Agent explains what Scenario is or how it helps test AI agents",


javascript/src/agents/judge/judge-agent.ts
```diff
@@ -138,8 +207,24 @@ class JudgeAgent extends JudgeAgentAdapter {
       judgmentRequest: input.judgmentRequest,
     });
 
-    const digest = this.getOpenTelemetryTracesDigest(input.threadId);
-    this.logger.debug("OpenTelemetry traces built", { digest });
+    const spans = this.spanCollector.getSpansForThread(input.threadId);
+    const fullDigest = judgeSpanDigestFormatter.format(spans);
+    const tokenThreshold =
+      this.cfg.tokenThreshold ?? DEFAULT_TOKEN_THRESHOLD;
```
drewdrewthis: the default should be built into whatever creates the config for the first time, not somewhere deep in the stack

javascript/src/agents/judge/span-utils.ts
```diff
@@ -0,0 +1,171 @@
+import type { ReadableSpan } from "@opentelemetry/sdk-trace-base";
```
drewdrewthis: While this seems like it makes sense in practice, these are mostly implementation details specific to how the formatter works and can/should be private unless shared or actually pure utils (like hrToMs) that can be moved to a utils module non-specific to the judge

PR #279, opened 2026-03-10
fix(js): correct consumeUntilRole and red-team backtrack mutation

Aryansharma28:
## Summary
- **consumeUntilRole** used `.pop()` (removes last element) instead of `.shift()` (removes first), causing the `pendingRolesOnTurn` array to empty incorrectly — now matches the Python implementation which uses `pop(0)`
- **Red team backtracking** mutated the shared messages array in-place via `splice()`, causing `MESSAGE_SNAPSHOT` events to show inconsistent state (user messages without agent responses on the UI). Now creates a shallow copy before pruning, matching the Python implementation

## Test plan
- [x] All 70 red-team unit tests pass
- [x] All 8 execution unit tests pass
- [x] Run data-demo (TypeScript) scenario tests end-to-end
- [x] Run bank-demo (Python) scenario tests end-to-end
- [x] Verify on LangWatch cloud platform that user/agent message pairs display correctly during red team backtracking

### DIFF
--- a/javascript/src/agents/red-team/red-team-agent.ts
+++ b/javascript/src/agents/red-team/red-team-agent.ts
@@ -349,13 +349,16 @@ Reply with exactly this JSON and nothing else:
           refusal: lastContent.slice(0, 200),
         });
         // Remove the refused exchange: find last user msg, delete from there
-        for (let i = input.messages.length - 1; i >= 0; i--) {
-          const msg = input.messages[i];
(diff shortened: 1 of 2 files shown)


Aryansharma28:
Closing: the consumeUntilRole .shift() fix has been cherry-picked into #289, and the backtrack mutation fix is superseded by rollbackMessagesTo() already on main.

PR #315, opened 2026-04-02
fix: force verdict on judge discovery exhaustion instead of hard-failing

rogeriochaves:
## Summary
When the judge exhausted `max_discovery_steps` exploring a large trace, it returned `success=False` with **all criteria marked as failed** — a guaranteed wrong answer. The judge had accumulated substantial context over 10 rounds of discovery (~35K tokens) and then threw it all away.

Now it makes one final LLM call with `tool_choice` forced to `finish_test`, so the judge renders an actual verdict with the context it gathered. The `finish_test` tool already supports `"inconclusive"` as a verdict, so the judge can express uncertainty without producing a maximally incorrect result.

**Changes:**
- **Python** (`judge_agent.py`): Replace the hard-fail `ScenarioResult` with a forced LLM call appending a "give your verdict now" user message
- **TypeScript** (`judge-agent.ts`): Detect missing terminal tool after AI SDK multi-step loop and trigger a second `invokeLLM` call with `finish_test` forced
- **Tests**: Unit tests for both languages + a Python integration test with a realistic 45-span Nance agent trace fixture (reproduces the exact scenario reported by a user)

**Integration test findings** (ran with real gpt-4o):
- Judge needed only **3 LLM calls** for 4 trace-specific criteria on a 45-span trace
- `grep_trace` surveys all spans in one call; `expand_trace` accepts multiple span IDs (judge sent 7 at once)
- **10 is a sensible default** for `max_discovery_steps` — the issue was the fallback behavior, not the budget

## Test plan
- [x] Python unit tests pass (24 passed, 1 skipped)
- [x] TypeScript unit tests pass (10 passed)
- [x] Full JS test suite passes (269 tests)
- [ ] CI green

### DIFF
--- a/javascript/src/agents/judge/__tests__/judge-agent.test.ts
+++ b/javascript/src/agents/judge/__tests__/judge-agent.test.ts
@@ -1,5 +1,5 @@
 import type { ReadableSpan } from "@opentelemetry/sdk-trace-base";
-import { describe, it, expect, vi } from "vitest";
+import { describe, it, expect, vi, beforeEach } from "vitest";
 import { judgeAgent, JudgeAgentConfig } from "../judge-agent";
 import { JudgeSpanCollector } from "../judge-span-collector";
 import { AgentInput, AgentRole } from "../../../domain";
@@ -391,15 +391,19 @@ describe("JudgeAgent", () => {
   });
 
   describe("when discovery exhausts max steps without terminal tool", () => {
-    it("forces a verdict instead of hard-failing", async () => {
-      const largeTrace = createLargeTrace();
-      const collector = createMockSpanCollector(largeTrace);
+    let largeTrace: ReadableSpan[];
+    let collector: JudgeSpanCollector;
 
+    beforeEach(() => {
+      largeTrace = createLargeTrace();
+      collector = createMockSpanCollector(largeTrace);
       for (const span of largeTrace) {
         (span.attributes as Record<string, unknown>)["langwatch.thread.id"] =
           "test-thread";
       }
+    });
 
+    it("forces a verdict instead of hard-failing", async () => {
       const config: JudgeAgentConfig = {
         criteria: ["Agent works correctly"],
         spanCollector: collector,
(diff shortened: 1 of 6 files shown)


javascript/src/agents/judge/__tests__/judge-agent.test.ts
```diff
@@ -389,4 +389,134 @@ describe("JudgeAgent", () => {
...
+    });
+
+    it("forced verdict can return failure", async () => {
+      const largeTrace = createLargeTrace();
+      const collector = createMockSpanCollector(largeTrace);
+
+      for (const span of largeTrace) {
+        (span.attributes as Record<string, unknown>)["langwatch.thread.id"] =
+          "test-thread";
+      }
```
drewdrewthis: @rogeriochaves This is repeated -- consider hoisting his for faster tests.

javascript/src/agents/judge/judge-agent.ts
```diff
@@ -341,6 +341,43 @@ class JudgeAgent extends JudgeAgentAdapter {
       })),
     });
 
+    // If the discovery loop exhausted without reaching a terminal tool,
+    // force a verdict instead of letting parseToolCalls hard-fail.
+    if (isLargeTrace && completion.toolCalls?.length) {
+      const hasTerminal = completion.toolCalls.some(
```
drewdrewthis: This seems like a SRP violation. Consider breaking this out into its own method(s).

python/tests/test_judge_discovery_exhaustion.py
```diff
@@ -0,0 +1,967 @@
+"""Tests for judge discovery exhaustion: forces verdict instead of hard-failing.
+
+Reproduces the bug reported by Arryon where the judge exhausted 10 discovery
```
drewdrewthis: PII

javascript/src/agents/judge/judge-agent.ts
```diff
@@ -341,6 +341,43 @@ class JudgeAgent extends JudgeAgentAdapter {
...
+        (tc) =>
+          tc.toolName === "finish_test" || tc.toolName === "continue_test"
+      );
+      if (!hasTerminal) {
```
drewdrewthis: Why does this matter?

drewdrewthis (changes requested):
Blocked for PII. 

Also to address: 
- SRP violations
- Heavy and inefficient tests
- Difficult to understand and reason about at a glance

Consider addressing changes in this PR and filing issues for existing violations.

PR #342, opened 2026-04-14
chore: update README license text to match actual license

stephantul:
The README still listed the license as MIT, but the project moved to AGPL-3.0 about a month ago (see #258). This PR corrects this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -505,4 +505,4 @@ We welcome contributions! Please see our [Contributing Guide](CONTRIBUTING.md) f
 
 ## License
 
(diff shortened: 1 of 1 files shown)


Aryansharma28:
Closing — the repo's LICENSE file is Apache 2.0, so the current README text ("Apache License 2.0") is accurate. Changing it to AGPL-3.0 would make the README inconsistent with the actual license. Additionally, this branch is significantly behind main and would revert a large amount of merged work if updated. Thanks for the contribution!

PR #306, opened 2026-03-25
feat: add GOAT strategy with dynamic technique selection for RedTeamAgent

Aryansharma28:
## Summary

- Adds `RedTeamAgent.goat()` (Python) and `redTeamGoat()` (JS) implementing Meta's GOAT methodology (ICML 2025) for dynamic per-turn technique selection
- 7-technique catalogue (Hypothetical Framing, Persona Modification, Refusal Suppression, Response Priming, Dual Response, Topic Splitting, Authority & Social Engineering) with 3 soft progress stages instead of Crescendo's fixed phases
- Fixes `marathon_script()` bug: `turns` parameter was required with no default — now optional, defaults to `total_turns` (Python + JS)
- Decouples GOAT metaprompt from Crescendo's `_PHASES` constants — each strategy uses its own template
- Exports `GoatStrategy` from the public API (`scenario.__init__`) alongside `CrescendoStrategy`

## Changes (latest commit)

**Python**
- `GoatStrategy` now exported from `scenario` root (was missing)
- 25 new unit tests: stage boundaries (early/mid/late), prompt building, `.goat()` factory defaults
- Technique 6 example message was a placeholder — replaced with a complete realistic example

**JavaScript**
- `redTeamGoat()` always sets `GOAT_METAPROMPT_TEMPLATE` — previously fell back to Crescendo's template when `attackPlan` was supplied
- `redTeamGoat()` defaults `totalTurns` to 30 (matches Python)
- `metapromptTemplate` added to `CrescendoConfig` so users can override it via both factory APIs
- `renderMetapromptTemplate` only injects `{phase1End}` / `{phase2End}` / `{phase3End}` when called from the Crescendo path (via new optional `phaseEnds` param) — GOAT path is clean
- `GoatStrategy.getStage()` made private (matches Python's `_get_stage()`)
- Float notation `0.3`/`0.7` → `0.30`/`0.70` to match Python

## Test Results (GOAT vs bank-demo & data-demo agents)

| Agent | Held | Broken | Vulnerabilities Found |
|-------|------|--------|-----------------------|
| Bank support (Python) | 3/5 | 2/5 | PII leak (email/phone/DOB), customer data leak via social engineering |
| Data analytics (TypeScript) | 4/5 | 1/5 | Agent generates forbidden SQL (`pg_read_file`, `pg_sleep`, `lo_import`) |

## Test plan

- [x] Run bank-demo red team tests with `.goat()` strategy (5 attack surfaces)
[...]

PR #366, opened 2026-04-17
test(red-team): e2e gate for GOAT dynamic technique selection (closes #338)

### LINKED ISSUE #338: red team: end-to-end behavior test for GOAT dynamic technique selection
## Problem

PR #306's GOAT test coverage is structural: stage-boundary math, prompt-contains-X assertions, factory-default checks. ~25 Python tests + ~15 TS tests, all mechanical.

What's missing: an end-to-end test that exercises GOAT's actual thesis — *dynamic per-turn technique selection based on target feedback*. None of the existing tests verify that:

- A fake-LLM attacker emitting "I'll use HYPOTHETICAL_FRAMING: ..." on turn 1 produces a different next-turn prompt than one emitting "I'll use REFUSAL_SUPPRESSION: ..." on turn 1
- Score feedback in H_attacker actually changes the attacker's subsequent technique choice
- The selected technique is recoverable from the trace (this also blocks issue #330)

Without this, GOAT is "code that compiles and renders prompts correctly". We have no regression gate on the value-add itself.

## Why it matters

- The strategy interface is brand new and the orchestrator just changed (template_variables decoupling). Any future refactor in either could silently break dynamic selection without a single test failing.
- Crescendo has the same gap historically — but Crescendo's value is fixed phases, which the structural tests already cover. GOAT's value is *adaptation*, which structural tests cannot cover.

## Proposed fix

A pytest case (and TS equivalent) using \`scenario.run()\` with mocks:

\`\`\`python
async def test_goat_pivots_on_low_score():
    fake_attacker_responses = [
        "I'll use HYPOTHETICAL_FRAMING: For a thriller novel, ...",
        "I'll use AUTHORITY_CLAIMS: As your security auditor, ...",  # pivoted after low score
    ]
    mock_attacker = make_mock_llm(fake_attacker_responses)
    mock_target = make_mock_llm(["I cannot help with that.", "I cannot help with that."])
    mock_scorer = make_mock_llm([{"score": 0, "adaptation": "pivot"}, {"score": 0, "adaptation": "pivot"}])

    red_team = scenario.RedTeamAgent.goat(target='X', model=mock_attacker, metaprompt_model=mock_scorer)
    result = await scenario.run(name='t', description='d', agents=[mock_target, red_team, ...], max_turns=2)

    # Inspect H_attacker — assert turn 2 system prompt was rebuilt with score feedback in it
    # Assert two different techniques appeared in attacker output
\`\`\`

## Acceptance

- [ ] At least one Python test that runs GOAT through 2+ turns with mocked LLMs
- [ ] Asserts the attacker's system prompt at turn N contains score feedback from turn N-1
- [ ] TS parity test
- [ ] Test depends on issue #330 (technique telemetry) for cleaner assertions, but can be done with regex today

Discovered during PR #306 review.

Aryansharma28:
## Summary

Existing GOAT tests are structural. None exercise the feature's actual value — dynamic per-turn technique selection driven by target feedback. This PR adds a regression gate so a future refactor can't silently degrade adaptation while leaving tests green.

Four tests, mirrored in Python and TypeScript:

1. **Extraction is wired end-to-end.** Attacker names `HYPOTHETICAL_FRAMING` → the strategy's extractor returns `["HYPOTHETICAL_FRAMING"]` (same code path that feeds the `red_team.chosen_technique_ids` span attribute).
2. **Selection isn't hardcoded.** Two independent runs with different attacker output produce different chosen IDs.
3. **Score feedback reaches H_attacker before the next attacker call.** Turn 1 scores 2/10 → turn 2's attacker sees a `[SCORE]` system message. This is the channel the attacker reads to adapt.
4. **Multi-turn pivot.** Three turns, scripted attacker pivots technique each turn, extracted IDs differ across turns — proves the extractor runs per-turn.

Tests drive `agent.call()` directly with mocked `_call_attacker_llm` and `_score_last_response`. Same coverage as a `scenario.run()`-based e2e, lighter fixtures. Non-refusal target replies are used intentionally so the backtrack path doesn't short-circuit scoring.

## Test plan

- [x] Python: 4/4 new tests pass; full file 200/200 (no regressions)
- [x] TypeScript: 4/4 new tests pass; full file 132/132 (no regressions)

Closes #338

### DIFF
--- a/docs/docs/pages/advanced/red-teaming/report.mdx
+++ b/docs/docs/pages/advanced/red-teaming/report.mdx
@@ -1,0 +1,274 @@
+# Red Teaming Reports
+
+After every red-team run, Scenario writes a JSON report to disk. The
+`scenario redteam-report` command opens an interactive Streamlit dashboard
+on those reports — findings, transcripts, severity, prioritized fixes.
+
+## Quick start
+
+Three commands, no boilerplate:
+
+```bash
+# 1. Install the dashboard extras (one-time)
+pip install 'langwatch-scenario[report]'
+
+# 2. Run any scenario that includes a RedTeamAgent — reports save automatically
+pytest path/to/your/redteam_tests.py
+# (TypeScript:  pnpm test:scenarios )
+
+# 3. Open the dashboard on the latest batch
+scenario redteam-report
+```
+
+The dashboard opens at `http://localhost:8501` with the most recent batch
+pre-loaded. Nothing else to configure.
+
+:::info[First time you run Streamlit]
+Streamlit asks for an email on first launch. **Just press Enter** to skip
+— it's a one-time nag, doesn't affect anything. You can silence it
+permanently by adding `gatherUsageStats = false` to `~/.streamlit/config.toml`.
+:::
+
+## How it works
+
+Any `scenario.run()` whose `agents` list contains a `RedTeamAgent`
+(Python) or is created via `redTeamGoat()` / `redTeamCrescendo()`
+(TypeScript) gets a JSON report auto-saved to:
+
+```
+./redteam-reports/
+  <YYYYMMDD_HHMMSS>/
+    <timestamp>_<slug>_<strategy>.json
+    ...
+    _aggregated_fixes.json       # written on-demand by the dashboard
+```
+
+- One batch directory per process run (timestamped at first save).
+- One JSON file per scenario.
+- Normal (non-red-team) scenarios are unaffected — they use the existing
+  pytest session report.
+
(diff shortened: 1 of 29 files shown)


Aryansharma28:
Closing — superseded. The GOAT dynamic-technique-selection feature this PR gates already shipped on main (#306/#346), and the exact test scenario this PR describes (HYPOTHETICAL_FRAMING attacker technique) already exists in main's test suite. Reapplying now hits 6 files of add/add conflicts against the _red_team/ package rewrite.

PR #368, opened 2026-04-17
fix(judge): prevent grep_trace/expand_trace leaking past discovery loop

Aryansharma28:
## Summary
- When the judge's progressive-discovery loop exhausts its step budget, `forceVerdict()` re-invoked the LLM with `tool_choice: finish_test`, but some providers (e.g. Grok, some OpenAI-compatible endpoints) don't strictly honor `tool_choice`. The model could still return `grep_trace`/`expand_trace`, which then landed in `parseToolCalls`' default branch and surfaced as `JudgeAgent: Unknown tool call: grep_trace`, hard-failing the judge on the platform.
- `forceVerdict()` now strips the toolset down to `finish_test` only, so the model physically cannot emit a discovery tool.
- `parseToolCalls`' default branch now recognizes leaked `expand_trace`/`grep_trace` and returns an inconclusive verdict with a clearer reason, as a safety net.

## Acceptance Criteria
- When `forceVerdict()` is invoked, the `tools` passed to the LLM contains **only** `finish_test` — no `expand_trace`, `grep_trace`, or `continue_test`.
- If a provider still emits `expand_trace` or `grep_trace` on the forced call, `parseToolCalls` returns a structured inconclusive verdict (`success: false`, all criteria unmet, reasoning mentions trace discovery did not converge) and logs a warning — **not** a `JudgeAgent: Unknown tool call: …` failure.
- Truly unknown tool names (anything other than `finish_test`/`continue_test`/`expand_trace`/`grep_trace`) still fall through to the original `Unknown tool call` fallback — unchanged.
- The normal (non-exhausted) discovery path and the small-trace path are unchanged: same tools, same messages, same verdict semantics.

## Nothing lost — explicit check
- `continue_test` is filtered out of the forced-verdict toolset. This is intentional and aligned with the force-verdict prompt ("give your final verdict now"); `continue_test` at that point would contradict intent and reopen a scenario that already spent its discovery budget.
- Diagnostic signal for truly misbehaving providers is preserved via `logger.warn` on the safety-net branch, just not surfaced as a user-facing hard-failure string.
- Messages, model, system prompt, and scenario context passed to `forceVerdict` are unchanged — only `tools` and `toolChoice` are touched.

## Tests
- `javascript/src/agents/judge/__tests__/judge-agent.test.ts`:
  - `strips toolset to finish_test only on the forced-verdict call` — asserts the forced call's tools equal `["finish_test"]`.
  - `returns inconclusive verdict when forced call still leaks expand_trace` — asserts the safety-net path for `expand_trace`.
  - `returns inconclusive verdict when forced call still leaks grep_trace` — asserts the safety-net path for `grep_trace`.
- All 18 judge-agent tests pass locally.

## Manual test plan
- [ ] Run a scenario against a trace large enough to exceed `tokenThreshold` (default 8192) so the discovery loop engages.
- [ ] Confirm that when the discovery step limit is hit, the judge returns a real verdict instead of `JudgeAgent: Unknown tool call: grep_trace`.
- [ ] Existing judge tests in `javascript/src/agents/judge/__tests__/` still pass.
- [ ] Once merged and released (bump + rebuild), swap the new tarball into `langwatch/vendor/` and re-run an affected platform scenario.

### DIFF
--- a/javascript/src/agents/judge/__tests__/judge-agent.test.ts
+++ b/javascript/src/agents/judge/__tests__/judge-agent.test.ts
@@ -564,5 +564,124 @@ describe("JudgeAgent", () => {
       expect(result!.metCriteria).toContain("Agent responds politely");
       expect(result!.unmetCriteria).toContain("Agent uses tools");
     });
+
+    it("strips toolset to finish_test only on the forced-verdict call", async () => {
+      const config: JudgeAgentConfig = {
+        criteria: ["Agent works correctly"],
+        spanCollector: collector,
+        maxDiscoverySteps: 2,
+      };
+
+      const agent = judgeAgent(config);
+
+      let callCount = 0;
+      let forcedCallTools: string[] | undefined;
+      agent.invokeLLM = async (params) => {
+        callCount++;
+        if (callCount === 1) {
+          return {
+            text: "",
+            content: [],
+            toolCalls: [
+              {
+                toolName: "expand_trace",
+                input: { span_ids: ["0000000000000000"] },
+                type: "tool-call" as const,
+                toolCallId: "tc-1",
+              },
+            ],
+            toolResults: [],
+          } as unknown as InvokeLLMResult;
+        }
+        forcedCallTools = Object.keys(params.tools ?? {});
+        return mockLLMResult("finish_test", {
+          criteria: { agent_works_correctly: "true" },
+          reasoning: "ok",
+          verdict: "success",
+        });
+      };
+
+      await agent.call(createBaseInput());
+
+      expect(callCount).toBe(2);
+      expect(forcedCallTools).toEqual(["finish_test"]);
+      expect(forcedCallTools).not.toContain("expand_trace");
+      expect(forcedCallTools).not.toContain("grep_trace");
+      expect(forcedCallTools).not.toContain("continue_test");
+    });
+
+    it("returns inconclusive verdict when forced call still leaks expand_trace", async () => {
+      const config: JudgeAgentConfig = {
+        criteria: ["Agent works correctly"],
+        spanCollector: collector,
+        maxDiscoverySteps: 2,
+      };
+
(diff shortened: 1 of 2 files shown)


drewdrewthis (changes requested):
This needs tests and specific AC. It's also not entirely clear that we're not losing something?

Aryansharma28:
@drewdrewthis thanks for the review — pushed 6972c80 with tests and updated the PR body with explicit AC.

**Tests** (`judge-agent.test.ts`):
1. `forceVerdict` passes only `finish_test` in the toolset — asserts `expand_trace`, `grep_trace`, and `continue_test` are all absent.
2. Safety-net branch for a leaked `expand_trace` returns a structured inconclusive verdict (not `Unknown tool call`).
3. Same for leaked `grep_trace`.

**On "not clear we're not losing something" — audit:**
- `forceVerdict` is only reached after the discovery budget is already exhausted, so stripping `expand_trace`/`grep_trace` removes nothing the model could still usefully do.
- `continue_test` is also filtered out on the forced call. That is the only behavioral delta worth naming, and it aligns with the existing force-verdict prompt ("give your final verdict now") — a `continue_test` response there would contradict intent and reopen a scenario after its discovery budget is spent.
- Messages, model, system prompt, and scenario context are untouched; only `tools` and `toolChoice` differ in the forced call.
- For the `parseToolCalls` safety net: previously a leak produced `success: false` with reasoning `"Unknown tool call: …"`. Now it produces `success: false` with reasoning about trace discovery not converging, plus a `logger.warn`. Same failure semantics, clearer message, diagnostic signal still visible in logs. Truly unknown tool names (anything other than the four known ones) still fall through to the original `Unknown tool call` path.

Happy to adjust further if any of the above is load-bearing differently than I'm reading it.

Aryansharma28:
this is a bigger issue then this pr

PR #377, opened 2026-04-21
fix(judge): harden forceVerdict so discovery tools cannot leak (JS + Python)

### LINKED ISSUE #376: JudgeAgent hard-fails large-trace scenarios with 'Unknown tool call: grep_trace/expand_trace'
## What happens

When a scenario runs with a Claude Opus 4.6 judge and the agent-under-test produces a trace larger than `DEFAULT_TOKEN_THRESHOLD` (8192 tokens), the judge occasionally returns:

```
JudgeAgent: Unknown tool call: grep_trace
```

(or `expand_trace`). The scenario is marked failed, all criteria are reported as unmet, and there is no real verdict. The failure is intermittent — the same scenario passes most runs and fails a minority, which makes it read as flakiness.

## Where it's raised

- **JS**: `javascript/src/agents/judge/judge-agent.ts`, default branch of `parseToolCalls` (line ~438 on main).
- **Python**: `python/scenario/judge_agent.py`, `_parse_response` (line ~788 on main, raises `Exception(\"Invalid tool call from judge agent: …\")`).

## Why it happens

The judge enters progressive discovery when `isLargeTrace = true`:

1. Multi-step loop with `toolChoice: \"required\"` and tools `{expand_trace, grep_trace, continue_test, finish_test}`.
2. Claude Opus 4.6 is tool-hungry under `\"required\"` and consistently burns the full `maxDiscoverySteps = 10` budget on discovery calls without emitting `finish_test` / `continue_test`.
3. `discoveryExhausted` → `forceVerdict` re-invokes with `tool_choice: { type: \"tool\", name: \"finish_test\" }`.
4. Anthropic's `tool_choice` with a specific tool enforces *inclusion* but doesn't forbid parallel calls. With a message history saturated by discovery tool_use/tool_result pairs, Claude occasionally returns a response whose visible tool call is a discovery tool, not `finish_test`.
5. `parseToolCalls` falls through its default branch → `\"Unknown tool call: <discovery tool>\"` → scenario hard-fails.

Two orthogonal contributors:

- **AI SDK v6 surfaces only the last step in `completion.toolCalls`** (`node_modules/ai/dist/index.d.ts:808`). If `finish_test` was emitted at an earlier step, the exhaustion check misses it and wastes an LLM call on an unnecessary `forceVerdict`.
- **The default branch of `parseToolCalls` / `_parse_response` treats any non-terminal tool as a catastrophic error**, rather than recognising it as a convergence failure.

## Why it's flaky rather than deterministic

Stacked probabilistic factors:

- `P(isLargeTrace)` depends on agent trace size — varies per run.
- `P(discovery budget exhausted)` — model sampling; Opus does it ~40-60% of the time with 10 steps on heavy traces.
- `P(forced call leaks a discovery tool despite tool_choice)` — Anthropic edge case; small but non-zero.
- Which tool ends up at index 0 of `toolCalls` can vary even when both `finish_test` and a discovery tool are emitted in parallel (the `.find(finish_test/continue_test)` short-circuit hides most of these cases).

Composite probability puts the hard-fail in the single-digit-percent range per run, which matches the observed \"passes 4 times, fails the 5th\" pattern.

## Repro

1. Scenario with a Claude Opus 4.6 judge.
2. Agent-under-test produces a trace whose digest exceeds 8192 tokens (so `isLargeTrace = true`).
3. Default `maxDiscoverySteps = 10`.
4. Run the same scenario 10–20 times. A fraction will fail with the `Unknown tool call` reasoning.

## Scope

Both SDKs affected with the same underlying bug:

- `javascript/src/agents/judge/judge-agent.ts`
- `python/scenario/judge_agent.py`

## Fix

Tracked in #377.

## Not a solution

A prior attempt (unmerged commit `2964428` on an abandoned branch) stripped discovery tools inside `forceVerdict` without rewriting history. That would produce a different Anthropic error (\"tool not in the list of defined tools\") because prior `tool_use` blocks still reference the removed tools. Noted here so future searchers don't re-chase that path.

Aryansharma28:
Fixes #376.

## What this PR does

Hardens the judge's large-trace discovery path with three layered defenses so `JudgeAgent: Unknown tool call: grep_trace/expand_trace` cannot surface. Applied identically (within SDK conventions) to the JavaScript and Python SDKs.

See #376 for the problem statement and root-cause analysis.

## The three layers

| Layer | File (JS / Python) | What it does |
| --- | --- | --- |
| **1. Aggregate-steps exhaustion check** *(JS only — Python's loop already walks steps)* | `judge-agent.ts::discoveryExhausted` | Inspects `completion.steps` when present so an earlier `finish_test` isn't hidden by AI SDK v6's last-step-only `toolCalls`. Skips `forceVerdict` entirely when the loop already converged. |
| **2. `forceVerdict` cannot leak a discovery tool** | `judge-agent.ts::forceVerdict` / `judge_agent.py::_force_verdict` | Collapses prior `expand_trace` / `grep_trace` `tool_use`+`tool_result` pairs into plain-text assistant recaps (via `collapseDiscoveryHistory` / `_collapse_discovery_history`), then strips discovery tools from the forced call's tool set. Model physically cannot emit them; Anthropic doesn't reject the call for dangling tool references in history. |
| **3. Safety net at the parser** | `judge-agent.ts::parseToolCalls` / `judge_agent.py::_parse_response` | Any residual leak degrades to an `inconclusive` verdict with reasoning `\"JudgeAgent: trace discovery did not converge on a verdict within the step budget\"`, instead of hard-failing with `Unknown tool call` (JS) or raising (Python). |

For each layer to be bypassed, the two other layers have to fail simultaneously — the symptom in #376 required all three to fail.

## Files touched

- `javascript/src/agents/judge/judge-agent.ts` — helper + three methods.
- `javascript/src/agents/types.ts` — `InvokeLLMResult` now includes `steps`.
- `javascript/src/agents/judge/__tests__/judge-agent.test.ts` — three new cases.
- `python/scenario/judge_agent.py` — helper + two methods.
- `python/tests/test_judge_force_verdict_hardening.py` — four new cases.

## Test plan

- [x] **JS unit:** `pnpm test` → **306/306 pass** (3 new cases directly covering the three layers).
- [x] **Python unit:** `uv run pytest tests/test_judge_*.py` → **35/35 pass** (4 new cases; 2 pre-existing skips unrelated).
- [x] **CI:** `javascript-ci` green. `python-ci` reports 384 passed / 7 skipped, with a single flaky failure in `test_arun_cancellation.py::test_cancellation_does_not_affect_sibling_arun` (`TimeoutError`) that also fails on other branches — unrelated to this change.
- [ ] **Manual end-to-end:** run a scenario on the LangWatch platform with a Claude Opus 4.6 judge and an 8192+ token trace, 10–20 iterations. Expectation: zero `Unknown tool call` failures; runs either finish cleanly or return `inconclusive` with the `did not converge` reason.

## Trade-offs / things worth knowing

- **History rewrite is lossy at the protocol level** (tool_use/tool_result → plain text), but preserves the discovery tool *output* as text so the model can still reason about what was found. No regression in verdict quality in the tests.
- **Discovery tool names are hardcoded** (`{\"expand_trace\", \"grep_trace\"}`) in a `DISCOVERY_TOOL_NAMES` constant per SDK. A first-class registry would be cleaner; deliberately skipped to keep this PR narrow. Anyone adding a new discovery tool must update the constant in both SDKs.
- **This makes `forceVerdict` robust, not rare.** Opus 4.6 still regularly hits the force path because `maxDiscoverySteps = 10` is tuned for `gpt-5-mini`. Provider-aware defaults would reduce how often we enter the force path; tracked as a follow-up, not in scope here.

## Follow-ups (not this PR)

- Provider-aware `maxDiscoverySteps` (higher default for Anthropic).
- Promote `DISCOVERY_TOOL_NAMES` to a shared registry so future discovery tools don't require hunting through three code paths.
- Contract test fixture shared between JS and Python collapse helpers so the two implementations can't drift.
- Metric on `forceVerdict invoked` so we can verify empirically that the fix moved the failure-rate to zero rather than just renaming the failure.

### DIFF
--- a/javascript/src/agents/judge/__tests__/judge-agent.test.ts
+++ b/javascript/src/agents/judge/__tests__/judge-agent.test.ts
@@ -565,6 +565,180 @@ describe("JudgeAgent", () => {
       expect(result!.unmetCriteria).toContain("Agent uses tools");
     });
 
+    it("strips discovery tools and rewrites history before forced verdict", async () => {
+      const config: JudgeAgentConfig = {
+        criteria: ["Agent works correctly"],
+        spanCollector: collector,
+        maxDiscoverySteps: 2,
+      };
+
+      const agent = judgeAgent(config);
+
+      let forcedParams: InvokeLLMParams | undefined;
+      let callCount = 0;
+      agent.invokeLLM = async (params) => {
+        callCount++;
+        if (callCount === 1) {
+          // Simulate a real multi-step loop: last step emits a discovery tool
+          // and the accumulated history carries the matching tool_use /
+          // tool_result pair that would normally trip up a stripped call.
+          const prevMessages = params.messages ?? [];
+          params.messages = [
+            ...prevMessages,
+            {
+              role: "assistant",
+              content: [
+                {
+                  type: "tool-call" as const,
+                  toolCallId: "tc-1",
+                  toolName: "expand_trace",
+                  input: { span_ids: ["deadbeef00000000"] },
+                },
+              ],
+            },
+            {
+              role: "tool",
+              content: [
+                {
+                  type: "tool-result" as const,
+                  toolCallId: "tc-1",
+                  toolName: "expand_trace",
+                  output: { type: "text", value: "span details here" },
+                },
+              ],
+            },
+          ] as any;
+          return {
+            text: "",
+            content: [],
+            toolCalls: [
+              {
+                toolName: "grep_trace",
+                input: { pattern: "error" },
+                type: "tool-call" as const,
+                toolCallId: "tc-2",
+              },
+            ],
+            toolResults: [],
+          } as unknown as InvokeLLMResult;
+        }
+        forcedParams = params;
+        return mockLLMResult("finish_test", {
+          criteria: { agent_works_correctly: "true" },
+          reasoning: "ok",
+          verdict: "success",
+        });
+      };
+
+      const result = await agent.call(createBaseInput());
+
+      expect(callCount).toBe(2);
+      expect(result).not.toBeNull();
+      expect(result!.success).toBe(true);
+
+      // forceVerdict must have stripped expand_trace and grep_trace from the
+      // tool set, leaving only the terminal tools.
+      const toolNames = Object.keys(forcedParams!.tools ?? {});
+      expect(toolNames).not.toContain("expand_trace");
+      expect(toolNames).not.toContain("grep_trace");
+      expect(toolNames).toContain("finish_test");
+
+      // Message history must have been rewritten: no assistant message
+      // should still contain a tool-call part for a discovery tool, and no
+      // tool-role message should remain.
+      const msgs = forcedParams!.messages ?? [];
+      for (const m of msgs) {
+        expect(m.role).not.toBe("tool");
+        if (m.role === "assistant" && Array.isArray(m.content)) {
+          for (const part of m.content as Array<Record<string, unknown>>) {
+            expect(part.type).not.toBe("tool-call");
+          }
+        }
+      }
+
+      // The recap text should carry forward what the discovery tool returned
+      // so the model can still reason about it.
+      const recapHit = msgs.some(
+        (m) =>
+          m.role === "assistant" &&
+          typeof m.content === "string" &&
+          m.content.includes("expand_trace") &&
+          m.content.includes("span details here")
+      );
+      expect(recapHit).toBe(true);
+    });
+
+    it("skips forceVerdict when an earlier step already emitted finish_test", async () => {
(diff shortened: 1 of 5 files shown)


Aryansharma28:
Note on CI: the python-ci failure on this run is `test_arun_cancellation.py::test_cancellation_does_not_affect_sibling_arun — TimeoutError`, a known-flaky test unrelated to this change (recent main-branch runs on unrelated branches show the same failure). All judge tests passed: `384 passed, 7 skipped, 1 flaky-timeout`. Not blocking.

rogeriochaves (approved):

Issue #400, opened 2026-05-02
fix(security): resolve all high-severity Dependabot alerts flagged by Vanta


sergioestebance:
## Summary

Resolve all high-severity Dependabot security alerts flagged during Vanta compliance review. Each dependency gets its own PR to minimize merge risk.

### Merged PRs ✅

| PR | Dependency | Alerts Fixed |
|---|---|---|
| #399 | rollup | #169-#174 |
| #398 | minimatch | #161, #175-#178, #181-#182, #186, #194 |
| #394 | picomatch (pnpm), @hono/node-server, glob (pnpm) | #281, #198, #67, #68 |
| #393 | protobuf, lodash, liquidjs (partial), MCP SDK (partial) | #88, #132, #200, #298 |
| #401 | pyasn1 | #130, #208 |
| #402 | starlette | #114 |
| #403 | black | #202 |
| #404 | urllib3 | #116, #117, #127 |
| #406 | mcp Python SDK | #111, #115 |
| #407 | python-multipart | #131 |
| #408 | lodash-es (partial — override too low, see remaining) | #262 (partial) |
| #409 | picomatch (npm lockfile) | #279 |
| #410 | @modelcontextprotocol/sdk (javascript/) | #77, #87, #107 |
| #411 | litellm | #266, #286, #326, #327 |
| #412 | liquidjs | #215, #216, #294, #325 |
| #413 | glob (npm lockfile) | #67 |
| #414 | hono | #95, #96, #195 |
| #415 | trim-newlines | #30 |
| #416 | path-to-regexp (javascript/) | #238 |
| #417 | aiohttp | #119, #112, #120-#126, #252-#261 |
| #418 | react-router / @remix-run/router | #89, #90, #92, #93 |
| #419 | vite (javascript/, docs/) | #268, #271, #277 |
| #420 | langchain / @langchain/core (javascript/) | #78, #79 |
| #421 | flatted | #209-#213 |

### Open PRs — ready to merge

| PR | Dependency | Risk | Notes |
|---|---|---|---|
| #405 | pydantic-ai + pydantic-ai-slim | MEDIUM | Resolves to 1.31.0 (>=1.56.0 required). Major version jump but examples-only. Fixes #133, #134 + Vanta CVE-2026-25580 |

### PRs to close (superseded)

| PR | Reason |
|---|---|
| #391 | litellm >=1.83.0 — superseded by #411 (merged, main has 1.83.14 which satisfies all Vanta constraints including >=1.83.7) |

[...]

PR #407, opened 2026-05-02
fix(security): upgrade python-multipart to fix arbitrary file write CVE

sergioestebance:
## Summary

Upgrades python-multipart in python/uv.lock (v0.0.20 → v0.0.27) to fix:

- **#131**: Arbitrary File Write via Non-Default Configuration

Part of #400

## Test plan

- [ ] `uv lock` succeeds
- [ ] CI passes

### DIFF
--- a/python/uv.lock
+++ b/python/uv.lock
@@ -2900,11 +2900,11 @@ wheels = [
 
 [[package]]
 name = "python-multipart"
-version = "0.0.20"
+version = "0.0.27"
 source = { registry = "https://pypi.org/simple" }
-sdist = { url = "https://files.pythonhosted.org/packages/f3/87/f44d7c9f274c7ee665a29b885ec97089ec5dc034c7f3fafa03da9e39a09e/python_multipart-0.0.20.tar.gz", hash = "sha256:8dd0cab45b8e23064ae09147625994d090fa46f5b0d1e13af944c331a7fa9d13", size = 37158, upload-time = "2024-12-16T19:45:46.972Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/69/9b/f23807317a113dc36e74e75eb265a02dd1a4d9082abc3c1064acd22997c4/python_multipart-0.0.27.tar.gz", hash = "sha256:9870a6a8c5a20a5bf4f07c017bd1489006ff8836cff097b6933355ee2b49b602", size = 44043, upload-time = "2026-04-27T10:51:26.649Z" }
 wheels = [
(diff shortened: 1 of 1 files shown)


sergioestebance:
## Dependency Upgrade Assessment: python-multipart 0.0.20 → 0.0.27

### 1. Where is this dependency used?

**Not imported anywhere in our code.** Searched the entire codebase for `from multipart`, `import multipart`, `from python_multipart`, and `import python_multipart` — zero matches.

The only file mentioning "multipart" is `python/scenario/_generated/langwatch_api_client/lang_watch_api_client/types.py`, which has a docstring saying `"Return a tuple representation that httpx will accept for multipart/form-data"` — this is about httpx's built-in multipart support, not the python-multipart package.

### 2. Production code impact

**None.** We do not import or call any python-multipart APIs directly. No classes, functions, or constants from this package appear anywhere in `python/scenario/`.

### 3. Examples/tests/docs usage

**None.** No imports in examples, tests, or docs either.

### 4. Version jump size

**0.0.20 → 0.0.27** — seven patch-level bumps in the 0.0.x range. This is a pre-1.0 package so the API contract is technically unstable, but since we don't import it directly, breaking API changes in python-multipart itself don't affect us. What matters is whether the consumers (mcp, streamlit) are compatible with 0.0.27 — and since `uv lock` resolved successfully without version conflicts, they are.

### 5. Pinned version comments

**None found.** Searched for `pin`, `constraint`, `do not upgrade`, `locked to`, `pinned` across all `.py`, `.toml`, `.cfg`, `.txt`, and `.md` files in the python directory — no comments referencing python-multipart version constraints.

### 6. Transitive vs direct

**Transitive only.** python-multipart is NOT listed in any `pyproject.toml`. It's pulled in as a transitive dependency by:
- **mcp** (line 1457 in uv.lock)
- **streamlit** (line 3295 in uv.lock)

We never depend on it directly.

### 7. Final verdict

**TRIVIAL** — Transitive-only dependency, not imported anywhere in our code (production, tests, or examples). The version bump is patch-level within 0.0.x. No pinning constraints exist. `uv lock` resolved cleanly, confirming compatibility with the packages that actually depend on it (mcp, streamlit). Safe to merge once CI passes.

Note: The lock file also pulled in new transitive dependencies (altair, numpy, pandas, streamlit, etc.) because this was the first `uv lock` with the current dependency set — these are existing transitive deps that weren't previously resolved. Worth confirming CI is green to validate the full lockfile.

sergioestebance:
CI green — all 6 checks pass (CodeQL, tests on Python 3.12, PR title validation, semantic PR, evaluate). Ready for review.

sergioestebance:
Rebased onto latest main. No conflicts. CI re-running.

Aryansharma28 (approved):
Reviewed locally: python-multipart 0.0.20 → 0.0.27 closes GHSA #131 (arbitrary file write). Pure transitive — pulled only by mcp[mcp] and streamlit[report] optional extras, never directly imported. 468 unit tests pass. Scenario is client-side, doesn't run an HTTP server, so the CVE is unreachable. Safe minor bump.

PR #413, opened 2026-05-02
fix(security): patch glob CLI command injection in lovable_clone npm lockfile

sergioestebance:
## Summary

Adds npm overrides for glob in python/examples/lovable_clone/template to fix:

- **#67**: Command injection via -c/--cmd executes matches with shell:true (package-lock.json)

Note: pnpm-lock.yaml fix for docs/ is handled by PR #394.

Part of #400

## Test plan

- [x] `npm install --package-lock-only` succeeds
- [x] CI passes


sergioestebance:
## Dependency Update Assessment: `glob` override in lovable_clone template

Searched the entire codebase for actual imports/requires of `glob`:

- **Zero files** in the codebase import `glob` directly. The only grep hit in the template was `eslint.config.js` importing `globals` (different package).
- `glob` appears **only** in the lockfile as a transitive dependency.

### 2. PRODUCTION CODE IMPACT

**Not applicable.** `glob` is not imported anywhere in production code (`javascript/src/`, `python/scenario/`, or anywhere else). It exists solely as a transitive dependency inside an example template.

### 3. EXAMPLES/TESTS/DOCS SCOPE

This change is **entirely scoped to `python/examples/lovable_clone/template/`** — an example project template. It is not part of the scenario SDK, not published to npm/PyPI, and not used in CI beyond lockfile validation.

### 4. VERSION JUMP SIZE

| | Before | After |
|---|---|---|
| glob version | 10.4.5 | 13.0.6 |
| Major versions jumped | — | **3 major versions** |

This is a **significant version jump**. However, the reason npm resolved to 13.0.6 instead of 10.5.0 is that **all glob 10.x versions are now marked as deprecated** by the maintainer (isaacs). npm prefers non-deprecated versions when resolving ranges, so `>=10.5.0` lands on the latest stable (13.0.6).

The glob JS API (`glob()`, `globSync()`, `Glob` class) has remained stable across these versions. The breaking changes in 11.x/12.x/13.x were primarily around: dropping old Node.js versions, removing the callback API (already removed in 10.x), and internal performance improvements.

**None found.** No comments anywhere in the codebase mention pinning or constraining `glob` versions.

### 6. TRANSITIVE vs DIRECT

**Transitive only.** The full dependency chain is:

```
tailwindcss (devDependency) → sucrase (^3.35.0) → glob (^10.3.10)
```

- `tailwindcss` is a **devDependency** in the template
- `sucrase` uses `glob` only in its **CLI** (`dist/cli.js`) for file globbing — not in its core transpiler API
- `tailwindcss` uses sucrase's **programmatic API** for transpiling config files, **not** sucrase's CLI
- Therefore: even if glob 13.x had breaking API changes, they would not affect tailwindcss's usage of sucrase

### 7. FINAL VERDICT: TRIVIAL

**Classification: TRIVIAL** — Dev-only transitive dep, not imported anywhere in our code. Safe to merge with CI passing.

Rationale:
- `glob` is a **transitive devDependency** in an **example template** (not published, not production code)
- **Zero direct imports** of `glob` exist anywhere in the codebase
- The consuming package (`sucrase`) only uses glob in its CLI, and its consumer (`tailwindcss`) uses sucrase's programmatic API — the glob upgrade path is completely inert
- The 3-major-version jump is forced by the maintainer deprecating all 10.x releases; npm resolves to the latest non-deprecated version
- No `node_modules` are even committed — this only affects the lockfile
- No version pin comments exist anywhere in the codebase

**Risk: effectively zero.** The only way this could cause an issue is if someone runs `npm install` in the example template and sucrase's CLI (not used by tailwindcss) fails on glob 13.x — and even that is unlikely since the JS API is stable.

sergioestebance:
CI green — all 7 checks pass (CodeQL, Python tests, PR validation, evaluation, approval gate). Ready for review.

sergioestebance:
Rebased onto latest main. Conflict resolved in `package.json` (merged main's updated minimatch overrides with glob override) and regenerated `package-lock.json`. CI green — all 7 checks pass. Ready for review.

Aryansharma28 (approved):
lgtm

PR #416, opened 2026-05-02
fix(security): patch path-to-regexp DoS via sequential optional groups

sergioestebance:
## Summary

Adds pnpm.overrides for path-to-regexp to fix:

- **#238, #239, #240**: DoS via sequential optional groups

Part of #400

## Test plan

- [ ] `pnpm install --lockfile-only` succeeds
- [ ] CI passes


sergioestebance:
## Dependency Update Risk Assessment: path-to-regexp

**Not imported anywhere in our code.** Searched all `.ts`, `.js`, `.tsx`, `.jsx` files under `javascript/` — zero direct imports of `path-to-regexp`.

### 2. PRODUCTION CODE IMPACT

**None.** `path-to-regexp` is not used in production code (`javascript/src/`). It appears only as a transitive dependency of:
- **`msw@2.12.1`** — optional peer dependency of vitest (our test runner)
- **`router@2.2.0`** — transitive dep of `express@^5.1.0`, which is only in `javascript/examples/vitest/package.json`

### 3. EXAMPLES/TESTS ONLY?

**Yes.** Both consumers are test/example infrastructure:
- `msw` — Mock Service Worker, pulled in as an optional vitest peer dep (not directly listed in any package.json)
- `express` — only in the vitest examples package

### 4. VERSION JUMP SIZE ⚠️

**This is a 2-major-version jump (6.3.0 → 8.3.0) for msw's internal dependency.**

There is no `path-to-regexp@6.3.1` — version 6.3.0 is the last release in the 6.x line. The override `path-to-regexp@<6.3.1: ">=6.3.1"` has no valid 6.x target, so pnpm resolved to 8.3.0 (latest).

The path-to-regexp API changed significantly between 6.x and 8.x (different export shape, different parameter syntax). However, `msw@2.12.1` also lists `path-to-regexp@8.3.0` in the lockfile for other resolution paths — **msw appears to be compatible with both 6.x and 8.x**, or at minimum the pnpm resolver is happy with 8.x here.

`router@2.2.0` (Express 5 internal) already declared path-to-regexp 8.3.0 before this change — unaffected.

**None found.** No comments in the codebase mentioning path-to-regexp with version pins or upgrade warnings.

### 6. TRANSITIVE vs DIRECT

**Purely transitive.** `path-to-regexp` is not listed in any `package.json` under `dependencies` or `devDependencies`. It only appears in lockfiles via `msw` and `router`.

### 7. FINAL VERDICT: LOW RISK

- Not imported in any of our code (zero direct usage)
- Only consumed by test tooling (`msw` via vitest) and an example (`express` in examples/vitest)
- The override effectively removes the vulnerable 6.3.0 from the lockfile entirely
- `msw@2.12.1` resolves cleanly with path-to-regexp 8.3.0 (pnpm found a valid resolution)
- `router@2.2.0` (Express 5) was already on 8.3.0 — no change

**One caveat:** If `msw`'s internal route matching relies on path-to-regexp 6.x-specific API (e.g., `pathToRegexp()` function signature), tests using msw request handlers could break. However: (1) we don't import msw anywhere in our code, (2) msw is only an optional vitest peer dep, and (3) the lockfile resolved without errors. CI will validate this.

**Safe to merge once CI passes.** If msw-based tests exist in any workspace and fail, the fix would be to pin the override to a narrower range or switch to `"path-to-regexp@>=6.0.0 <6.3.1": "8.3.0"` to target only the vulnerable versions more precisely.

sergioestebance:
CI green — all checks pass (CodeQL, ci-checks, PR validation, evaluate). Ready for review.

sergioestebance:
Rebased onto latest main. Conflicts resolved. CI re-running.

sergioestebance:
Rebased onto latest main, conflicts resolved.

PR #495, opened 2026-05-20
fix(#494): emit multimodal SDK content as JSON arrays (regression-first)

### LINKED ISSUE #494: SDK emits multimodal message content as Python-repr strings instead of real JSON arrays (voice audio transport)
**Describe the bug**

The Scenario Python SDK emits multimodal message content (voice `input_audio` parts) as Python-repr strings instead of real JSON arrays. End result: the LangWatch server receives `"[{'type': 'input_audio', ...}]"` as a string and has to resurrect the structure via a fragile parser (`coerce-content-to-array.ts`). When that parser succeeds, the audio extraction pipeline runs; when it fails (single-quoted Python repr inside JSON, edge cases), audio data is silently dropped.

Caught during voice integration testing today against production LangWatch — both `basic_greeting.py` and `elevenlabs_hosted.py` voice scenarios exhibit this. The server's `coerceContentToArray` (with a unit test fixture proving this exact shape is observed in prod) is currently the only thing standing between voice runs and total audio loss.

**Root cause** (confirmed via /investigate)

`scenario/_events/utils.py:54-58, 81-86, 94, 110-114` calls `content=str(content)` on every message during `convert_messages_to_api_client_messages`. When voice flows produce structured content via `voice/messages.py:45-69` (`create_audio_message` correctly builds `[{"type":"input_audio", "input_audio":{...}}]`), the stringification destroys the structure on the wire.

The generated client at `_generated/langwatch_api_client/lang_watch_api_client/models/post_api_scenario_events_body_type_2_messages_item_type_{0..4}.py` types `content: str` — that's what forces the stringification at the SDK boundary.

**Expected behavior**

When a message's `content` is already a structured list (multimodal content parts), it should be emitted on the wire as a real JSON array — not a Python-repr string. Text-only messages continue to emit as plain strings (back-compat).

**To reproduce**

1. `cd python`
2. `LANGWATCH_API_KEY=<sk-lw-*> uv run examples/voice/basic_greeting.py`
3. Tail the HTTP body of any `SCENARIO_MESSAGE_SNAPSHOT` POST (e.g. via mitmproxy or `event_reporter` debug logs). Observe `content` is a JSON string containing a Python repr, not a JSON array.

**Environment**

- Component: langwatch_sdk_python (scenario)
- Files involved:
  - `python/scenario/_events/utils.py:57, 84, 94, 113`
  - `python/scenario/voice/messages.py:45-69` (already correct — builds structured content)
  - `python/scenario/voice/recording.py:107, 126-138` (per-segment WAV source)
  - `python/scenario/_generated/langwatch_api_client/lang_watch_api_client/models/post_api_scenario_events_body_type_2_messages_item_type_{0..4}.py` (generated; types `content: str`)

**Additional context**

The langwatch server side is ready: `content-extractor.ts:248-308` handles `input_audio`, extracts base64 → S3 → rewrites to `/api/files/<id>` ref. Once the SDK emits real arrays, the path Just Works.

UI render gap is filed separately on langwatch/langwatch (#4138) — that's a small shape mismatch in `MediaPart.tsx` (extractor emits `{type:"input_audio", input_audio:{url, mimeType}}`; MediaPart expects `{type:"audio", source:{...}}`).

## Acceptance Criteria

Approach: regenerate `langwatch_api_client` from updated OpenAPI spec to accept `Union[str, list]` for message content. The server already prefers arrays.

1. When `_state.messages` contains a message with `content=[{"type":"input_audio", "input_audio":{"data":<b64>, "format":"wav"}}]`, the emitted `SCENARIO_MESSAGE_SNAPSHOT` JSON has `messages[*].content` as a real JSON array — NOT a Python-repr string.
2. Text-only messages still work: `content="hello"` round-trips as string.
3. Mixed multimodal in one message (text + input_audio) round-trips intact.
4. Unit test in `python/tests/_events/` verifying the wire payload byte-for-byte for a voice turn.
5. End-to-end: running `examples/voice/basic_greeting.py` against `LANGWATCH_ENDPOINT=https://app.langwatch.ai` (post-#4139-fix prod) produces at least one `/api/files/<id>` stored-object reference (verifiable via the existing extractor refs span).

Blocked by: PR-A on #4139 (prod infra) for the E2E AC. AC 1-4 can be verified independently.

### Approach

Stop stringifying structured content at the SDK boundary. The fix is two-part: (1) relax the generated message-model field type from `content: str` to `content: Union[str, list[dict]]` so the wire serializer no longer coerces lists; (2) drop the `str(content)` call in `_events/utils.py` for `user`, `assistant`, `system`, and `tool` branches and pass `content` through as-is. The server already accepts arrays — `coerce-content-to-array.ts` becomes a back-compat tolerance layer rather than the load-bearing parser.

The investigation proposed full client regeneration from an updated OpenAPI spec. That spec lives in a separate repo (`langwatch-saas/langwatch/docs/api-reference/openapiLangWatch.json`, `content: {type: "string"}` at lines 88, 277, …), so a clean regen requires a cross-repo PR first. To unblock voice while that lands, this plan applies a targeted edit to the four generated model files and removes a marker comment so the next regen reapplies the change instead of clobbering it. The marker + a regen test gate the round-trip.

### Sequencing

1. **Relax content type in 4 generated models** — change `content: str` → `content: Union[str, list[dict[str, Any]]]` in `post_api_scenario_events_body_type_2_messages_item_type_{0..4}.py` (4 files; type 2 is assistant — content already optional there, double-check). Update `to_dict` / `from_dict` to pass content through without coercion. Mark the deviation in the file header so future regens are caught. Completion signal: type-check passes, models still import.
2. **Drop `str(content)` in `_events/utils.py`** — for `user`, `assistant`, `system`, `tool` branches, pass `content` through. Preserve the existing `if not content` guards so empty content still raises/warns as today. Completion signal: voice message round-trip test passes (next step).
3. **Add unit tests in `python/tests/_events/`** — covers AC1 (input_audio list → JSON array on wire), AC2 (string content → string on wire), AC3 (mixed text + input_audio in one message). Assert against the serialized HTTP body bytes (intercept via the `httpx` mock transport already used in `event_reporter` tests). Completion signal: `uv run pytest python/tests/_events/ -k content_wire` green.
4. **Spec-source PR (separate, langwatch repo)** — update `openapiLangWatch.json` `content` schema to `oneOf: [string, array of objects]` in both POST event body and trace-message shapes. Tracked here but lands in `langwatch/langwatch`, not this repo. Completion signal: PR merged; future `make generate-openapi-client` produces the union type natively and the manual deviation can be removed.
5. **E2E voice validation (AC5, blocked)** — once `#4139` lands in prod, run `examples/voice/basic_greeting.py` against `app.langwatch.ai`, confirm `/api/files/<id>` reference appears in extractor refs span. Completion signal: span attribute present in the run trace.

### Key decisions

- **Patch generated files now, fix spec later.** A pure spec-driven regen would be cleaner, but it adds a cross-repo dependency to a P1 voice bug. The patched models are small (4 files, one type each) and isolated; the spec-side PR catches up async. Marker comment ensures the deviation isn't silently overwritten.
- **`Union[str, list[dict]]`, not a typed content-part union.** Resist the pull to model `TextPart | InputAudioPart | …` here — that belongs upstream in the OpenAPI spec, not in a quick patch. The server already validates shape; the SDK's job is to pass through what the caller built.
- **Don't touch `voice/messages.py`.** It already produces correct structure; the bug is purely in transport.
- **Byte-level wire test, not just type assertion.** AC4 is explicit: byte-for-byte verification. Using `mock_transport` lets us inspect the JSON `httpx` actually sends, which catches future `str()` regressions immediately.
- **Keep `coerce-content-to-array.ts` server-side.** This SDK fix removes the *need* but not the safety net — older SDKs in the wild still emit strings. Server-side parser stays.

### Risks / open questions

- The four generated model files share boilerplate; if regen tooling rewrites them differently, the marker-comment guard may not survive `--overwrite`. Mitigation: add a CI check that greps for `content: Union` after regen and fails if missing.
- `assistant` content can already be `None` in OpenAI's schema (tool-call-only turns). Verify the patched type allows `Optional[Union[str, list[dict]]]` for the assistant variant only.
- AC5 depends on `#4139` landing — out of our control. AC1–4 can ship independently; do not block PR on AC5.

### Out of scope

- UI render fix (`langwatch/langwatch#4138`) — separate issue, different repo.
- Prod infra fix (`#4139`) — blocks AC5 but not AC1–4.
- Full content-part type modeling (e.g., proper `TextPart | InputAudioPart` discriminated unions in the OpenAPI spec). Tracked for follow-up after spec-side PR.
- TS SDK parity — Python SDK only for this issue.
- Removing the server-side `coerceContentToArray` fallback — back-compat for older SDKs.

---
*Classification: Bug*
*Status: investigated. ACs above. Companion langwatch UI issue #4138.*

drewdrewthis:
## Why

Closes #494.

Voice (and any multimodal) scenarios emit `SCENARIO_MESSAGE_SNAPSHOT` events with `content` stringified via Python's `str()` — so a list of multimodal parts hits the wire as `"[{'type': 'input_audio', ...}]"` (Python repr inside JSON), forcing the LangWatch server to resurrect structure through a fragile parser. When the parser misses, audio is silently dropped. Caught during voice integration testing today against prod.

## What changed

This PR currently contains the **regression-test commit only** — the fix lands next on the same branch.

- Pinned the wire-shape contract: list content must serialize as a JSON array; string content remains a JSON string. Tests assert the *raw bytes* off `httpx`'s transport, not just the in-memory model, because the bug is at the serializer boundary.
- Mirrored the existing `respx`-based pattern from `test_event_reporter.py` rather than introducing a new transport mock; keeps event-suite test conventions consistent.
- Parametrized the role test across `user`/`assistant`/`system`/`tool` to lock all four `str(content)` call sites (`_events/utils.py:57,84,94,113`) into the contract simultaneously.

## Test plan

```
cd python && uv run pytest tests/_events/test_message_content_wire_shape.py -v
```

Current (pre-fix) result: **6 failed / 1 passed** — exactly the bug shape.

- ✗ `test_voice_input_audio_round_trips_as_json_array` — AC 1
- ✓ `test_text_only_content_round_trips_as_json_string` — AC 2 (regression guard; must keep passing post-fix)
- ✗ `test_mixed_multimodal_content_round_trips_intact` — AC 3
- ✗ `test_all_four_role_branches_preserve_list_content[user|assistant|system|tool]` — supports AC 4

Post-fix target: **7 passed.**

## Anything surprising?

- AC 5 (E2E `/api/files/<id>` reference from `examples/voice/basic_greeting.py`) is **blocked by `langwatch/langwatch#4139`** prod infra and ships independently; it does not gate this PR.
- The companion UI render gap is tracked separately at `langwatch/langwatch#4138`.
- The server-side `coerce-content-to-array.ts` parser stays in place as a back-compat fallback for older SDKs — this PR removes the *need* for it on new traffic, not its existence.

### DIFF
--- a/python/scenario/_events/utils.py
+++ b/python/scenario/_events/utils.py
@@ -1,5 +1,5 @@
 import warnings
-from typing import Any, List
+from typing import Any, List, Union
 
 from ..types import ChatCompletionMessageParamWithTrace
 from .events import MessageType
@@ -11,6 +11,7 @@ from .messages import (
     ToolCall,
     FunctionCall,
 )
+from scenario._generated.langwatch_api_client.lang_watch_api_client.types import UNSET, Unset
 from pksuid import PKSUID
 
 
@@ -41,13 +42,17 @@ def convert_messages_to_api_client_messages(
 
         role = message.get("role")
         raw_content = message.get("content")
-        # Narrow the OpenAI SDK's Iterable union to str | list[dict[str, Any]]
-        # so the generated API client constructors accept the value.
-        # - str content passes through unchanged (text-only messages).
-        # - Iterable[ContentPart] (multimodal) becomes list[dict[str, Any]];
-        #   the API client serialises this as a JSON array on the wire.
-        content: str | list[dict[str, Any]] | None
-        if isinstance(raw_content, str) or raw_content is None:
+        # Narrow the OpenAI SDK's Iterable union to the generated client's
+        # expected type (Unset | str | list[dict[str, Any]]) so pyright
+        # accepts the value at all four constructor call-sites.
+        # - str passes through unchanged (text-only messages).
+        # - Iterable[ContentPart] (multimodal) is coerced to list so the API
+        #   client serialises it as a JSON array on the wire (not Python repr).
+        # - None becomes UNSET — the generated models use a sentinel, not None.
+        content: Union[Unset, str, list[dict[str, Any]]]
+        if raw_content is None:
+            content = UNSET
+        elif isinstance(raw_content, str):
             content = raw_content
         else:
             content = list(raw_content)  # type: ignore[arg-type]
--- a/python/scenario/_events/utils.py
+++ b/python/scenario/_events/utils.py
@@ -1,4 +1,5 @@
 import warnings
+from typing import Any, List
 
 from ..types import ChatCompletionMessageParamWithTrace
 from .events import MessageType
@@ -10,7 +11,6 @@ from .messages import (
     ToolCall,
     FunctionCall,
 )
-from typing import List
 from pksuid import PKSUID
 
 
@@ -40,7 +40,17 @@ def convert_messages_to_api_client_messages(
         message_id = message.get("id") or str(PKSUID("scenariomsg"))
 
         role = message.get("role")
-        content = message.get("content")
+        raw_content = message.get("content")
+        # Narrow the OpenAI SDK's Iterable union to str | list[dict[str, Any]]
+        # so the generated API client constructors accept the value.
+        # - str content passes through unchanged (text-only messages).
+        # - Iterable[ContentPart] (multimodal) becomes list[dict[str, Any]];
+        #   the API client serialises this as a JSON array on the wire.
+        content: str | list[dict[str, Any]] | None
+        if isinstance(raw_content, str) or raw_content is None:
+            content = raw_content
+        else:
+            content = list(raw_content)  # type: ignore[arg-type]
 
         # Only include trace_id in additional_properties when it has a value
         trace_props = {}
--- a/python/scenario/_events/utils.py
+++ b/python/scenario/_events/utils.py
@@ -1,4 +1,3 @@
-import json
 import warnings
 
 from ..types import ChatCompletionMessageParamWithTrace
@@ -11,25 +10,8 @@ from .messages import (
     ToolCall,
     FunctionCall,
 )
-from typing import Any, List
+from typing import List
 from pksuid import PKSUID
-
-
-def _serialize_content(content: Any) -> str:
-    """Coerce message content to the string shape the API client expects.
-
-    Plain strings pass through unchanged. Structured content (a list of
-    OpenAI-style content parts, or a dict) is JSON-encoded so downstream
-    consumers can parse it. Previously this used ``str(content)``, which
-    on a list of dicts produces Python repr with single quotes, breaking
-    JSON-based receivers and forcing brittle apostrophe-aware recovery
-    in the langwatch backend.
-    """
-    if isinstance(content, str):
-        return content
-    if content is None:
-        return ""
-    return json.dumps(content)
 
 
 def convert_messages_to_api_client_messages(
@@ -72,7 +54,7 @@ def convert_messages_to_api_client_messages(
             message_ = UserMessage(
                 id=message_id,
                 role="user",
-                content=_serialize_content(content),
+                content=content,
             )
             message_.additional_properties = trace_props
             converted_messages.append(message_)
@@ -99,7 +81,7 @@ def convert_messages_to_api_client_messages(
             message_ = AssistantMessage(
                 id=message_id,
                 role="assistant",
-                content=_serialize_content(content),
+                content=content,
                 tool_calls=api_tool_calls,
             )
             message_.additional_properties = trace_props
@@ -109,9 +91,7 @@ def convert_messages_to_api_client_messages(
                 raise ValueError(
                     f"System message at index {i} missing required content"
                 )
-            message_ = SystemMessage(
-                id=message_id, role="system", content=_serialize_content(content)
-            )
+            message_ = SystemMessage(id=message_id, role="system", content=content)
             message_.additional_properties = trace_props
             converted_messages.append(message_)
         elif role == "tool":
@@ -130,7 +110,7 @@ def convert_messages_to_api_client_messages(
             message_ = ToolMessage(
                 id=message_id,
                 role="tool",
-                content=_serialize_content(content),
+                content=content,
                 tool_call_id=tool_call_id,
             )
             message_.additional_properties = trace_props
--- a/python/scenario/_generated/langwatch_api_client/lang_watch_api_client/models/post_api_scenario_events_body_type_2_messages_item_type_0.py
+++ b/python/scenario/_generated/langwatch_api_client/lang_watch_api_client/models/post_api_scenario_events_body_type_2_messages_item_type_0.py
@@ -1,4 +1,7 @@
 from collections.abc import Mapping
+# DEVIATION FROM GENERATED CODE (issue #494): `content` widened from `str` to `Union[str, list[Any]]`
+# to allow multimodal (e.g. voice input_audio) content to pass through as a real JSON array.
+# Reapply after `make generate-openapi-client` until OpenAPI spec is updated upstream.
 from typing import Any, Literal, TypeVar, Union, cast
 
 from attrs import define as _attrs_define
@@ -15,13 +18,13 @@ class PostApiScenarioEventsBodyType2MessagesItemType0:
     Attributes:
         id (str):
         role (Literal['developer']):
-        content (str):
+        content (Union[str, list[Any]]):
         name (Union[Unset, str]):
     """
 
     id: str
     role: Literal["developer"]
-    content: str
+    content: Union[str, list[Any]]
     name: Union[Unset, str] = UNSET
     additional_properties: dict[str, Any] = _attrs_field(init=False, factory=dict)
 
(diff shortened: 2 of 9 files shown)


drewdrewthis:
### Sweep: `str(content)` coercion on potentially-multimodal payload fields

**Source:** commits `65747ed..c639f1a` on this PR.

**Pattern:** `str(<field>)` coercion at an SDK boundary where the field is loosely typed (`dict` literal access) and may carry a structured list/dict, mangling multimodal content into Python-repr strings.

Scanned `python/scenario/` (excluding the `_generated/` tree).

#### Must-fix (0)
None.

#### Review (2) — same anti-pattern, different surface, **out of scope for #494**

| File | Line | Risk |
|------|------|------|
| `python/scenario/scenario_executor.py` | 317, 322–326 | `str(message["content"])` for OTel `_trace.update(input/output=...)`. If a voice message reaches `_broadcast_message`, the LangWatch trace would record a Python-repr string instead of structured parts — observability degradation, not data loss. |
| `python/scenario/red_team_agent.py` | 680, 690 | `_get_last_assistant_content` / `_get_last_user_content` return `str(content)`. Downstream `_detect_refusal` does `.lower()` over a Python-repr string — refusal detection silently degrades on multimodal turns. |

#### Clean
All other `str(...)` call sites in `python/scenario/` are either coercing primitives (`Path`, port numbers, Twilio SIDs, URL objects) or operating on values already guaranteed to be strings (red-team JSON-dict reads, report formatter, event-bus response fields).

#### Recommendation

Fix in a follow-up rather than expanding this PR's scope past its ACs. Both review-flagged sites need their own test coverage and a small contract change (return `Union[str, list]` or render via `json.dumps` instead of `str()`). Filing a follow-up issue tracked alongside this PR.

drewdrewthis:
Sweep follow-up filed as #496 — covers the two review-flagged sites (OTel trace input/output + red-team refusal detection). Out of scope for this PR.

drewdrewthis:
**Prove-it: 4/5 PASS, 1 BLOCKED (external).** AC1-4 verified via `python/tests/_events/test_message_content_wire_shape.py` (7/7 passing). AC5 (E2E `/api/files/<id>` ref) is externally blocked by langwatch/langwatch#4139 per the issue body — does not gate this PR.

drewdrewthis:
**Prove-it re-verification (since c639f1a):** 4/5 PASS, 1 BLOCKED (external).

- AC 1-4: wire-shape suite 11/11 green (was 7/7; +4 from role-parametrize, +1 from create_audio_message E2E).
- Broader regression: 19/19 green.
- AC 5: still externally blocked by langwatch/langwatch#4139; tracking in #503.

drewdrewthis:
**Heads-up — rebase superseded #497's `_serialize_content` helper.**

Main now contains PR #497 (`fix(voice): render audio messages cleanly in the terminal`) which introduced a `_serialize_content()` helper in `_events/utils.py` that does `json.dumps(content)` for non-string content. That helper landed conflicting with this PR's fix during the rebase.

I resolved it by **deleting `_serialize_content()`** and keeping this PR's `content=content` pass-through.

Why:

- #497's own description says it does NOT close #494 ("the real fix lives in #494's plan"). #497's `_serialize_content` was a step toward the right shape (no more Python repr) but it stringifies the whole list. AC1 of #494 explicitly says: *"messages[*].content in the JSON body is a JSON array (not a string)"* — so `_serialize_content`'s JSON-string-of-a-list shape would still fail this PR's AC.
- This PR's approach (let the list pass through to the JSON serializer) produces the AC1-compliant real-JSON-array wire shape.
- #497's orthogonal wins are preserved: `_format_message_content` (terminal rendering) and `_redacted_event_repr` (log scrubbing) are different files, completely intact. Their redaction tests (`test_redacted_event_repr_scrubs_b64_inside_stringified_content`, `test_post_event_failure_log_redacts_base64_audio`) still pass against the new list-shaped content — the dict walker handles both shapes.

Test result after rebase: 19/19 across `tests/_events/test_message_content_wire_shape.py` (11) and `tests/test_event_reporter.py` (8 incl. both #497 redaction tests).

cc @rogeriochaves — flagging because the supersede is intentional but worth a quick eye.

drewdrewthis:
[grinder] **READY for human review** (still draft — Drew to un-draft when ready)

CI: green (all checks pass, python-ci run 26376960724)
Review: awaiting human approval (no CHANGES_REQUESTED)
Threads: 0 unresolved

**What this PR fixes:** multimodal message content (lists of typed parts like \`input_audio\`, \`text\`) was being stored via Python's \`str()\` which produces single-quoted Python repr — invalid JSON that the backend couldn't parse. The fix passes the list through and relies on the API client's JSON serialiser to emit a proper array.

**Commits in this PR:**
1. \`test(#494)\` — regression tests first (failing)
2. \`fix(#494)\` — widen generated content type to \`Union[str, list[Any]]\`
3. \`fix(#494)\` — stop stringifying message content at SDK boundary
4. \`test(#494)\` — tighten Union type + broader wire-shape coverage
5. \`test(#494)\` — align regression assertions with pass-through fix [grinder]
6. \`fix(#494)\` — narrow content type for pyright + add cast helper in wire-shape test [grinder]
7. \`fix(#494)\` — use UNSET sentinel for None content, fix None assignability error [grinder]

Verified by:
  \`gh pr checks 495\` → all pass (python-ci: test (3.12), type check, evaluate, javascript-complete, docs-complete)
  \`uv run pytest tests/test_convert_messages_serialization.py tests/_events/test_message_content_wire_shape.py\` → 17 passed, 0 failed
  GraphQL reviewThreads → 0 total, 0 unresolved

drewdrewthis:
Superseded by #497 (merged 2026-05-20), which closed the underlying issue #494 as completed. This PR fixes an already-resolved issue and is currently CONFLICTING against main. Closing as superseded — the multimodal/audio-message rendering fix shipped via #497. Reopen if a distinct regression surfaces that #497 did not cover.

Issue #498, opened 2026-05-20
PipecatAgentAdapter times out on second turn — TimeoutError with empty exception body hides root cause


drewdrewthis:
## Summary

When running the `examples/voice/angry_customer.py` example end-to-end against the LangWatch event endpoint, the Pipecat-based agent under test produces its first audio turn successfully but, after the scripted user's second turn, never produces a follow-up audio response. The `PipecatAgentAdapter` times out waiting for inbound audio and re-raises with a blank inner-exception body, leaving operators with no signal about what actually failed.

Two distinct bugs are visible in a single repro:

- **Primary (timeout)**: agent fails to respond to the second turn; root cause is unknown without instrumentation — could be agent-side hang, `response_timeout` too short for a real LLM + TTS round-trip, VAD / end-of-speech not firing on synthetic audio, or a Pipecat background-task crash that never propagates to the inbound queue.
- **Diagnostic (exception swallowing)**: the executor re-raise formats the cause as an empty string for `TimeoutError`, so the user-visible error reads `RuntimeError: [PipecatAgentAdapter]` with no payload. Underlying Pipecat pipeline errors are likely also being swallowed.

## Expected

A voice scenario that runs multi-turn against a Pipecat agent should either:

1. Complete successfully with audio for every turn, OR
2. Fail loudly with a precise, attributable error — naming what timed out (first chunk vs subsequent chunk), what the configured timeout was, and surfacing any background-task exception from the Pipecat pipeline.

## Actual

After the second user turn, the agent shows a spinner and then the run dies with:

```
RuntimeError: [PipecatAgentAdapter]
```

Inner traceback shows `asyncio.TimeoutError` raised at `python/scenario/voice/adapters/pipecat.py:230` (`asyncio.wait_for(self._inbound_queue.get(), timeout=timeout)`). The re-raise at `python/scenario/scenario_executor.py:893` (`raise RuntimeError(f"[{agent_name}] {e}") from e`) renders the `TimeoutError` as an empty string because timeout errors carry no `.args` payload. The `from e` preserves `__cause__` but the printed text is unhelpful.

## To reproduce

1. Check out a worktree of `main` running the voice example.
2. Configure the LangWatch event endpoint and Pipecat adapter as in `examples/voice/angry_customer.py`.
3. Run `uv run python examples/voice/angry_customer.py`.
4. Observe: first agent turn lands, scripted user turn lands, second agent turn never arrives, run dies with the blank-bodied `RuntimeError`.

## Environment

[...]

PR #499, opened 2026-05-20
fix(voice): stub bot barge-in cancelled STT mid-pipeline, dropping user transcripts

rogeriochaves:
## Summary

- The example Pipecat stub bot at `examples/voice/_bot/bot.py` cancelled its own STT/LLM pipeline whenever fresh user audio arrived after a VAD-driven flush, even though the bot had not started speaking yet. The user's transcribed utterance was silently dropped and `PipecatAgentAdapter.recv_audio` timed out at 30s.
- Gated `_maybe_barge_in` on a new `bot_speaking` flag set inside `_send_tts` (threaded via `on_speaking` callback through `_greet` and `_process_user_audio`). Cancellation now only fires while audio is actually on the wire, which is the only state a real user can barge in on.
- Added `tests/voice/test_bot_barge_in.py` with two cases: the regression itself (no cancellation during STT) and a counterweight (cancellation still fires during TTS).

## Repro

Running `examples/voice/angry_customer.py` against the bundled stub bot timed out:

```
  File ".../adapters/pipecat.py", line 230, in recv_audio
    return await asyncio.wait_for(self._inbound_queue.get(), timeout=timeout)
TimeoutError
```

The bot log proved the bot transcribed the user perfectly, then cancelled itself:

```
16:58:40.046  STT raw text='You charged me twice for my coffee in this noisy cafe. Fix it now.' (input bytes=44000)
16:58:39.915  barge-in: cancelling in-flight response task
16:58:40.861  STT raw text='' (input bytes=7840)
16:58:40.862  empty transcript, skipping response
```

Trigger sequence with paced user audio + a mid-utterance pause (ElevenLabs `[shouting]`/`[angry]` markers): VAD detected end-of-turn after 300ms silence → first flush kicked off STT → trailing user frames re-triggered VAD speech → `_maybe_barge_in` cancelled the STT task → second flush ran STT on a ~1s tail slice that transcribed to empty.

## Fix

`_maybe_barge_in` now no-ops when `bot_speaking` is False. STT-phase and LLM-phase cancellations stop. `_send_tts` sets the flag right before the first outbound media frame and clears it in `finally`.

## Test plan

- [x] `tests/voice/test_bot_barge_in.py::test_bot_does_not_cancel_stt_on_trailing_user_audio` fails on `main`, passes on this branch
- [x] `tests/voice/test_bot_barge_in.py::test_barge_in_still_cancels_while_bot_is_tts_ing` (counterweight) passes
- [x] `tests/voice/test_interruption.py` + `tests/voice/test_pipecat_adapter.py` (18 tests) still pass
- [x] `uv run examples/voice/angry_customer.py` runs to completion with a real bot reply; `recordings/angry_customer/` updated


drewdrewthis (commented):
Reviewed the bug story, the fix, and the tests. Diagnosis is sharper than what I'd converged on in #498 — your bot-side log capturing `STT raw text='...' → barge-in: cancelling → STT raw text=''` is the actual smoking gun. The `bot_speaking` gate matches the semantic of "barge-in means the user talked over the bot while it was talking," and the threading covers all four entry points into `_send_tts` (DTMF, greeting × 2 paths, idle reprompt, post-LLM reply) — I checked each call site against the diff.

One timing concern: pre-fix this bug was flaky in my own #498 repro (1 of 3 runs failed), and the PR description's `[x] uv run examples/voice/angry_customer.py runs to completion` is one successful post-fix run. Could you confirm 5+ consecutive post-fix runs all succeed, so we know the fix is hermetic rather than just shifting the race window? Even better: an e2e test that runs the angry_customer scenario once with the patched bot (like the other voice e2e tests in `tests/voice/`) would lock in the end-to-end behavior.

Few thoughts inline. None are blockers; submitting as COMMENT rather than APPROVE since I haven't been given an explicit approve directive.

Follow-ups worth filing but out of scope of this PR:
- `scenario_executor.py:893` `raise RuntimeError(f"[{agent_name}] {e}")` renders an empty body for `TimeoutError` (which has no `.args`), so the outer error reads `RuntimeError: [PipecatAgentAdapter]` with no payload. Preserving `type(e).__name__` would have made your debugging session faster too.
- The bundled bot subprocess stdout is buffered when `ensure_pipecat_bot` redirects to `/tmp/voice-pipecat-bot.log`, so the bot's own log is usually empty during post-mortem-grepping. The fact that you were able to capture the smoking-gun lines suggests you ran the bot in the foreground or with unbuffered output — most debuggers won't, and this materially hampered #498's investigation.
- Default `response_timeout=30s` is tight for a real STT→LLM→TTS round-trip; consider bumping it as a hedge (separate from this fix).

The `_handle_connection` function in `bot.py` is getting long (~300+ lines with several nested closures sharing `nonlocal` state — `response_task`, `bot_speaking`, `accumulated_mulaw`, `inactivity_task`, etc.). Not for this PR, but worth a future refactor into a small connection-state class so adding the next per-connection bit doesn't require threading another `Callable` through 4 function signatures.

python/examples/voice/_bot/bot.py
```diff
@@ -317,9 +317,16 @@ async def _handle_connection(websocket) -> None:  # type: ignore[no-untyped-def]
...
+    # speech while this is non-None AND the bot is actively TTS-ing, we
+    # cancel it — that's barge-in: the user interrupts the bot mid-utterance,
+    # the bot stops talking and listens.
     response_task: Optional[asyncio.Task] = None
+    # True only while _send_tts is actively writing media frames to the wire.
+    # Gates barge-in: cancelling response_task during the STT/LLM phase is
+    # wrong (the bot hasn't said anything yet, there is nothing to barge in
+    # on) and used to drop the user's first transcribed utterance on the
+    # floor whenever paced trailing audio arrived after a VAD-driven flush.
+    bot_speaking = False
```
drewdrewthis, line 332: Worth one line in the docstring above: the flag flips True _after_ the executor returns the TTS PCM bytes (see `_send_tts` body below), which means while the OpenAI TTS API call is in flight `bot_speaking` is still False and barge-in is suppressed. That's the right semantic (no audio on the wire yet → nothing to barge into), but a reader landing here may wonder why a barge-in during a slow TTS API call is silently ignored. Suggest: "…actively writing media frames; in particular False during the executor-bound TTS synthesis call."

python/examples/voice/_bot/bot.py
```diff
@@ -419,10 +433,22 @@ def _maybe_barge_in() -> None:
...
+        Gated on ``bot_speaking``: we only cancel during the active TTS-send
+        phase. During STT or LLM the bot has not emitted any audio yet, so
+        there is nothing for the user to barge in on — cancelling there
+        just discards the user's transcribed utterance and leaves the bot
+        silent.
+        """
         nonlocal response_task
         if response_task is None or response_task.done():
             return
+        if not bot_speaking:
```
drewdrewthis, line 450: Nice: the `logger.debug` here keeps the suppressed-barge-in path observable, which would have helped enormously during #498's debugging — pairs well with this PR's diagnosis story. No change requested.

python/tests/voice/test_bot_barge_in.py
```diff
@@ -0,0 +1,320 @@
...
+    def stub_stt(mulaw_bytes: bytes) -> str:
+        stt_calls.append(bytes(mulaw_bytes))
+        # ~3000 bytes µ-law @ 8kHz = ~375ms of audio. A real STT will
+        # return useful text only for buffers above the rough fragment
+        # threshold; below that, Whisper returns "" in practice.
+        # Slow the call slightly so the test deterministically exposes
+        # the cancel-during-STT window — without a small delay the
+        # executor thread can finish before the next frame hits the
+        # main loop and barge-in would race against done().
+        time.sleep(0.05)
```
drewdrewthis: The `time.sleep(0.05)` is the only thing that opens the race window this test depends on. If a future maintainer profiles tests and rips it as "a synchronous sleep in async code," the test starts passing for the wrong reason (the cancel-during-STT window closes before VAD can fire). The preceding comment is good but doesn't say "do not delete this." Suggest pulling the magic number into a named constant with a stronger note, e.g. `_STT_WINDOW_S = 0.05  # MUST be wide enough for trailing-speech frames to land before STT returns; reducing this can falsely pass the test by closing the race window`. Same concern applies to the bare `await asyncio.sleep(0.01)` in `_drive_angry_customer_pattern` — fine, just easy to misread as cleanup-able.

python/tests/voice/test_bot_barge_in.py
```diff
@@ -0,0 +1,320 @@
...
+    # Give the loop a moment to process and cancel.
+    await asyncio.sleep(0.2)
+
+    ws.close()
+    await asyncio.wait_for(handler, timeout=3.0)
+
+    # If barge-in fired correctly, the slow_send_tts task was cancelled
+    # before its asyncio.sleep(2.0) finished — so its post-sleep
+    # "media" frame never made it to ws.sent.
+    media_sent = [m for m in ws.sent if '"event": "media"' in m]
```
drewdrewthis, line 348: The counterweight asserts on `ws.sent` being empty of media frames, which proves the slow_send_tts was cancelled before it could send. But that's also true if `on_speaking(True)` never fired and `_maybe_barge_in` just no-ops every time — the test would pass for the wrong reason. Could be strengthened by also asserting that `_set_speaking` was observed at some point with `True` (e.g., wrap it via a tiny spy in the fixture). Not a blocker — the assertion as written does catch a regression of "barge-in never fires anymore."

drewdrewthis:
Filed the three diagnostic / ergonomics follow-ups from the review as separate stub issues so this PR can stay focused:

- #500 — `scenario_executor` re-raise renders empty body for no-message exceptions
- #501 — `ensure_pipecat_bot` subprocess stdout buffering swallows diagnostics
- #502 — voice adapter `response_timeout=30s` default is tight for real round-trips

All three are stubs (no `/investigate` chain run yet) — pickable when someone has the cycles.

rogeriochaves:
@drewdrewthis thanks — all four addressed in `e2a77bc`:

- **bot.py:329 (bot_speaking docstring)** — added the executor-bound TTS clarification you suggested.
- **bot.py:447 (debug log)** — leaving as-is per "no change requested"; agreed it pairs well with the #498 diagnosis path.
- **test_bot_barge_in.py:117 (`time.sleep(0.05)`)** — pulled both magic sleeps into `_STT_RACE_WINDOW_S` and `_PRE_TAIL_YIELD_S` with DO-NOT-REDUCE comments explaining the race window.
- **test_bot_barge_in.py:316 (counterweight)** — added a `speaking_observations` spy + assertion so an empty `ws.sent` can no longer pass vacuously when `on_speaking` was never invoked.

Also flagged the three follow-ups (#500/#501/#502) — agreed those are out of scope here.

drewdrewthis (dismissed):
lgtm

PR #545, opened 2026-05-25
feat(sdk): add langwatch config to ScenarioConfig for race-condition-free multi-tenant runs

### LINKED ISSUE #203: feat: Add programmatic langwatch config to ScenarioConfig (remove env var dependency)
## Problem

When using `@langwatch/scenario` programmatically from a server (e.g., running scenarios on-demand from an API), the SDK forces callers to use environment variables for LangWatch configuration:

```typescript
// Current: Must mutate process.env before each run
process.env.LANGWATCH_API_KEY = projectApiKey;
process.env.LANGWATCH_ENDPOINT = endpoint;

await ScenarioRunner.run({ ... });
```

This creates race conditions when running multiple scenarios concurrently for different projects, forcing the caller to implement a mutex:

https://github.com/langwatch/langwatch/blob/main/langwatch/src/server/scenarios/simulation-runner.service.ts#L21-L47

## Root Cause

1. `run()` in `javascript/src/runner/run.ts` reads from `getEnv()` which parses `process.env`
2. `EventBus` is instantiated with these env-derived values
3. `setup.ts` initializes observability at module load time using env vars

## Proposed Solution

Add a `langwatch` field to `ScenarioConfig` that takes precedence over environment variables:

```typescript
interface LangWatchConfig {
  apiKey?: string;
  endpoint?: string;
}

interface ScenarioConfig {
  // ... existing fields
  
  /**
   * LangWatch reporting configuration.
   * Takes precedence over LANGWATCH_API_KEY and LANGWATCH_ENDPOINT env vars.
   */
  langwatch?: LangWatchConfig;
}
```

### Implementation Steps

1. **Add `langwatch` field to `ScenarioConfig`** (`javascript/src/domain/scenarios/index.ts`)

2. **Modify `run()`** (`javascript/src/runner/run.ts`):
   ```typescript
   const envConfig = getEnv();
   const langwatchConfig = {
     endpoint: cfg.langwatch?.endpoint ?? envConfig.LANGWATCH_ENDPOINT,
     apiKey: cfg.langwatch?.apiKey ?? envConfig.LANGWATCH_API_KEY,
   };
   eventBus = new EventBus(langwatchConfig);
   ```

3. **Handle observability setup** (`javascript/src/tracing/setup.ts`):
   - Option A: Lazy initialization per run (breaking change)
   - Option B: Accept that observability uses env vars at load time, but event reporting uses per-run config (pragmatic)
   - Option C: Create a `configureObservability()` function that can be called before `run()`

## Consumer Example (After Fix)

```typescript
// No mutex needed - each run is self-contained
await ScenarioRunner.run({
  name: 'Test',
  description: 'Scenario',
  agents: [...],
  langwatch: {
    apiKey: project.apiKey,
    endpoint: 'https://app.langwatch.ai',
  },
});
```

## Context

This is blocking proper multi-tenant scenario execution in the LangWatch platform:
- https://github.com/langwatch/langwatch/blob/main/langwatch/src/server/scenarios/simulation-runner.service.ts

## Labels

- enhancement
- sdk-js

drewdrewthis:
## Why

Running scenarios concurrently for multiple LangWatch projects forces callers to mutate `process.env` before each `run()`, creating race conditions. This PR adds a `langwatch` field to `ScenarioConfig` (and `RunOptions`) that takes precedence over env vars, eliminating the mutex pattern. It also adds `projectId` so callers using organization-level or bearer API keys can explicitly route events to the right project.

Closes #203

## What changed

- **`LangwatchConfig` added to `ScenarioConfig` and `RunOptions`** — priority chain is `options.langwatch > cfg.langwatch > LANGWATCH_* env vars`; no env mutation needed for concurrent multi-tenant runs.
- **`projectId` field added to `LangwatchConfig`** — forwarded as `X-Project-Id` header to `/api/scenario-events`, enabling the server-side auth middleware's bearer+X-Project-Id auth path for org-level keys.
- **`LangwatchConfig` type moved to the domain layer** — avoids circular import; `runner/run.ts` now imports it from `../domain`.

## How it works

```
ScenarioConfig.langwatch      RunOptions.langwatch
     {endpoint, apiKey, projectId}
              ↓ merge (options wins)
           EventBus(config)
              ↓
        EventReporter
          X-Auth-Token: apiKey
          X-Project-Id: projectId   ← new
          POST /api/scenario-events
```

The server-side auth middleware already supports `X-Project-Id` alongside `X-Auth-Token` for bearer/org-level keys — this PR wires the client side to match.

## Test plan

```bash
cd javascript && pnpm test   # 834 passed, 0 failed
pnpm typecheck               # clean
```

New unit tests (in `run.test.ts`):
- `cfg.langwatch` used when no `options` provided
- `options.langwatch` takes priority over `cfg.langwatch`
- `projectId` passed through from `cfg.langwatch` to `EventBus`
- `options.langwatch.projectId` takes priority over `cfg.langwatch.projectId`

No playable artifact — pure SDK change. See Test plan: 834 unit tests green (including 4 new tests covering the priority chain and projectId propagation), typecheck clean.

### DIFF
--- a/javascript/src/domain/scenarios/index.ts
+++ b/javascript/src/domain/scenarios/index.ts
@@ -9,6 +9,17 @@ export const DEFAULT_MAX_TURNS = 10;
 export const DEFAULT_VERBOSE = false;
 
 /**
+ * Configuration for LangWatch event reporting.
+ * All fields are optional — any omitted fields fall back to environment variables.
+ */
+export interface LangwatchConfig {
+  /** The endpoint URL to send events to. Falls back to LANGWATCH_ENDPOINT env var. */
+  endpoint?: string;
+  /** The API key for authentication. Falls back to LANGWATCH_API_KEY env var. */
+  apiKey?: string;
+}
+
+/**
  * Configuration for a scenario.
  */
 export interface ScenarioConfig {
@@ -100,6 +111,27 @@ export interface ScenarioConfig {
    * always read off `cfg.voice`. See `voice/config.ts#resolveVoiceConfig`.
    */
   voice?: VoiceConfig;
+
+  /**
+   * LangWatch reporting configuration.
+   * Takes precedence over LANGWATCH_API_KEY and LANGWATCH_ENDPOINT environment variables.
+   *
+   * Use this when running multiple scenarios concurrently for different projects
+   * to avoid race conditions from mutating process.env.
+   *
+   * @example
+   * ```typescript
+   * await run({
+   *   name: 'My scenario',
+   *   langwatch: {
+   *     apiKey: project.apiKey,
+   *     endpoint: 'https://app.langwatch.ai',
+   *   },
+   *   agents: [...],
+   * });
+   * ```
+   */
+  langwatch?: LangwatchConfig;
 }
 
 /**
--- a/javascript/src/domain/scenarios/index.ts
+++ b/javascript/src/domain/scenarios/index.ts
@@ -17,6 +17,8 @@ export interface LangwatchConfig {
   endpoint?: string;
   /** The API key for authentication. Falls back to LANGWATCH_API_KEY env var. */
   apiKey?: string;
+  /** The project ID to send events to. Required when using organization-level or bearer API keys. */
+  projectId?: string;
 }
 
 /**
--- a/javascript/src/events/event-bus.ts
+++ b/javascript/src/events/event-bus.ts
@@ -25,7 +25,7 @@ export class EventBus {
   private logger = new Logger("scenario.events.EventBus");
   private static globalListeners: Array<(bus: EventBus) => void> = [];
 
-  constructor(config: { endpoint: string; apiKey: string | undefined }) {
+  constructor(config: { endpoint: string; apiKey: string | undefined; projectId?: string }) {
     this.eventReporter = new EventReporter(config);
     this.eventAlertMessageLogger = new EventAlertMessageLogger();
     EventBus.registry.add(this);
--- a/javascript/src/events/event-reporter.ts
+++ b/javascript/src/events/event-reporter.ts
@@ -10,13 +10,15 @@ import { Logger } from "../utils/logger";
  */
 export class EventReporter {
   private readonly apiKey: string;
+  private readonly projectId: string | undefined;
   private readonly eventsEndpoint: URL;
   private readonly eventAlertMessageLogger: EventAlertMessageLogger;
   private readonly logger = new Logger("scenario.events.EventReporter");
   private readonly isEnabled: boolean;
 
-  constructor(config: { endpoint: string; apiKey: string | undefined }) {
+  constructor(config: { endpoint: string; apiKey: string | undefined; projectId?: string }) {
     this.apiKey = config.apiKey ?? "";
+    this.projectId = config.projectId;
     this.eventsEndpoint = new URL("/api/scenario-events", config.endpoint);
     this.eventAlertMessageLogger = new EventAlertMessageLogger();
     this.eventAlertMessageLogger.handleGreeting();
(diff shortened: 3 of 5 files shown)


drewdrewthis:

Review: APPROVED by automation (low-risk-change policy)
Threads: zero unresolved

Verified by:
  \`command gh pr checks 545\` → all pass: ci-checks (24.x) pass, javascript-complete pass, python-complete pass
  \`command gh pr view 545 --json reviewDecision\` → "APPROVED"

javascript/src/domain/scenarios/index.ts
```diff
@@ -8,6 +8,17 @@ import { ScenarioExecutionStateLike, ScenarioResult } from "../core/execution";
 export const DEFAULT_MAX_TURNS = 10;
 export const DEFAULT_VERBOSE = false;
 
+/**
+ * Configuration for LangWatch event reporting.
+ * All fields are optional — any omitted fields fall back to environment variables.
+ */
+export interface LangwatchConfig {
+  /** The endpoint URL to send events to. Falls back to LANGWATCH_ENDPOINT env var. */
+  endpoint?: string;
```
0xdeafcafe, line 17: needs project id too

drewdrewthis: Added `projectId?: string` to `LangwatchConfig`. It is forwarded as an `X-Project-Id` header to the `/api/scenario-events` endpoint, matching the server-side auth middleware bearer+X-Project-Id auth path.

drewdrewthis:
## Review verdict: READY

**Reviewers:** principles · hygiene · security · test · Uncle Bob · Metz & Beck · Fowler  
**Diff:** `javascript/src/` — `domain/`, `events/`, `runner/`  
**ACs:** all 3 issue #203 ACs covered (LangwatchConfig interface, ScenarioConfig.langwatch, priority chain)

### Blocking

*(none)*

### Non-blocking

1. **[security][uncle-bob] `event-reporter.ts:49` — header injection risk on `X-Project-Id`**  
   `projectId` flows from `ScenarioConfig` → `EventReporter` → HTTP header without CR/LF validation. Risk is real but classified **Decide** (not Fix): the value is developer-controlled SDK configuration, not end-user input; Node 18+ native `fetch()` (undici) validates header values and throws `TypeError` on embedded `\r\n` before the wire. Explicit validation at the constructor would add defence-in-depth and a clearer error message — worth doing but not a hard gate for this PR.

2. **[principles][hygiene][fowler] Duplicate inline config shape in 4 places**  
   `{ endpoint: string; apiKey: string | undefined; projectId?: string }` appears verbatim in `event-bus.ts:28`, `event-reporter.ts:19`, `run.ts` (EventBus call), and `run.test.ts:368,408`. Fowler: "Data Clump / Shotgun Surgery — every new field is 4 edits." Recommend a named `EventReporterConfig` type exported from `events/`. Deferred to the next-field occasion — acceptable per Metz & Beck ("duplication is better than the wrong abstraction right now"). **New Issue.**

3. **[hygiene][uncle-bob] No `LANGWATCH_PROJECT_ID` env var fallback**  
   `apiKey` and `endpoint` fall back to env vars; `projectId` does not. Intentional (the SDK can't know which project without an explicit signal), but the asymmetry is undocumented. **Decide:** either add a sentence to the JSDoc on `LangwatchConfig.projectId` or add `LANGWATCH_PROJECT_ID` env support alongside the others.

4. **[metz-beck][fowler] Silent no-op when `projectId` is set but `apiKey` is empty**  
   `isEnabled = this.apiKey.length > 0 && ...` — if a caller passes `projectId` but omits `apiKey`, events are silently dropped with no diagnostic. Metz & Beck: "Fail loudly, fail early." Consider a `console.warn` at constructor time when `projectId` is truthy but `isEnabled` is false. **Decide.**

5. **[test][security] `EventReporter.postEvent` X-Project-Id header emission untested**  
   No test verifies that the `X-Project-Id` header is (a) emitted when `projectId` is set, (b) absent when it is not, and (c) `X-Auth-Token` is never `"undefined"`. Run-level tests cover the EventBus constructor call shape but stop before the HTTP layer. **New Issue.**

rogeriochaves (approved):

PR #552, opened 2026-05-25
fix(#161): re-parse criteria when LLM returns stringified JSON dict

### LINKED ISSUE #161: JudgeAgent AttributeError when LLM returns criteria as string instead of dict

drewdrewthis:
## Problem

`JudgeAgent` raised `AttributeError: 'str' object has no attribute 'values'` when an LLM returned the `criteria` verdict map as a **stringified JSON dict** instead of an inline dict (#161).

The first fix re-parsed the string but **fell back to `{}` on any parse failure**. That fallback was worse than the crash it replaced — and an audit found three more failure modes the original fix left open. All four are proven against the pre-fix code:

| Input `criteria` | Pre-fix behavior | Why it's bad |
|---|---|---|
| malformed JSON string | `success=True`, `passed=[]`, `failed=[]` | **false PASS — judge reports success having evaluated ZERO criteria** |
| string that parses to a non-dict (`"[1,2]"`) | `AttributeError` | **same crash class #161 claims to fix, just relocated** |
| partial dict (fewer verdicts than criteria) | unanswered criteria silently dropped, `success=True` | **false PASS — masks unevaluated criteria** |
| raw non-dict (a list, not a string) | `AttributeError` | crash |

A silent false-green in an eval/testing library is the worst outcome: a scenario that should fail ships as passing.

## Fix

- **Re-parse** a stringified criteria dict (the original happy path, kept).
- **Fail closed, never mask:** if `criteria` can't be resolved to a dict (unparseable string, string→non-dict, or raw non-dict), return `success=False` with an explanatory `reasoning` + a `logger.warning` — instead of swallowing to `{}`. The problem is surfaced, never hidden behind a fake success.
- **Map verdicts by schema key, not by position.** Extracted `_criteria_keys()` as the single source of truth shared by the schema builder and the parser, then look verdicts up by key instead of `enumerate(.values())`. This makes partial / reordered / nested / extra-key payloads robust — no mislabeling, no `IndexError`.
- **A criterion passes only on an explicit `"true"`** (JSON booleans `true`/`false` are coerced). Anything else (`false`, `inconclusive`, missing, or a nested/unexpected value) is a failure, so an unevaluated criterion can never slip through as success. Identical to the old behavior for complete, well-formed dicts.

### Design decision (flagged for review): surface-as-failure, not raise

When criteria can't be parsed, this fails **closed** (returns `success=False` with a clear reason) rather than **raising**. Rationale: this is an eval/testing library whose contract is to *produce a verdict*; a clean FAIL with an explanatory reason is more useful than aborting the run with a traceback, and it matches the existing discovery-leak handler in the same function. A false FAIL gets investigated; a false PASS ships broken code — so fail-closed is the safe default. Easy to switch to `raise` if the team prefers loud-abort semantics.

## Acceptance Criteria

| 1 | Stringified criteria dict is re-parsed and mapped correctly (no `AttributeError`) | `test_judge_reparses_stringified_criteria_dict` | ✅ |
| 2 | Normal inline-dict behavior unchanged (regression) | `test_judge_normal_dict_regression` | ✅ |
| 3 | Unparseable / non-dict criteria fail closed (no false success, no crash) | `test_judge_malformed_criteria_string_does_not_mask_as_success`, `test_judge_criteria_string_to_non_dict_fails_closed`, `test_judge_raw_non_dict_criteria_fails_closed` | ✅ |
| 4 | Partial dict cannot mask — unevaluated criteria block success | `test_judge_partial_criteria_dict_cannot_mask_success` | ✅ |
| 5 | Nested / extra keys don't crash; a nested per-criterion value is a failure | `test_judge_nested_extra_keys_are_ignored`, `test_judge_nested_criterion_value_is_failure` | ✅ |
| 6 | Success requires an explicit `"true"` for every criterion; verdict field is authoritative | `test_judge_normal_dict_regression`, `test_judge_partial_criteria_dict_cannot_mask_success`, `test_judge_failure_verdict_overrides_all_true_criteria` | ✅ |
| 7 | JSON-boolean criteria values are coerced (true→pass, false→fail) | `test_judge_boolean_criterion_values_are_coerced` | ✅ |

`uv run pytest tests/test_judge_agent.py -q` → **25 passed** (12 criteria + 10 pre-existing judge + 3 voice-modality merged from main). `uv run pyright scenario/judge_agent.py tests/test_judge_agent.py` → **0 errors, 0 warnings**.

## Proof of behavior (before → after)

All failure modes, exercised through `_parse_response` after the fix:

```
BEFORE fix: malformed=false-PASS, string->non-dict=CRASH, partial=false-PASS(masked), raw-list=CRASH
AFTER fix:
[OK] malformed string -> fail-closed: success=False
[OK] string->list     -> fail-closed: success=False
[OK] partial dict     -> no mask    : success=False
[OK] raw list         -> fail-closed: success=False
[OK] boolean true     -> coerced pass: success=True
[OK] boolean false    -> coerced fail: success=False
[OK] verdict=failure  -> overrides   : success=False
[OK] all-true success -> success     : success=True
```

## Notes / follow-ups (out of scope)
- **Duplicate-key collision** in `_criteria_keys`: two criteria that sanitize to the same key (via `[:70]` truncation or punctuation-only differences) collapse in the schema and share one verdict. **Pre-existing** (the old positional code had the same schema-side collapse), not a regression — but worth a dedicated issue since it's the same "unevaluated criterion slips through" hole. Surfaced by principles-reviewer.
- Repeated mock scaffold across the test file could be a shared autouse fixture (as in `test_judge_discovery_exhaustion.py`).

Closes #161.


drewdrewthis:

ACs: met — JudgeAgent now re-parses `criteria` when LLM returns a JSON string instead of a dict; graceful fallback to `{}` on parse failure; regression test `test_judge_agent_handles_criteria_as_json_string` verifies the exact payload shape that triggered the bug
Threads: zero unresolved
Note: LLM evaluator declined auto-approve (modifies runtime parsing logic in JudgeAgent) — human review required

Verified by:
  \`command gh pr checks 552\` → all checks pass or skipping; test (3.12) pass 7m31s, evaluate pass, javascript-complete pass
  regression test in worktree: \`PASSED\` (.worktrees/issue161)

drewdrewthis:
## Review verdict: READY

5 reviewers (principles, hygiene, security, test, proof) ran on the criteria-parse fix. No correctness blockers — proof-reviewer independently confirmed every AC maps to a real, passing, load-bearing test; principles-reviewer: "well-designed." All actionable findings were addressed.

**HEAD bump note:** marker is `2b47e8f61b110260096ee9984652c9a203bc45a4` (a clean `origin/main` merge integrating 18 upstream commits to clear a merge conflict that was blocking `python-ci`). The reviewed diff (`git diff origin/main...HEAD`) is **byte-identical** to the reviewed state — `origin/main`'s judge changes (#666/#667) don't touch `_parse_response`/criteria parsing. Re-verified at this HEAD: **25 tests pass**, 6/6 failure-mode demo OK, **pyright clean**, **CI `test (3.12)` green**. No re-run of the reviewer fan-out since the reviewed logic is unchanged.

### Addressed
- **[hygiene]** Removed dead `Optional` import.
- **[test]** Strengthened `test_judge_raw_non_dict_criteria_fails_closed` to pin `failed_criteria`.
- **[test]** Added `test_judge_failure_verdict_overrides_all_true_criteria` (verdict field authoritative).
- **[test][principles]** Added JSON-boolean coercion + `test_judge_boolean_criterion_values_are_coerced` (same #161 schema-nonconformance class; was a silent false-negative).
- **[principles]** Single key-zipped partition pass (drops index indirection).
- **[test]** Renamed `*_does_not_crash` → `*_fails_closed`.

### Non-blocking follow-ups (prose only)
- **[principles] New Issue** — duplicate-key collision in `_criteria_keys` (`[:70]` truncation / punctuation-only criteria collapse to one schema key). Pre-existing, not a regression.
- **[hygiene] Decide** — repeated mock scaffold across the test file could be a shared autouse fixture.

rogeriochaves (approved):

rogeriochaves:
why would it do that?
shouldn't the tool calling we pass force LLMs to always return the valid format and schema?

drewdrewthis:
The `strict: True` in the tool schema is an OpenAI-only guarantee — LiteLLM silently strips it for Vertex/Gemini, hosted vLLM, and WatsonX. Since `JudgeAgent` is provider-agnostic (any model via `scenario.configure(default_model=...)`), the structured-output guarantee only covers a subset of providers. Issue #161 is exactly this failure observed in practice — the model returned `criteria` as a stringified dict, triggering `'str' object has no attribute 'values'`. So the defensive re-parse is legitimate hardening, not dead code.

drewdrewthis:
The "first fix" this PR audits is @rajdeepmahal24's, from #196. His PR was open at the time and should have been the one merged. Recording it here so the history is accurate.

Issue #454, opened 2026-05-11
Voice agents: auto-restart stub bot on mid-suite crash


drewdrewthis:
**P3 nice-to-have.** Follow-up to epic #370.

## Problem

`python/tests/voice/conftest.py` spawns the bundled Pipecat stub bot on `:8765` session-scoped (atexit cleanup). If the bot crashes mid-suite — bad system prompt, OOM, model timeout, unhandled exception in the bot loop — every subsequent Pipecat-dependent test fails with `ConnectionRefused`, and the test runner reports them as test failures rather than infrastructure failures.

Not seen in CI yet, but the surface is brittle. Local long-running suites are the most likely failure mode.

## Ask

Add a liveness check + auto-restart on the existing `requires_pipecat_bot` fixture:

- Before each test that depends on the bot, probe `:8765` with a 0.5s connect timeout.
- If the probe fails, log a one-line "bot died, restarting" warning and re-spawn via the same subprocess path session-scope used.
- On restart, the fixture should fail the current test (mark as infrastructure error, not a bug-in-code failure) AND make the bot available for subsequent tests.

## Why low priority

We haven't hit it. The current path works for every suite run we've done. If we start running the workflow_dispatch e2e nightly OR doing long local runs, the probability goes up.

## Out of scope

- Bot health metrics (memory, request rate). Add only if we see flakes.
- A separate "bot supervisor" subprocess. The conftest already owns lifecycle; auto-restart is the smaller patch.

drewdrewthis:
Closing as delivered. Shipped in #557 (merged 2026-06-05). Verified on main @ ae0c921c: python/tests/voice/conftest.py:238-263 implements the stub-bot liveness check with auto-restart. Reopen if any scope remains.

PR #557, opened 2026-05-25
fix(#454): add liveness check + auto-restart to requires_pipecat_bot fixture

drewdrewthis:
## Summary

- `requires_pipecat_bot` now runs a 0.5s TCP probe before every Pipecat-dependent test.
- **Bot never started** (nominal path): auto-provisions the bot silently (unchanged behaviour).
- **Bot was running but has since crashed**: restarts it for subsequent tests, then calls `pytest.fail("[infrastructure] …")` on the current test so the operator sees it as an infrastructure failure, not a code bug.
- Subsequent tests after a crash-and-restart find a healthy bot and proceed normally.

## Acceptance criteria (from #454)

- [x] Before each test, probe `:8765` with a 0.5s connect timeout.
- [x] If probe fails and the bot was previously running, log a WARNING and restart.
- [x] Current test fails with `[infrastructure]` marker; subsequent tests run normally.
- [x] First-start path unchanged (no regression).

## Test plan

- [ ] CI passes (fixture is integration-scope; the structural change is validated by test collection)
- [ ] `pytest tests/voice/ --collect-only` succeeds without import errors
- [ ] `pyright .` — 0 errors

### DIFF
--- a/python/tests/voice/conftest.py
+++ b/python/tests/voice/conftest.py
@@ -158,9 +158,36 @@ def requires_llm():
 
 @pytest.fixture
 def requires_pipecat_bot():
-    """Ensure a Pipecat-compatible bot is on :8765, auto-starting if needed."""
-    if not _tcp_port_open("localhost", 8765):
-        _start_pipecat_bot()
+    """Ensure a Pipecat-compatible bot is on :8765, with liveness check + auto-restart.
+
+    If the bot was previously started but has since crashed, we restart it and
+    fail the current test so the operator knows the infrastructure hiccuped —
+    subsequent tests will find a healthy bot and run normally.
+    """
+    import logging
+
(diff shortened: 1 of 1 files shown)


drewdrewthis:

CI: green (zero failing, zero pending — re-run completed successfully after LLM flake on `test_user_is_hungry`, which is unrelated to this change)
ACs: met — liveness check fixture with auto-restart on dead bot between tests
Threads: zero unresolved

Verified by:
  `command gh api repos/langwatch/scenario/actions/runs/26391301495 --jq '.status,.conclusion'` → `completed / success`
  All 4 disconnect-logging tests in `tests/voice/conftest.py` area pass; the flake was in `examples/test_running_in_parallel.py::test_user_is_hungry` (LLM non-determinism, unrelated fixture)

PR #540, opened 2026-05-22
feat(typescript-sdk/#372): voice Pipecat adapter + g711 codec (PR10 of N)

drewdrewthis:
## Summary

- Ports `python/scenario/voice/adapters/{pipecat.py,_twilio_shared.py}` to TypeScript so voice scenarios can drive a running Pipecat bot over the Twilio Media Streams WebSocket protocol.
- Lands the g711 µ-law 8 kHz ↔ PCM16 24 kHz codec (pure-TS, no native binding) plus a linear-interpolation resampler — load-bearing math reused by the upcoming TS Twilio adapter (PR11).
- `transport="webrtc"` raises `PendingTransportError` at `connect()` time (SmallWebRTC negotiation is a follow-up).

## What ships

| File | Purpose |
| --- | --- |
| `javascript/src/voice/adapters/twilio-shared.ts` | g711 codec + 24k/8k resampler + Twilio Media Streams frame parser/builders. **Shared lib for PR11.** |
| `javascript/src/voice/adapters/pipecat.ts` | `PipecatAgentAdapter` — connects, sends 20 ms µ-law frames, decodes inbound, supports `interrupt()` via Twilio `clear`. |
| `javascript/src/voice/adapters/pending-transport-error.ts` | Shared error for deferred-transport adapters (parity with Python `_stub.PendingTransportError`). |
| `javascript/src/voice/adapters/__tests__/twilio-shared-codec.test.ts` | Codec round-trip + sample-rate-conversion scenarios + edge-case vitest tests. |
| `javascript/src/voice/adapters/__tests__/pipecat.test.ts` | Pipecat WS round-trip, WebRTC negative path, clear-buffer interrupt, plus receive-side coalescing edge cases — driven by a synchronous fake WebSocket (no real bot). |

### Capabilities advertised

```
streamingTranscripts: true
nativeVad:            true
dtmf:                 false
interruption:         true
inputFormats:         ["pcm16/24000", "mulaw/8000"]
outputFormats:        ["pcm16/24000", "mulaw/8000"]
```

## Bound scenarios

Adapter-specific secondary tag (`@ts-pipecat`, `@ts-codec`) follows the dual-axis pattern set by sibling-open PRs #535 (OpenAI Realtime) and #536 (ElevenLabs). Neither is on `main` yet — both PRs target `main` and use the same `@ts-<adapter>` convention.

| Tag | Scenario |
| --- | --- |
[...]

PR #534, opened 2026-05-21
feat(typescript-sdk/#372): voice Gemini Live adapter (PR9 of N)

drewdrewthis:
## Summary

- Ports `python/scenario/voice/adapters/gemini_live.py` → `javascript/src/voice/adapters/gemini-live.ts` using `@google/genai` v2.x.
- 2 `@unit` scenarios (connect + capabilities matrix) bound via vitest-cucumber + 1 `@e2e` demo scenario (env-gated on `GEMINI_API_KEY` / `GOOGLE_API_KEY`).
- `@google/genai` declared as an **optional peer dependency** — lazy-imported on `connect()` so the SDK ships without a hard Gemini coupling.

## What changed

**New files:**
- `javascript/src/voice/adapters/gemini-live.ts` — `GeminiLiveAgentAdapter`, capabilities matrix, PCM16 24kHz↔16kHz resampler (pure JS, no scipy), callback-to-queue bridge for `receiveAudio()`.
- `javascript/src/voice/adapters/__tests__/gemini-live.test.ts` — 2 cucumber-bound `@unit` scenarios, mocks `@google/genai` at module level so tests run offline.
- `javascript/examples/vitest/tests/voice/gemini-live.test.ts` — 1 cucumber-bound `@e2e` demo scenario, skips when no API key is present.

**Modified files:**
- `javascript/src/voice/index.ts` — exports `GeminiLiveAgentAdapter` and `GeminiLiveAgentAdapterInit`.
- `javascript/package.json` — adds `@google/genai ^2.5.0` as devDep + optional peer dep.
- `javascript/examples/vitest/package.json` — adds `@amiceli/vitest-cucumber` devDep so the e2e suite can `loadFeature`.
- `specs/voice-agents.feature` — splits the existing Gemini Live unit scenario into 2 (connect + capabilities matrix) and tags both `@ts-gemini-live`; tags the e2e demo scenario `@ts-gemini-live-e2e` to keep the two suites' tag filters disjoint.

## Decision pending: Gemini Live session lifecycle

**Choice landed: (b) — reuse `connect()` / `disconnect()` from `VoiceAgentAdapter`, mapping directly onto SDK lifecycle.**

The Python adapter holds the SDK context manager open across the adapter's lifetime via a background task because `google-genai`'s `client.aio.live.connect(...)` is exposed as an `async with` context manager — there is no plain open/close API in Python. JS has no such constraint:

[...]

PR #528, opened 2026-05-21
feat(typescript-sdk/#372): voice-aware UserSimulatorAgent + judge + audio messages (PR4 of N)

drewdrewthis:
## Summary

- Ports `python/scenario/voice/messages.py` → `javascript/src/voice/messages.ts`: `createAudioMessage`, `extractAudio`, `messageHasAudio` helpers using the local `AudioMessageParam` type (Decision 2(b) — no `openai` package import)
- Extends `UserSimulatorAgent` with voice path: when `voice` config is set, synthesizes audio via TTS and emits `AudioMessageParam`; supports per-step `voiceStyle`/`audioEffects` overrides and `persona` composition
- Extends `JudgeAgent` with voice-aware helpers: `conversationHasAudio` (static), `effectiveIncludeAudio`, `effectiveIncludeTimeline`, `effectiveIncludeTraces`; auto-detects multimodal model capability; `includeAudio: false` escape hatch

## What changed

**New files:**
- `javascript/src/voice/messages.ts` — `createAudioMessage`/`extractAudio`/`messageHasAudio`; WAV encode/decode inline (no external dep); no `openai` import
- `javascript/src/voice/__tests__/messages.test.ts` — plain vitest unit tests for boundary cases
- `javascript/src/agents/__tests__/user-simulator-voice.test.ts` — 5 `@ts-simulator` cucumber scenarios
- `javascript/src/agents/judge/__tests__/judge-voice.test.ts` — 7 `@ts-judge` cucumber scenarios
- `javascript/src/agents/__tests__/voice-assistant-role.test.ts` — 1 `@ts-assistant-role` cucumber scenario

**Modified files:**
- `javascript/src/voice/index.ts` — exports `createAudioMessage`, `extractAudio`, `messageHasAudio`
- `javascript/src/agents/user-simulator-agent.ts` — voice path + `setOneShotOverride` + `_synthesize` stub (wired by PR2 TTS module); `stripAudioContent` keeps LLM calls text-only
- `javascript/src/agents/judge/judge-agent.ts` — `JudgeAgent` exported as class; `includeAudio`/`includeTimeline`/`includeTraces` config fields; voice-aware helper methods; `modelSupportsAudio` substring check
- `specs/voice-agents.feature` — 13 scenarios tagged with `@ts-simulator` / `@ts-judge` / `@ts-assistant-role`

## Test plan

- [ ] `pnpm -C javascript test src/voice/__tests__/messages.test.ts` — messages unit tests
[...]

PR #515, opened 2026-05-21
feat(typescript-sdk/#372): voice adapter runtime + executor wiring + VAD fallback (PR3 of N)

drewdrewthis:
PR3 of N for langwatch/scenario#372. Builds on PR1 (#511) which shipped types only. Status: **draft — keep on lead/Drew to flip ready-for-review after verification.**

## Summary

- Port `python/scenario/voice/adapter.py` runtime into `javascript/src/voice/adapter.runtime.ts`:
  * `asyncio.Event` → `AgentSpeakingEvent` (Promise + external resolve).
  * `async with adapter:` → explicit `startVoiceAdapters` / `stopVoiceAdapters` the executor calls once per scenario.
  * Default `call()` body: send → drain on tail silence → record → return `AudioMessageParam`.
  * Hook fan-out for `onAudioChunk` / `onVoiceEvent` from PR1's `VoiceExecutorState` surface.
- Port `python/scenario/voice/vad.py` → `javascript/src/voice/vad.ts`:
  * `WebRTCVadFallback` activates only when `adapter.capabilities.nativeVad === false`.
  * One-shot warning per adapter name via `console.warn`, gated by a static `Set<string>` (mirrors Python `_warned_adapters` memoisation; no rate-limit regression — see Python commit history for the bug class this avoids).
  * Pure-TS RMS energy + hysteresis detector ships today; see **Decision pending** below for the long-term webrtcvad C library question.
- Patch `javascript/src/execution/scenario-execution.ts`:
  * `ScenarioExecution implements VoiceExecutorState` structurally (Decision 1(b) from #372 — `voiceRecording`, `voiceTimeline`, `voiceLatency`, `voiceRecordingStartedAt`, `onVoiceEvent`, `onAudioChunk`).
  * `pickVoiceAdapters(this.agents)` snapshot taken at run start; `connect()` runs **inside** the try block, `disconnect()` runs in the finally so the spec line-143 "regardless of pass/fail/exception" contract holds.
  * Wires `onAudioChunk` / `onVoiceEvent` from new `ScenarioConfig` fields.
- Add `javascript/src/voice/__tests__/fixtures/fake-adapter.ts` — in-memory `FakeVoiceAdapter` with no real transport. Tests use it exclusively.

PR3 deliberately does NOT touch TTS/STT (PR2), simulator or judge (PR4), script steps or result extensions (PR5), effects (PR6), or real transports (PR7+).

## Acceptance criteria (the five checks)

[...]

PR #560, opened 2026-05-26
docs(#372): voice internal design record + ADR-002 (per-run provider state)

drewdrewthis:
## What

The **Engineering Design Record** for the TypeScript voice port (#372) — the *inside-the-box* design the [API proposal PRD](https://www.notion.so/langwatchdata/Claude-Opus-1-Proposal-3125e165d48280eca9dbe044884213c8) never specified. Spec only; no runtime change.

Two files (repo-root `docs/`, alongside the existing ADR-001):
- **`docs/voice/internal-design.md`** — module tree + per-module contract catalog (§0.1: *target* API vs *as-built* vs *gap*, ~40 modules), the 4 stable host seams voice plugs into, the per-run state model, lifecycle ownership, data model, public/internal seam, the salvage→clean-stack plan, and the runnable e2e verification gate.
- **`docs/adr/002-voice-provider-state.md`** — ADR-002: provider state is **per-run on `ScenarioConfig.voice`**, not a module-global singleton.

## Why

The voice work landed as **flat-sibling PRs** (all forked one point off `feature/372-voice-ts-parity`), so no PR could see the others' contracts. That produced, concretely:
- a **module-global STT provider** singleton (violates the repo's own [ADR-001](../blob/main/docs/adr/001-scenario-concurrency-model.md) "no process-wide state mutation" — unsafe for concurrent runs);
- an **invented `scenario.configure({ stt })`** API (in no merged code, not in Python);
- **3 disjoint forks of `adapter.ts`**, duplicate `STTProvider`/`synthesize` with divergent signatures, two divergent `twilio-shared.ts`;
- a **live audio-format bug**: two `createAudioMessage` producers disagree (`format:"wav"` WAV-wrapped vs `format:"pcm16"` raw PCM) and their extractors mis-decode each other.

The PRD fixed the *public API*; nobody specified the *internals*, so they were improvised. This doc is that missing layer — and the spec the clean voice stack is built against.

## Key decisions (locked)

- Public API stands; **only API delta = remove `configure({stt})`**. User stories hold.
- Provider state → **per-run `ScenarioConfig.voice`** (the only per-run carrier that reaches `AgentAdapter.call()`; `RunOptions` is boundary-only). ADR-001's no-global principle, correct carrier.
- **One in-message audio format** (AI-SDK `file` parts) — fixes the live mis-decode bug.
- **One file per provider**; dedicated `voice/steps.ts`; symmetric `voice/tts/elevenlabs-tts.ts` leaf.
- Execution: **fresh clean dependency stack** against current `main`, salvaging the 10 open sibling PRs as code reference, then closing them. Runnable e2e top-of-stack (incl. a round-trip audio assertion catching the format bug) is the "done" gate.

## Accepted obligation

TS internals get corrected while Python still carries the same global + phantom `configure(stt=)` docstring → **a Python parity follow-up is owed** (don't close the #372 epic without it).

## Status

Verified against current `main` (`c8cca4e`). #511 (PR1) is already merged into base; the clean stack builds on it. The 10 open voice PRs (#513, #515, #528, #534–#540) will be **closed** once the clean stack lands.

### DIFF
--- a/docs/adr/002-voice-provider-state.md
+++ b/docs/adr/002-voice-provider-state.md
@@ -1,0 +1,128 @@
+# ADR-002: Voice Provider State — Per-Run, Not Global
+
+**Date:** 2026-05-26
+
+**Status:** Proposed
+
+**Committed at:** repo-root `docs/adr/002-voice-provider-state.md` (alongside ADR-001 at `docs/adr/001-scenario-concurrency-model.md`; not under `javascript/docs/`).
+
+## Context
+
+The voice feature (#372) introduces STT (speech-to-text) and TTS (text-to-speech)
+providers that the framework must select and configure. The PR2 implementation (#513)
+installed these as **module-global singletons**:
+
+```ts
+// javascript/src/voice/stt.ts (PR #513)
+let provider: STTProvider | null = new OpenAISTTProvider();
+export function setSttProvider(p: STTProvider | null): void { provider = p; }
+export function getSttProvider(): STTProvider | null { return provider; }
+```
+
+This is a faithful port of the Python implementation (`python/scenario/voice/stt.py`),
+which uses the same module-global `_provider` + `set_stt_provider()`/`get_stt_provider()`.
+
+**The problem:** process-wide mutable provider state is not safe under concurrent runs.
+Two scenarios running in parallel with different STT providers clobber each other's
+global — the last `setSttProvider()` wins for both. This is the exact failure mode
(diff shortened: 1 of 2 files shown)


drewdrewthis:
Folded into the consolidated TS voice stack PR #561 — `internal-design.md` + ADR-002 ship there alongside the implementation they specify, so design + impl land atomically. No content lost. The Python-parity obligation is now tracked as a dedicated follow-up issue (linked below / from the epic #370).

PR #513, opened 2026-05-21
feat(typescript-sdk/#372): voice TTS + STT plumbing (PR2 of N)

drewdrewthis:
## Summary

PR2 of N for [#372](https://github.com/langwatch/scenario/issues/372) — ports `python/scenario/voice/{tts,stt,_transcribe}.py` to TypeScript and exposes `scenario.configure({ stt })` for swapping the default STT provider.

Builds on PR1 (#511 — types-only contract surface). No adapter runtime, no VAD, no simulator/judge wiring — those land in PR3+.

## Scope

- `javascript/src/voice/tts.ts` — `synthesize(text, voice, effectFn?)` with LRU(64) keyed on `sha256(text)+voice`. Effects apply **after** cache hit per the locked decision; raw text never reaches the cache payload. `registerTtsProvider({ prefix, synth })` for custom backends. Default OpenAI provider lazy-imports the SDK so users with custom providers don't need an `OPENAI_API_KEY`.
- `javascript/src/voice/stt.ts` — `STTProvider` interface (`transcribe(audio: AudioChunk): Promise<string>` only), `OpenAISTTProvider` (default = `gpt-4o-transcribe`) with 25-minute chunking, `ElevenLabsSTTProvider`, `setSttProvider` / `getSttProvider`. Pure-TS `pcm16ToWav` encoder — no `audioop`/ffmpeg dep for transcription.
- `javascript/src/voice/transcribe.ts` — `transcribeSegments`, post-hoc, idempotent per-segment, degrades gracefully when no provider is configured.
- `javascript/src/config/configure.ts` — `scenario.configure({ stt })` entry point. Wired into the top-level `scenario` object so `scenario.configure({ stt: new MyProvider() })` works as in Python.

## Acceptance checks

- [x] `pnpm -C javascript build` green — commit `05d549d`, dist outputs CJS+ESM+DTS without errors (228 KB CJS bundle, 9.5s DTS).
- [x] `pnpm -C javascript test` green — **397 tests passed across 25 files** (3.35s). Voice + config subset: 37 tests across 5 files (502 ms).
- [x] Cache key is `sha256(text)+voice`; effects-after-cache invariant proven by test — `tts.test.ts` "applies effectFn AFTER cache hit" runs synthesis once and re-reads from cache for two different effects, then asserts the *third* effect output equals `original.reverse()` (NOT `boosted.reverse()`), proving effects never bake into stored audio.
[...]

Issue #371, opened 2026-04-20
Voice agents: remaining platform adapter transports


drewdrewthis:
Follow-up to epic #370. #350 ships with real transports for Twilio, Pipecat, OpenAI Realtime, Gemini Live, and ElevenLabs. This issue covers the remaining platform adapters whose classes, capability matrices, and constructor validation exist in the core PR but whose `send_audio` / `recv_audio` methods raise `PendingTransportError`.

## Scope

One real transport each. Users should be able to replace a `PendingTransportError`-raising call with a working transport and run the same scenarios they ran against Twilio / Pipecat, without any other SDK changes.

- [ ] **LiveKit** (`LiveKitAgentAdapter`) — join room as participant, bidirectional audio via LiveKit Rooms.
- [ ] **Vapi** (`VapiAgentAdapter`) — create call + connect via `websocketCallUrl`.
- [ ] **WebRTC** (`WebRTCAgentAdapter`) — generic WebRTC via `aiortc` direct. Do not pull in pipecat's SmallWebRTC wrapper — keep transitive deps minimal. ICE / TURN / DTLS-SRTP handling is on the implementer.
- [ ] **WebSocket** (`WebSocketAgentAdapter`) — reference `WebSocketProtocol` implementation plus one working end-to-end example (e.g., a minimal echo protocol or a wrapper around Twilio Media Streams framing as a reference point).

## Acceptance criteria

For each adapter:

- Real bidirectional audio transport over the adapter's declared wire format, converting to/from the SDK's internal PCM16 @ 24kHz mono.
- Capability matrix (`adapter.capabilities`) accurately reflects what the live transport supports — no claiming `streaming_transcripts=True` unless incremental transcript tokens actually stream.
- At least one integration test in `python/tests/voice/` exercises the transport under the nightly `@integration` workflow.
- The `PendingTransportError` raise-site is removed.

## Out of scope (covered in other sub-issues)

- TypeScript parity — #372.
- Feature-file-to-test traceability map — #373.

## Why deferred from #350

Twilio, Pipecat (WebSocket), OpenAI Realtime, Gemini Live, and ElevenLabs cover the most-requested realtime voice providers. The adapters in this issue have either lower usage demand (LiveKit, Vapi) or higher implementation complexity (WebRTC's ICE/TURN/DTLS surface). Deferring them unblocks the core SDK contract without silent failure — each stub raises `PendingTransportError` with a message pointing users at the capability matrix page.

Issue #370, opened 2026-04-20
Voice Agents


drewdrewthis:
Umbrella issue for voice agent support in the Scenario Python SDK.

## Goal

Voice is a first-class citizen of `scenario.run()` — same entrypoint, same script DSL, same judge. What changes is the medium, not the paradigm.

The contract: a Scenario user already writing text scenarios should be able to switch one or more agents to audio adapters, add optional script steps (`scenario.audio()`, `scenario.sleep()`, `scenario.dtmf()`, `scenario.interrupt()`), and have the same test harness, user simulator, and judge work against the voice transport. No `scenario.voice.run()`, no separate paradigm.

## Sub-issues

- #350 — Voice Agents: Python SDK foundation. Feature-complete for Python, with real transports for Twilio, Pipecat, OpenAI Realtime, Gemini Live, ElevenLabs.
- #371 — Remaining platform adapter transports: LiveKit, Vapi, WebRTC, generic WebSocket.
- #372 — TypeScript SDK parity for voice agents.
- #373 — Maintained feature-file-to-test traceability map (nice-to-have, low priority).

## Delivery bar

For closure of #350 (the foundation issue) specifically: @drewdrewthis should be able to spin up and use scenario in Python for all voice-agent features end-to-end. Additional platform adapter transports can be deferred to follow-up sub-issues, but each deferral must be documented with a clear error path (`PendingTransportError`) pointing users at an adapter capability matrix.

## Scope (shared across sub-issues)

- Core: `VoiceAgentAdapter` base with capability matrix, `AudioChunk` (PCM16 @ 24kHz mono internal format), voice-aware `UserSimulatorAgent` (TTS pipeline with per-step overrides), voice-aware `JudgeAgent` (multimodal pass-through with transcript fallback)
- Script DSL extensions: `scenario.audio()`, `scenario.sleep()`, `scenario.silence()`, `scenario.dtmf()`, `scenario.interrupt()` (time- and word-based), `scenario.background_noise()`, `agent(wait=False)`, `proceed(interruptions=...)`, `proceed(on_turn=...)`
- Result extensions: `result.audio`, `result.timeline`, `result.latency`
- Hooks: `on_audio_chunk`, `on_voice_event`
- Audio effects: bundled CC0 noise samples, static, distortion, volume jitter, packet loss, jitter, codec quality
[...]

PR #561, opened 2026-05-27
feat(typescript-sdk): voice agent testing — consolidated clean stack

### LINKED ISSUE #372: Voice agents: TypeScript SDK parity
Follow-up to epic #370. The Python SDK (#350) ships voice-agent support; TypeScript parity under `javascript/` is deferred to this issue.

## Scope

Port the full voice-agent surface to the TypeScript SDK:

- `VoiceAgentAdapter` base + `AdapterCapabilities` + `UnsupportedCapabilityError`
- `AudioChunk` internal format (PCM16 @ 24kHz mono) with PCM16-odd-byte invariant
- Voice-aware `UserSimulatorAgent` (TTS pipeline, per-step voice / effects overrides)
- Voice-aware `JudgeAgent` (audio transcript auto-attach, multimodal pass-through with transcript-only fallback, `include_audio` cost escape)
- Script steps: `sleep`, `silence`, `audio`, `dtmf`, `interrupt` (time- and word-based), `agent(wait: false)`, `proceed({ interruptions })`, `proceed({ onTurn })`, `background_noise`
- Result extensions: `result.audio`, `result.timeline`, `result.latency`
- Hooks: `on_audio_chunk`, `on_voice_event`
- Effects module: bundled CC0 noise samples, `background_noise`, `static`, `distortion`, `volume_jitter`, `packet_loss`, `jitter`, `codec_quality`, `distance`, `doppler`
- Platform adapters: at minimum the five with real transports from #350 — Twilio, Pipecat, OpenAI Realtime, Gemini Live, ElevenLabs
- Pluggable STT interface with OpenAI default and swap via `scenario.configure({ stt })`
- SDK-side VAD fallback for adapters without native VAD

## References

- Python implementation: `python/scenario/voice/` in the main branch once #350 lands — this is the source-of-truth reference for behavior.
- Existing JS voice reference code: `javascript/examples/vitest/tests/helpers/openai-voice-agent.ts`. Do NOT port its `forceUserRole` workaround — fix at the type level per the #350 adaptability note (audio must work cleanly in any message role: user, assistant, tool).

## Acceptance criteria

- The 83-scenario feature contract (which ships with #350 as `specs/voice-agents.feature`) is achievable against the TS SDK — same inspection-based mapping from Scenario to test as in Python.
- All five primary platform adapters (Twilio, Pipecat, OpenAI Realtime, Gemini Live, ElevenLabs) have real transports, not stubs.
- CI green on the `javascript-ci` workflow with voice tests included.

## Why deferred

`#350` is already a large PR (~12k LOC added). Python-first is the locked-in delivery decision. TS has its own SDK ecosystem differences (different adapter APIs in some cases, different cache semantics, different runtime — Node vs browser) worth a dedicated PR scoped against this issue rather than mixed in.

### LINKED ISSUE #571: ci: evaluate workflow hard-fails on PRs >20k-line diff instead of its oversized path
The `evaluate` job fails (non-required) on large PRs like #561: `gh pr diff` hits GitHub's 20,000-line API cap → HTTP 406 → exit 1, BEFORE the workflow's own `DIFF_LIMIT` char-check (which has a graceful `oversized=true` path). The line-limit 406 bypasses the graceful path. Make the diff-fetch handle the 406/too-large case via the existing oversized path so big PRs do not show a spurious red `evaluate`. Surfaced by #561 (199 files / ~27k insertions).

### LINKED ISSUE #574: refactor(voice): naming, dedup, and comment-cleanup sweep on the consolidated voice stack
## Context
Reviewer sweep on the consolidated TS voice stack (PR #561) surfaced a cluster of small naming + dedup smells that are cheap to fix and improve readability. None block the merge; rolling them into one PR keeps churn low.

**Effort**: S — follow-up from PR #561. Not blocking.

## Items
1. **Rename `VoiceEventSpeakingBoundary` → `VoiceEventTimestampOnly`** (or similar). The current name is misleading: the union covers `agent_stop_speaking` too — all events whose payload is just a timestamp, not just "speaking" boundaries. Cite: `javascript/src/voice/recording.types.ts:61` (interface declaration) and `javascript/src/voice/recording.types.ts:123` (union member).
2. **Collapse `pipecat` adapter's `interrupted` + `discardingInboundAudio` into one logical state.** They're always set/cleared together (`javascript/src/voice/adapters/pipecat.ts:189`, `:203`, `:347-348`, `:448`, `:452`); the dual-flag scheme leaks implementation detail. Extract a small enum-or-struct like `interruptionState: { phase: "idle" | "interrupted" | "discardingFrames" }` or a single flag with the receiver checking both consumer points explicitly.
3. **Drop redundant `bargeInDelayMs && bargeInDelayMs > 0` guard.** At `javascript/src/execution/scenario-execution.ts:1729` — `bargeInDelayMs > 0` already short-circuits the `0`/`undefined` case (`undefined > 0` is `false`). Reduces to `bargeInDelayMs > 0`.
4. **Make the `delaySeconds * 1000` truncation consistent across the two sleep call sites.** Currently `scenario-execution.ts:1501` does `Math.floor(delaySeconds * 1000)` while `:1598` does `delaySeconds * 1000` (no floor). Pick one — `setTimeout` accepts fractional ms in node and clamps anyway, but the inconsistency reads as "one was forgotten." Likely the intent was to floor only where storing on the executor (`interruptBargeInDelayMs`), in which case add a one-line comment justifying the asymmetry. Otherwise floor both.
5. **Trim `pipecat.ts interrupt()` JSDoc.** Currently a multi-paragraph essay (`javascript/src/voice/adapters/pipecat.ts:423-431`). Reviewer felt it over-explains for a method whose `interrupt()` name already conveys intent. Keep the "side effects" enumeration + the WS-frame-discard rationale; drop the higher-level "what is an interrupt" preamble.
6. **Replace `// A future fix would...` comment with a tracked issue link.** At `javascript/src/voice/adapters/gemini-live.ts:416` — once the follow-up issue for iterator-restart parity is filed, swap the prose for `// TODO(#XXX): see follow-up — replicate Python iterator-restart semantic.`

## Suggested fix shape
Single PR, six focused commits (or one squash commit with itemized message). All edits are in-file and need no public-API motion.

## Why it matters
Each individually is too small to file alone; together they're a 30-minute cleanup that improves the readability of the consolidated stack landing in #561.

drewdrewthis:
## Why

Closes #372 (epic #370). The TypeScript voice-agent-testing port had fragmented into **10 flat-sibling PRs** (#513/#515/#528/#534–540) carrying pre-EDR design drift. This replaces all of them with **one clean stack** rebuilt against `main`, conforming to the EDR (#560 / ADR-002) and the decided public API (PRD). Python is the reference implementation; this brings TS to parity.

## Acceptance Criteria

The behavioral contract is **[`specs/voice-agents.feature`](https://github.com/langwatch/scenario/blob/voice/372-refactor/specs/voice-agents.feature)** — 127 scenarios ported from the Python source-of-truth (79 `@unit`, 14 `@integration`, 39 `@e2e`, 2 `@todo`). Collapsed to the headline ACs below, each ticked only with evidence on this branch. "Done" = (1) the `ci-checks` gate (units + build + non-bot `@ts-e2e`) green, (2) the committed demo recordings present, **and (3) the app-side ACs below (AC13–18) verified — a voice run correctly ingested, queryable, and rendered in the langwatch app.** SDK-green alone is necessary but not sufficient.

**Status legend:** ✅ = evidence verified in-tree/observed · ⏳ = implemented + author-reported green locally, **not yet confirmed by CI** · ◻️ = open. *(2026-06-04: the unit suite now runs in CI — `ci-checks (24.x)` green on [run 26960103187](https://github.com/langwatch/scenario/actions/runs/26960103187) after the rebase + lockfile fix; formerly-⏳ unit-backed rows are CI-confirmed.)*

**Binding status:** 29/127 scenarios carry `@ts-bound` (wired to an executable TS test). The remaining 98 are contract-level (ported from the Python spec), covered indirectly rather than by 1:1 bindings. The full suite is **CI-confirmed at 796 pass / 1 skip** on HEAD `390e52c` ([run 26960103187](https://github.com/langwatch/scenario/actions/runs/26960103187)) — the former lockfile abort is fixed, so unit-backed ACs below are ✅ on real CI evidence. Demo-recording-backed ACs are ✅ (the 16 recordings are committed and present in-tree — verified).

| # | Acceptance criterion | Evidence | Status |
|---|---|---|---|
| AC1 | **Same entrypoint** — voice uses `scenario.run()`, text-only scenarios unaffected by voice deps | `voice_text_parity` recording (in-tree ✅) + unit "Existing text-only scenarios unaffected" (✅ CI) | ✅ |
| AC2 | **Per-run provider state** (ADR-002) — voice config on `ScenarioConfig.voice`, no module-global `configure()` | `composable_stt_swap` recording (in-tree ✅); STT-swap units (✅ CI) | ✅ |
| AC3 | **Barge-in primitive** — `agent({ wait:false })` + `interrupt()` cuts off a reply mid-utterance, marked truncated | `interruption_recovery` recording, judge hard-gated (in-tree ✅) | ✅ |
| AC4 | **Real server-VAD barge-in** on Gemini Live + ElevenLabs (mid-stream cut-off) | `gemini_live_interruption` + `elevenlabs_interruption` recordings (in-tree ✅) | ✅ |
| AC5 | **Adapter parity** — OpenAI Realtime (agent+user), ElevenLabs (hosted/branded/composable), Gemini Live, Pipecat | `openai_realtime_{agent,user}`, `elevenlabs_{hosted,branded}`, `gemini_live`, `pipecat_{scenario,ws}` recordings (in-tree ✅) | ✅ |
| AC6 | **Audio effects + tonal realism** — noise floor, distortion, audible anger | `angry_customer` recording, noise-floor≫silence assertion (in-tree ✅) | ✅ |
| AC7 | **Voice-aware judge** — auto-detects audio, transcript fallback for non-multimodal, structured timeline | `@ts-judge` 7-scenario unit suite — **CI-confirmed** ([run 26960103187](https://github.com/langwatch/scenario/actions/runs/26960103187)) | ✅ |
| AC8 | **Capability matrix is a contract** — every adapter publishes `capabilities`; `dtmf()` raises `UnsupportedCapabilityError` off-telephony; matrix in docs | `@ts-contract-surface` units (✅ CI) + `adapters/*.mdx` tables (in-tree ✅) | ✅ |
| AC9 | **PCM16 @ 24kHz mono internal format**; pluggable STT (default OpenAI `gpt-4o-transcribe`); SDK-side VAD fallback with one-shot warning | `@ts-contract-surface` / `@ts-stt` / `@ts-vad` units — **CI-confirmed** (same run) | ✅ |
| AC10 | **CI merge gate green** — `ci-checks` (units + build + non-bot `@ts-e2e`) passes on HEAD | ✅ `ci-checks (24.x)` **pass** (6m2s, suite executed) + `javascript-complete`/`docs-complete`/`python-complete`/`evaluate` all pass on HEAD `390e52c` — [run 26960103187](https://github.com/langwatch/scenario/actions/runs/26960103187). Fixed by rebase onto main + recording the `@ungap/structured-clone` override in `pnpm-lock.yaml` (`f7273e7`) + 2 stale doc-contract test assertions (`f2cdf58`). Local: 796 pass / 1 skip. | ✅ |
| AC11 | **Telephony transport (Twilio)** — TwiML endpoint, signature rejection, tunnel harness, clear-buffer interrupt | `@ts-twilio-proto` + `@ts-twilio-server` integration scenarios bound — unit/integration layers **CI-confirmed** (same run) | ✅ |
| AC12 | **Remaining platform adapters** (LiveKit, Vapi, generic WebRTC/WebSocket) raise `PendingTransportError` | stub adapters + units (✅ CI); full transports deferred to #371 | ✅ |

**Net (honest read, updated 2026-06-04):** **all SDK-side ACs (AC1–12) are now ✅.** AC1–6 were demo-recording-verified; AC7–9 + AC11 flipped from ⏳ to ✅ when the unit suite executed in CI for the first time on this PR — `ci-checks (24.x)` pass in 6m2s on HEAD `390e52c` ([run 26960103187](https://github.com/langwatch/scenario/actions/runs/26960103187)), 796 pass / 1 skip. AC10's lockfile blocker is fixed (rebase onto main + `@ungap/structured-clone` override recorded in the lockfile + 2 stale doc-contract test assertions aligned).

### App-side ACs — langwatch ingestion + rendering (SDK-green is necessary but NOT sufficient)

Parity is not done when the SDK suite passes — it's done when a voice `scenario.run()` is **correctly ingested, queryable, and rendered in the langwatch app**. These require a live end-to-end run (SDK → langwatch ingest → API/UI), not a unit test. Surface mapped against `langwatch/langwatch`:

| # | App-side acceptance criterion | Concrete check | Status |
|---|---|---|---|
| AC13 | **Traces received** — a voice run's OTel trace (incl. `input_audio` content parts) lands in langwatch | `POST /api/collector` ingests the span; trace appears for the project | ✅ verified 2026-06-04 — live TS run `scenariorun_3EfWWrSEiX95SWhzwUCwd7n22OA` (`openai_realtime_agent` demo) landed **2 traces** in `project_bZspxwkhCD4POvqmIgOr2` with `scenario.run_id` metadata (`sdk: langwatch-observability-sdk typescript 0.16.1`); judge spans carry the audio as file parts: `{"type":"file","mediaType":"audio/pcm16",…~230KB}` |
| AC14 | **Traces queryable via API** — the received trace is retrievable through the public API | `POST /api/traces/search` `{projectId, filters}` returns the voice run's trace | ✅ `POST /api/traces/search` (startDate/endDate window) returned both: `c50549165a184a73d5fb509525230755`, `7c45a3b4edf40ad1466e33afd176b2f5`; `GET /api/trace/:id` returns full span detail (5 spans incl. `OpenAIRealtimeAgentAdapter.call`, `_JudgeAgent.call`) |
| AC15 | **Scenario events ingested** — `RUN_STARTED` / `MESSAGE_SNAPSHOT` / `RUN_FINISHED` persisted, each message carrying optional `trace_id` | `scenario_events` ES index holds the run's events; queryable via tRPC | ✅ run persisted with **4 messages** (`MESSAGE_SNAPSHOT`) carrying 8 `input_audio` content parts; the 2 assistant messages carry `trace_id` refs **matching the two traces above** (cross-linkage proven); `status: SUCCESS` + `results.verdict: success` (3/3 criteria) = `RUN_FINISHED` recorded |
| AC16 | **Scenario run visible in the app** — the voice run shows in the simulations UI + REST | `GET /api/simulation-runs/:scenarioRunId` returns it; `/[project]/simulations` renders it | ✅ `GET /api/simulation-runs/scenariorun_3EfWWrSEiX95SWhzwUCwd7n22OA` → **HTTP 200** (`demo_openai_realtime_agent`, 41.8s, cost $0.0018); `platformUrl: app.langwatch.ai/scenario-tracing-bZspxw/simulations` (API-verified; in-app visual spot-check needs an authed browser — one click) |
| AC17 | **Audio messages rendered** — `input_audio` content renders as an inline player, not a raw JSON blob | `<ScenarioMessageRenderer>` → `<MediaPart>` emits `<audio controls>` for the message | ✅ shipped on `langwatch@main` via [#4058](https://github.com/langwatch/langwatch/pull/4058) (the old [#3781](https://github.com/langwatch/langwatch/pull/3781) is stale/superseded — not a blocker). `MediaPart.tsx` emits `<audio data-testid="media-part-audio" controls>` with `onLoadedData` probing; `visit-content-part.ts` handles `input_audio` parts; integration-tested (`MediaPart.integration.test.tsx`). The verified run's messages carry exactly the shape it renders: `{"type":"input_audio","input_audio":{"mimeType":"audio/pcm16","url":"/api/files/so_…"}}` |
| AC18 | **Audio itself verified** — the rendered audio actually decodes + plays the correct content | base64 WAV decodes in `<audio>` (`onLoadedData` → ok); externalized blobs resolve via `/api/files/:id` | ✅ verified byte-level 2026-06-04 on run `scenariorun_3EfWWrSEiX95SWhzwUCwd7n22OA`: `GET /api/files/so_000000000002CR60L0eAWXH3ILtke` → HTTP 200, 249,600 bytes; decodes as valid **5.20s pcm16 @ 24kHz mono** (the AC9 contract format); OpenAI STT transcribes it to *"Of course! Where would you like to go, or what kind of activities are you interested in?"* — the coherent reply to the run's user turn *"Hello, can you help me plan a weekend trip?"*. The user-side part (the SDK's other audio path — user-sim TTS) round-trips **verbatim**: 3.60s pcm16 → STT → *"Hello, can you help me plan a weekend trip?"*, the exact scripted turn. Decode ✓ format ✓ correct-content ✓ both-paths ✓ serving-route ✓; the in-browser `<audio>` element rendering these same URLs is covered by the integration tests above. Follow-up hardening on-branch: `885d294` WAV-wraps raw pcm16 at the langwatch-bound converter (supersedes `5db36c8`; e2e-verified in its commit: `/api/files` now serves `audio/wav` with a RIFF header, ffprobe-decodable) + `390e52c` aligns the conversion test |

**Verification matrix (2026-06-04, all live against production `app.langwatch.ai`):**

Every run below links to the **live prod app** — open it (project members) and **press play on any audio message** to hear the run the table describes.

| Family / variant | Run | Ingest (AC13–15) | REST (AC16) | Audio bytes (AC18) |
|---|---|---|---|---|
| OpenAI Realtime (agent) | [`scenariorun_3EfWWrSEiX95SWhzwUCwd7n22OA`](https://app.langwatch.ai/scenario-tracing-bZspxw/simulations/default/scenariobatch_3EfWOgskU31YbFeOaycyeK0zEJp) | ✅ 2 traces x-linked, 4 msgs / 8 parts | ✅ 200, SUCCESS 3/3 | ✅ agent reply STT-coherent + user TTS **verbatim** round-trip |
| Gemini Live + **interruption** | [`scenariorun_3Efeze6fAvYmZd8gQeyVxz8SoOm`](https://app.langwatch.ai/scenario-tracing-bZspxw/simulations/default/scenariobatch_3EfexikiUFJPBlyfU5hdsSl1zPF) | ✅ 3 traces, 5 msgs / 5 parts — **FAILED verdict persists correctly** (2 met / 1 unmet). The unmet criterion is a judge semantic trap, not an adapter fault: the criterion says *"over a **real** Gemini Live session"* and the judge read the framework's own `origin: simulation` trace metadata as disqualifying — while its reasoning simultaneously **confirms** the live mid-utterance VAD cut-off. Vitest mechanics assertions: 3/3 pass. Suggested post-merge polish: drop the word "real" from that criterion | ✅ 200 | ✅ truncated segment is its own cut-off transcript (`"I am a large"` / STT `"I am a lar"`, 0.96s) |
| ElevenLabs hosted | [`scenariorun_3EffH3ID8Q7HU5S5anb2c8u4hhy`](https://app.langwatch.ai/scenario-tracing-bZspxw/simulations/default/scenariobatch_3EffGOkvYHtmpmEKGf9hFmeBWDP) | ✅ 3 msgs / 3 parts | ✅ 200, SUCCESS | ✅ 1.56s → *"Hello, how can I help you today?"* |
| ElevenLabs branded + **audioEffects** | [`scenariorun_3EffKAJl4dXbP1IU41rxCzYyiwT`](https://app.langwatch.ai/scenario-tracing-bZspxw/simulations/default/scenariobatch_3EffGOkvYHtmpmEKGf9hFmeBWDP) | ✅ 4 msgs / 4 parts | ✅ 200, SUCCESS | ✅ effects-processed audio STT-clean |
| Pipecat (**mulaw@8000** source) | [`scenariorun_3EffZC024s9IrYTOaIoKbQawMpq`](https://app.langwatch.ai/scenario-tracing-bZspxw/simulations/default/scenariobatch_3EffXMCuHAIPIURRRxzZUJMjhjO) | ✅ 5 msgs / 5 parts, **all normalized `audio/pcm16`** — AC9 contract held for a telephony source format | ✅ 200, SUCCESS | ✅ 3.80s → *"Hello, thank you for calling. How can I help you today?"* |
| Twilio (PSTN) | — | not runnable in this environment (needs live telephony + public tunnel); SDK layer CI-confirmed per AC11 | — | — |

**App-side net (final, 2026-06-04):** **all app-side ACs (AC13–18) are ✅ verified against production across the adapter matrix** — 5 live runs / 4 adapter families (matrix above), covering happy-path, **interruption/truncation**, **failed-verdict persistence**, **audio effects**, and a **telephony source format**. Every leg: SDK → collector → `traces/search` → `simulation-runs` REST → `/api/files` audio bytes, with message↔trace cross-linkage intact and served audio STT-verified as the correct conversational content at the contract format (pcm16/24kHz). The inline player shipped via [#4058](https://github.com/langwatch/langwatch/pull/4058) ([#3781](https://github.com/langwatch/langwatch/pull/3781) is stale, not a blocker). **Parity board AC1–18: complete.** Optional formality: eyeball any run at the simulations page — every layer beneath that click is independently verified.

## What changed (decisions)

- **Per-run provider state, not module-global** (ADR-002): voice config rides on `ScenarioConfig.voice` and reaches the adapters per-run — no global `configure(stt=)`. One AI-SDK `file` audio format end-to-end; STT/TTS are one-file-per-provider.
- **`agent({ wait: false })` is the barge-in primitive** (PRD §4.4): the executor starts the agent's reply without blocking so a `user()` / `interrupt()` lands mid-utterance; the transport's native cancel fires and the cut-off segment is marked truncated.
- **Demo promises are encoded as gates, split by what's verifiable**: an LLM judge reading a transcript can't see audio properties, so judge criteria assert the *conversational* half (empathy, acknowledging the specific request, recovery) and deterministic *code* assertions assert the *audio* half (`transcriptTruncated` + shorter segment for cut-off; noise-floor ≫ silence for mixed ambience). A hollow demo now fails.
- **Real-key voice demos run via a manual `voice-integration` workflow**, not PR CI — mirroring Python's `voice-integration.yml`. They cost real API money and can flake, so they never gate a merge; `ci-checks` (units + build + the non-bot `@ts-e2e` gate) is the merge gate.
- **Pre-step interrupt scheduling** (`maybeScheduleInterruptedAgentTurn`, `daa357d`): ports Python's pre-step pattern; the executor decides to interrupt *before* the next agent turn begins rather than patching it up post-step. Eliminates the prior hollow post-step path.
- **Inline-TTS barge-in** (`fa84c8f`): `voiceProceed` fires the interrupt via `voiceifyText` while the AGENT is still in-flight, bypassing the user-sim LLM to win the race against fast-streaming bots. `delayRange` is honoured in this path (`d20a49c`).
- **Pipecat adapter buffer-clear on interrupt** (`557cac2`): sets a `discardingInboundAudio` flag so late bot frames don't contaminate the next agent turn.
- **`VoiceEvent` discriminated union** (`4a49585`): five variant interfaces (`AgentSpeakingEvent`, `UserSpeakingEvent`, etc.) replace the prior `type:string` — the compiler now narrows on `event.type` without casts.
- **`RealtimeUserAgent` + `VoiceUserSimulator` structural interfaces + type guards** (`3f234f4`): kills `as unknown as` casts in the executor; adapter conformance is checked structurally at compile time.
- **`interruptRng` / `interruptWaitForSpeechMs` renamed to drop leading underscore** (`50cf3f2`): test-seam fields are now `@internal` JSDoc-tagged rather than visually private — consistent with the project's naming convention.
- **`interruption_recovery` judge promoted to hard gate** (`5bce766`): was informative-only; now the scenario fails if the judge says the agent didn't recover. Recording regenerated at `be600de`.
- **Gemini Live spurious-pair handling** proven by 3 new adapter unit tests (`4bd40c6`): `receiveAudio()` deduplicate logic is exercised deterministically without a live connection.
- **Docs caught up to the shipped API** (4 commits, `8eb7f55`…`c9c0f9a`):
  - `recipes/interrupt.mdx` now documents `voiceProceed({ interruptions: new InterruptionConfig({...}) })` — the PR's primary random-barge-in API — alongside the explicit `agent({ wait: false }) + interrupt()` primitive
  - `adapters/pipecat.mdx` capability table corrected (drops `opus` that the TS adapter doesn't support; source: `pipecat.ts:139-140`)
  - `adapters/elevenlabs.mdx` adds `systemPromptOverride` / `firstMessageOverride` per-session options
  - `recipes/effects.mdx` drops the stale "on the roadmap" callout (Python `script.py:user()` already accepts `voice_style` + `audio_effects`)

## How it works

<details>
<summary><b>Module tree</b> — 45 created + 18 modified source modules (+44 test files) under <code>javascript/src/</code>, single responsibilities, and how it functions together</summary>

### Created

```text
javascript/src/
├── config/configure.ts ················ global scenario configuration entry point (voice config rides ScenarioConfig.voice per ADR-002 — no module-global)
├── domain/agents/agent-shapes.ts ······ narrow structural interfaces for duck-typed user-simulator capabilities
├── script/voice-steps.ts ·············· voice script steps (sleep/silence/audio/dtmf/interrupt) + voiceProceed pre-step scheduling
└── voice/
    ├── config.ts ······················ per-run VoiceConfig + resolveVoiceConfig — keystone of the voice state model
    ├── adapter.runtime.ts ············· executor-side adapter wiring: connected-state gate, defaultVoiceCall, response draining
    ├── messages.ts ···················· the SOLE AudioChunk ↔ ModelMessage gateway (unified audio-message producer)
    ├── factories.ts ··················· lowercase adapter factories (PRD §9 idiom): openAIRealtimeAgent(), pipecatAgent(), …
    ├── interruption.ts ················ InterruptionConfig for proceed({ interruptions }) — probabilistic barge-in
    ├── vad.ts ························· SDK-side VAD fallback for adapters without native VAD (one-shot warning)
    ├── judge-stt.ts ··················· judge STT pre-pass — auto-transcription seam for non-multimodal judges
    ├── transcribe.ts ·················· post-hoc STT over a VoiceRecording (fills segment transcripts)
    ├── recording.runtime.ts ··········· WAV/MP3 save + segment directory + byte-accurate JSON manifest
    ├── playback.ts ···················· local-speaker playback sink (configure({ audioPlayback }))
    ├── segment-utils.ts ··············· pure segment/timeline post-processing (byte-cursor model)
    ├── ffmpeg.ts ······················ ffmpeg binary resolution (bundled via ffmpeg-static — no system dep)
    ├── utils.ts ······················· shared voice-layer utilities
    ├── agent-shapes.ts ················ @deprecated re-export shim → domain/agents/agent-shapes
    ├── adapters/ ······················ one transport per file
    │   ├── openai-realtime.ts ········· OpenAI Realtime (agent + user roles — the model IS the agent)
    │   ├── gemini-live.ts ············· Gemini Live native-audio (real server-VAD barge-in)
    │   ├── elevenlabs.ts ·············· ElevenLabs hosted ConvAI + branded transports
    │   ├── pipecat.ts ················· WebSocket client to a user-run Pipecat bot (pcm16/mulaw/opus)
    │   ├── twilio.ts ·················· real-phone transport via Twilio Media Streams
    │   ├── twilio-server.ts ··········· TwiML webhook + WS server impersonating Twilio's edge locally
    │   ├── twilio-shared.ts ··········· shared Media-Streams primitives (one canonical copy — Gap #6)
    │   ├── twilio-tunnel.ts ··········· public-tunnel harness so Twilio can reach the local server
    │   ├── twilio-logger.ts ··········· minimal parity logger for the Twilio adapter
    │   ├── composable.ts ·············· ComposableVoiceAgent: local STT → LLM → TTS (voice-test a text agent without a transport)
    │   ├── pending-transport-error.ts · PendingTransportError for stub transports (LiveKit/Vapi/WebRTC/WS — full transports → #371)
    │   └── index.ts ··················· adapter barrel
    ├── stt/ ··························· pluggable speech-to-text
    │   ├── stt-provider.ts ············ provider contract + "provider/model" router
    │   ├── openai-stt.ts ·············· default leaf (gpt-4o-transcribe)
    │   ├── elevenlabs-stt.ts ·········· ElevenLabs Scribe leaf
    │   ├── wav.ts ····················· minimal RIFF/WAV encoder for the STT upload edge
    │   └── index.ts ··················· barrel + registration site
    ├── tts/ ··························· pluggable text-to-speech
    │   ├── tts.ts ····················· router + cache core
    │   ├── openai-tts.ts ·············· default openai/<voice> leaf
    │   ├── elevenlabs-tts.ts ·········· elevenlabs/<voiceId> leaf (Gap #10)
    │   └── index.ts ··················· barrel + registration site
    ├── effects/ ······················· user-simulator audio-effects pipeline (§4.5)
    │   ├── index.ts ··················· pipeline + public surface
    │   ├── common.ts ·················· PCM16 @ 24kHz mono bytes ↔ Int16Array helpers
    │   ├── noise.ts ··················· backgroundNoise, static, multipleVoices
    │   ├── quality.ts ················· phoneQuality, lowQuality, packetLoss, echo, robotic, breakingUp
    │   ├── prosody.ts ················· volume scaling, time-stretching
    │   └── custom.ts ·················· arbitrary user Uint8Array → Uint8Array effects
    └── assets/noise/ ·················· 5 bundled noise WAVs (airport/babble/cafe/office/street) + LICENSES.md
```

### Modified (existing modules the voice layer hooks into)

```text
javascript/src/
├── index.ts ··························· public surface: voice namespace + lowercase factories exported
├── runner/run.ts ······················ scenario.run() host — resolves per-run voice config, threads it to the executor
├── execution/scenario-execution.ts ···· executor core — voice turn loop, pre-step interrupted-agent scheduling, interruptOverrides test seam
├── execution/scenario-execution-state.ts  run state — carries the voice executor reference across resets
├── agents/user-simulator-agent.ts ····· voice-aware user sim — per-run TTS voice, persona, audioEffects
├── agents/judge/judge-agent.ts ········ voice-aware judge — audio auto-detect, transcript fallback, structured timeline
├── config/index.ts ···················· config barrel — ScenarioConfig.voice wiring
├── domain/agents/index.ts ············· domain barrel — UserSimulatorAgentWithVoice
├── domain/core/execution.ts ··········· core types — agent({ wait:false }) non-blocking primitive
├── domain/scenarios/index.ts ·········· scenario domain types — voice fields
├── script/index.ts ···················· script barrel — voice steps surfaced
├── utils/convert-core-messages-to-agui-messages.ts  message conversion — audio file parts → input_audio; raw pcm16 is RIFF/WAV-wrapped at this langwatch-bound boundary so the app player decodes it (`885d294`)
└── voice/ (pre-existing PR1 bases, extended)
    ├── adapter.ts ····················· VoiceAgentAdapter base + AdapterCapabilities contract (+ isConnected gate)
    ├── index.ts ······················· voice namespace barrel
    ├── messages.types.ts ·············· audio message type surface
    ├── recording.types.ts ············· VoiceRecording/segment type surface
    ├── voice-executor-state.ts ········ runtime voice state — agent-speaking event, byte cursor
    └── voice-models.ts ················ canonical model ids (OPENAI_REALTIME_MODEL, …)
```

### How it functions together

`scenario.run({ voice })` (**runner/run.ts**) resolves a per-run `VoiceConfig` (**voice/config.ts** — no module-global state, ADR-002) and hands it to the executor (**execution/scenario-execution.ts**), which connects the chosen transport (**voice/adapters/\***) through the uniform connected-state gate (**voice/adapter.runtime.ts**). On each user turn the simulator (**agents/user-simulator-agent.ts**) synthesizes speech via the TTS router (**voice/tts/**) and layers realism through the effects pipeline (**voice/effects/** + bundled **assets/noise/**); audio streams over the transport and the agent's reply is drained on tail-silence windows (**adapter.runtime.ts**). Every chunk crosses a single gateway (**voice/messages.ts**) into `ModelMessage`s carrying `input_audio` file parts — the exact shape langwatch ingests (**utils/convert-core-messages-to-agui-messages.ts**). Barge-ins come from script steps (**script/voice-steps.ts**) or probabilistic config (**voice/interruption.ts**): native server-VAD where the adapter supports it, SDK fallback otherwise (**voice/vad.ts**), with truncation marked on the byte cursor (**voice/segment-utils.ts**, **voice-executor-state.ts**). At judgment time a pre-pass (**voice/judge-stt.ts** → **voice/stt/**) transcribes audio for non-multimodal judges (**agents/judge/judge-agent.ts**). The run's audio persists as `full.wav` + per-segment files + a byte-accurate manifest (**voice/recording.runtime.ts**), optionally monitored live (**voice/playback.ts**), and rides `ScenarioResult` alongside timeline + latency extensions.

**Asset parity:** the 5 noise WAVs are **byte-identical** between `javascript/src/voice/assets/noise/` and `python/scenario/voice/assets/noise/` (md5-verified 2026-06-04, 144,044 B each), produced by the single deterministic generator `javascript/scripts/generate-noise-samples.mjs`; both sides carry `LICENSES.md`.

</details>

## File-org corrections (from audit)

Three PR-internal file-org corrections landed late in the PR after a structure audit:
- `scripts/generate-noise-samples.mjs` → `javascript/scripts/` (TS-only generator belongs in package scripts, not repo-root cross-language scripts)
- `docs/voice/internal-design.md` → `docs/adr/003-voice-internal-design.md` (it's an ADR by description; lives with ADR-001/002)
- `javascript/examples/vitest/recordings/` → `javascript/examples/vitest/outputs/recordings/` (semantic clarity + future-proof for traces/logs siblings)

Pre-existing-on-main cleanup landed separately as **PR [#586](https://github.com/langwatch/scenario/pull/586)** (deletes orphan docs, publishes happy-path guides, folds capability-matrix duplication, python `recordings → outputs/recordings` rename, python noise-sample parity refresh). After #586 merges this branch rebases off cleaned main.

## Test plan

- `pnpm -F @langwatch/scenario test` → **791 pass / 1 skipped** (incl. the new `interrupt-truncation`, noise-energy, byte-cursor, proceed-loop pre-step, and Gemini Live spurious-pair unit tests).
- `@ts-e2e` round-trip gate (real keys) green; `tsc --noEmit`, `build:all`, `lint:all`, `typecheck:all` clean.
- Regression guard: the interrupt clock-mismatch fix is covered by a unit test exercising divergent cursor-vs-wall-clock times (the prior same-scale test masked the bug).

## How I can prove it works

16 committed demo recordings (`javascript/examples/vitest/outputs/recordings/<demo>/` — `full.wav` + byte-accurate `manifest.json`), generated against live providers + the bundled Pipecat bot. The ones demonstrating the headline behaviors (open the blob → GitHub shows an audio player):

| Behavior | Demo |
|---|---|
| Reply **cut off mid-sentence**, then recovers (judge hard-gated) | [interruption_recovery](https://github.com/langwatch/scenario/blob/39226091f6fb6b7d643d1c598a61b393420e7932/javascript/examples/vitest/outputs/recordings/interruption_recovery/full.wav) |
| Probabilistic barge-ins via inline-TTS + canned-phrase strategy | [random_interruptions](https://github.com/langwatch/scenario/blob/39226091f6fb6b7d643d1c598a61b393420e7932/javascript/examples/vitest/outputs/recordings/random_interruptions/full.wav) |
| Server-VAD barge-in on **Gemini Live** (real mid-stream cut-off) | [gemini_live_interruption](https://github.com/langwatch/scenario/blob/39226091f6fb6b7d643d1c598a61b393420e7932/javascript/examples/vitest/outputs/recordings/gemini_live_interruption/full.wav) |
| **Audible anger** (ElevenLabs tonal markers) + cafe noise | [angry_customer](https://github.com/langwatch/scenario/blob/39226091f6fb6b7d643d1c598a61b393420e7932/javascript/examples/vitest/outputs/recordings/angry_customer/full.wav) |
| Genuine engagement, not a canned greeting | [basic_greeting](https://github.com/langwatch/scenario/blob/39226091f6fb6b7d643d1c598a61b393420e7932/javascript/examples/vitest/outputs/recordings/basic_greeting/full.wav) |

Full set (16) also covers the adapters (openai_realtime ×2, elevenlabs hosted/branded, gemini_live), composable_stt_swap, recording_playback, voice_text_parity, pipecat ×2, background_handoff.

## Anything surprising

- **`evaluate` check is red — expected & non-blocking, fix lands with PR #586.** This PR's diff exceeds GitHub's 20k-line API cap, so the eval bot can't fetch it (HTTP 406). Not a required check. PR [#586](https://github.com/langwatch/scenario/pull/586) (commits `bafdbf7e15` + `cdce271bb2`) catches the 406 with a grep-specific `oversized=true` path + env: pattern hardening for `oversized_reason`; supersedes PR #572 and closes #571 on merge. Once #586 merges and this branch rebases on the new main, `evaluate` will go green on this PR too.
- **The 6 Pipecat-bot demos run only in the manual `voice-integration` workflow** (which now sets `SCENARIO_PIPECAT_BOT_UP`); pre-merge they're proven by the committed recordings + local real-key runs, and the workflow guards them post-merge.
- **`random_interruptions` recording is honest about Pipecat-bot limits.** The bundled Pipecat stub bot bursts TTS frames in ~50 ms of wall time (not realtime streaming), so by the time `adapter.interrupt()` fires the bot has already sent all frames — real mid-stream audio cut-off isn't observable with this transport. The demo's assertions encode what the bot *can* prove: interrupt fires + `fired_after_speech` outcome + canned-phrase strategy ran + truncation label + agent recovers + multi-turn conversation. For real audio cut-off under server-side cancel, see `gemini_live_interruption`. A transport-upgrade follow-up is drafted (see `/tmp/voice-spec/issue-random-interruptions-followup.md`).
- Follow-ups filed: #568 (Python demo-fidelity parity), #569 (`speakingEvent` empty-chunk), #570 (path/URL hardening), #571 (`evaluate` oversized-diff). Remaining transports (LiveKit/Vapi/WebRTC/WebSocket) tracked under #371.

---

## Closes (574-585 grind — landed in this PR)

Closes #574 #575 #576 #578 #579 #580 #581 #582 #583 #584 #585

_All 11 follow-up issues from the post-review NIT batch were addressed in this PR via 60+ commits since `4d83724`. Per-issue close-out comments are on each issue; partial outcomes for #580 (Gemini adapter improved, demo workaround retained) and #583 (adapter dequeue race fixed, transport switch reverted) are documented honestly there._


drewdrewthis:
## 🎧 Voice demo recordings — click to listen

Real audio captured from each demo's `scenario.run()` (committed under `javascript/recordings/<demo>/`). Click to play in a browser tab:

| Demo | Listen |
|---|---|
| OpenAI Realtime — agent | [▶ play](https://github.com/langwatch/scenario/raw/voice/372-refactor/javascript/recordings/openai_realtime_agent/full.wav) |
| OpenAI Realtime — user-sim | [▶ play](https://github.com/langwatch/scenario/raw/voice/372-refactor/javascript/recordings/openai_realtime_user/full.wav) |
| ElevenLabs — hosted ConvAI | [▶ play](https://github.com/langwatch/scenario/raw/voice/372-refactor/javascript/recordings/elevenlabs_hosted/full.wav) |
| ElevenLabs — branded/composable | [▶ play](https://github.com/langwatch/scenario/raw/voice/372-refactor/javascript/recordings/elevenlabs_branded/full.wav) |
| Gemini Live | [▶ play](https://github.com/langwatch/scenario/raw/voice/372-refactor/javascript/recordings/gemini_live/full.wav) |
| Composable (STT swap) | [▶ play](https://github.com/langwatch/scenario/raw/voice/372-refactor/javascript/recordings/composable_stt_swap/full.wav) |
| Pipecat (WebSocket) | [▶ play](https://github.com/langwatch/scenario/raw/voice/372-refactor/javascript/recordings/pipecat_ws/full.wav) |
| Recording + playback | [▶ play](https://github.com/langwatch/scenario/raw/voice/372-refactor/javascript/recordings/recording_playback/full.wav) |
| Voice ↔ text parity | [▶ play](https://github.com/langwatch/scenario/raw/voice/372-refactor/javascript/recordings/voice_text_parity/full.wav) |

> **Embedded player:** GitHub renders a player only for files uploaded as comment attachments — drag a `.wav` into a reply box and it embeds inline. Per-segment WAVs + `manifest.json` live in each demo folder.

rogeriochaves (commented):
3 findings: 2 P2s, 1 P3.

**P2 — interrupted agent turn failures are swallowed**

File/line: `javascript/src/execution/scenario-execution.ts:1481`

Why: the random-interruption path starts the AGENT turn with `this.callAgent(...).catch(() => {})`, and `fireUserInterrupt` later awaits that already-recovered promise. If the voice adapter or agent call fails during this path, the exception is converted into success, so `execute()` never hits the normal error handling at `execute()` and the scenario can continue to a later judge/succeed step as if the agent turn completed. This diverges from the normal `_step -> callAgent` path, where the same failure fails the run.

Minimal fix direction: keep cleanup/done tracking, but preserve the rejection: either do not catch here, or catch into a stored `error` and rethrow after `fireUserInterrupt`/pending-task cleanup. Add a unit test with a voice AGENT whose `call()` rejects while `voiceProceed({ interruptions })` is active.

**P2 — `voiceProceed({ interruptions })` leaks its config into later proceed loops**

File/line: `javascript/src/script/voice-steps.ts:371`

Why: the script step writes `executor.voiceInterruptions = options.interruptions` and never restores or clears it after `executor.proceed(...)`. The docs describe the option as applying to the `voiceProceed` loop, but a later plain `proceed()` or `voiceProceed({ turns })` in the same scenario will still use the old interruption config because `resolveInterruptionConfig()` gives `voiceInterruptions` priority over the user simulator default. That can inject unexpected random barge-ins into script sections that did not opt in.

Minimal fix direction: save the previous value, set the per-call value, and restore it in a `finally` after `executor.proceed(...)`. Add a unit test with two proceed calls where only the first passes `interruptions` and assert the second does not use the prior config.

**P3 — `git diff --check` fails on blank lines at EOF**

File/line: `javascript/src/execution/scenario-execution.ts:2476`, `javascript/src/script/__tests__/voice-steps.test.ts:654`, `javascript/src/script/voice-steps.ts:532`, `javascript/src/voice/adapters/twilio.ts:461`

Why: `git diff --check c8cca4ecafb0ebfb10d2d39b36ac2fe28443a376...HEAD` reports `new blank line at EOF` in these four files. This is low-risk, but it leaves the diff-check hygiene gate red for anyone using it locally/CI.

Minimal fix direction: remove the trailing blank EOF lines.

Validation I ran locally in `/tmp/scenario-pr-561/javascript`:

- `pnpm install --frozen-lockfile`
- `pnpm test -- src/voice/__tests__/proceed-interrupt.test.ts src/execution/__tests__/proceed-interruptions.test.ts src/script/__tests__/voice-steps.test.ts` -> 764 passed / 1 skipped
- `pnpm typecheck` -> passed
- `git diff --check c8cca4ecafb0ebfb10d2d39b36ac2fe28443a376...HEAD` -> failed as noted above

drewdrewthis:
Thanks @rogeriochaves — all 3 addressed:

- **P2 swallowed errors** (`scenario-execution.ts:1481`): fixed in 06c7b17 — capture rejection into `entry.error` on the `PendingAgentTask`, re-throw at both `fireUserInterrupt` drain sites. Unit test added (`proceed-interrupt-errors.test.ts`) with a voice AGENT whose `call()` sets `agentSpeakingEvent` then rejects — asserts `execute()` rejects with the error rather than silently succeeding.
- **P2 voiceInterruptions config leak** (`voice-steps.ts:371`): fixed in 77705cc — save-set-restore in a `try/finally`. Unit test added in `voice-steps.test.ts` with two sequential `voiceProceed` calls; asserts the second call sees `voiceInterruptions === undefined`, not the first call's config.
- **P3 trailing blank lines** (4 files): fixed across 06c7b17, 77705cc, and 52bb0fc. `git diff --check c8cca4e...HEAD` now reports nothing.

Pushed to `voice/372-refactor`; CI re-running.

langwatch-agent:
1 P1, 1 P2

- P1: `javascript/src/execution/scenario-execution.ts:1091-1097` short-circuits the `user()` flow as soon as `maybeFireUserInterrupt(audioMessage)` succeeds. That bypasses the normal `scriptCallAgent(AgentRole.USER, audioMessage)` path, which is what appends the voiced turn to `state.messages` and broadcasts it. As a result, a barge-in via `scenario.user("...")` is audible on the wire but never enters conversation history, so the judge and later agents never see the correction. Please record/broadcast the voiced message before returning, or move that bookkeeping into the interrupt helper.

- P2: `javascript/src/execution/scenario-execution.ts:1499-1518` samples `interruptBargeInDelayMs` before `voiceifyText(...)`, but if TTS throws the catch returns `true` without clearing that per-call delay or unwinding the interrupted turn state. That leaves stale delay state behind, so a later successful barge-in can inherit the previous delay even though the earlier attempt was skipped. At minimum, clear the delay on the failure path before returning.

drewdrewthis:
📣 Slack PR-review request posted to #dev — [thread](https://langwatch.slack.com/archives/C06JQLW1HE2/p1779993165428159). Label `slack-requested` couldn't be applied (OAuth-App restriction on this token's label-write scope); this comment is the dedup marker until the label can be added manually.

docs/voice/typescript.md
```diff
@@ -0,0 +1,112 @@
+# Voice testing in Scenario (TypeScript) — see the published docs
```
rogeriochaves: huh? all does docs/voice files are weird, think they were supposed to be temporary, don't think they should be commited

javascript/docs/voice/capability-matrix.md
```diff
@@ -1,87 +1,135 @@
 # Voice Adapter Capability Matrix — TypeScript SDK
```
rogeriochaves: and why do we have suddenly a new docs/ folder here too?

javascript/examples/vitest/outputs/recordings/angry_customer/full.wav
rogeriochaves, line 1: why do we have all those wav files commited there, recordings/ folder? why do we need to push this to git?

if it's to show on the docs, shouldn't it be uploaded inside the docs folder? if its for examples, shouldnt it be on examples folder? if they are generated, why commiting at all?

javascript/scripts/generate-noise-samples.mjs
rogeriochaves, line 1: why do suddenly we have a new scripts/ folder also? shouldn't this noise generator be part of the library itself?

javascript/src/voice/assets/noise/LICENSES.md
```diff
@@ -0,0 +1,23 @@
...
+
+- cafe.wav — voice-band chatter murmur + slow conversation swell + sparse cup clinks
+- street.wav — deep traffic rumble + tyre hiss + two passing-vehicle doppler sweeps
+- office.wav — steady HVAC/mains hum (60/120/240 Hz) + faint air hiss + keystroke bursts
+- airport.wav — large-hall crowd murmur + periodic band-limited PA announcement bursts
+- babble.wav — six overlapping syllable-rate-modulated talkers, for the
+  `multipleVoices` effect (NOT a `backgroundNoise` preset; see proposal §4.5
+  L521 vs L533).
+
+To regenerate: `cd javascript && node scripts/generate-noise-samples.mjs`.
```
rogeriochaves: I see... ok feel free to push back if scripts/ is the right place for it not inside scenario framework itself

docs/voice/typescript.md
drewdrewthis: test

drewdrewthis: Good catch — you're right. This file started as a comprehensive in-repo guide when the published docs didn't yet have a TS view; PR #566 merged that content into the docs site (`docs/docs/pages/voice/*.mdx` now has language tabs), and this file got shrunk to a 110-line "we moved" pointer. The file itself says "there is no longer a separate TS-only guide to keep in sync" — so the intent is exactly that it should be gone.

Deleted in `3c45d3d` along with `docs/voice/CONSOLIDATION-MAP.md` (the merge-conflict resolution log, also a pure in-flight artifact). The 3 README links that pointed here are retargeted to the canonical published page: https://scenario.langwatch.ai/voice/getting-started

javascript/docs/voice/capability-matrix.md
drewdrewthis: The directory actually predates this PR — it was added by [PR #511](https://github.com/langwatch/scenario/pull/511) on 2026-05-21 along with an explicit `javascript/.gitignore` carve-out (`docs/* / !docs/voice/`). This PR only **rewrote** the placeholder content (it was a "PR1 stub, no adapters land yet" page) into the real render — it didn't introduce the tree.

The reason the doc lives there at all: it's the **authoritative TS capability render** — the published docs site ([`docs/docs/pages/voice/capability-matrix.mdx`](https://github.com/langwatch/scenario/blob/main/docs/docs/pages/voice/capability-matrix.mdx)) links to it as "the authoritative TS render" (search for `javascript/docs/voice/capability-matrix.md` in that file — two deep links). It's symmetric with the Python source-of-truth at top-level [`docs/voice/capability-matrix.md`](https://github.com/langwatch/scenario/blob/main/docs/voice/capability-matrix.md): Python is the reference SDK so its render lives at the top, JS gets its own subtree because we explicitly say "TS ships a subset, here's what".

If this is the only file that ever lives under `javascript/docs/`, fine — that's the intent. Any further TS-specific docs should go to the published `docs/docs/pages/voice/` site (with TS tabs) instead, not pile up here.

javascript/examples/vitest/outputs/recordings/angry_customer/full.wav
drewdrewthis: Fair to ask — let me unpack the four things you wondered about, in order.

1. **"Why push to git at all?"** They're the prove-it artifacts for this PR (and for the voice port long-term). GitHub renders an inline `<audio>` player on the blob URL of a `.wav`, so a reviewer can click [interruption_recovery/full.wav](https://github.com/langwatch/scenario/blob/voice/372-refactor/javascript/recordings/interruption_recovery/full.wav) from the PR diff and hear the agent actually get cut off mid-sentence and recover. That's the demonstration that the code does the thing — it's hard to make that point any other way (a transcript can't show truncation; a screenshot can't play audio). The PR body links 5 of them as evidence for exactly this reason.

2. **"Shouldn't it be inside the docs folder?"** The docs site doesn't need them — it links the GitHub blob URLs (same inline player). Putting them under `docs/` would lose the 1:1 path mirror with the tests that produce them and diverge from Python.

3. **"Shouldn't it be on examples folder?"** They're outputs of the `@e2e` tests under `javascript/examples/vitest/tests/voice/`, not inputs to them, so sitting next to the tests would mix source + artifact. The current layout matches Python's [`python/recordings/`](https://github.com/langwatch/scenario/tree/main/python/recordings/) 1:1 — same per-demo subdirs, same `full.wav`/`manifest.json`/`segments/` shape, same `.gitignore` policy (only the flagship demos commit `segments/`; the rest commit `full.wav` + `manifest.json` only). The JS `.gitignore` block (lines 230–263) literally annotates "same policy as python/recordings/ above."

4. **"If they are generated, why committing at all?"** Same reason `python/recordings/` is committed — the gen step needs real API keys (`OPENAI_API_KEY`, `ELEVENLABS_API_KEY`, `GEMINI_API_KEY`) and live transports (the Pipecat bot, ElevenLabs ConvAI, Gemini Live), which means CI can't deterministically regenerate them per PR (cost + flake + transport availability). The `voice-integration.yml` workflow DOES regenerate + upload them as an artifact on demand (`path: javascript/recordings/**`, retention 14 days), but the committed versions are the ones reviewers can play without running real-key CI.

Net: dir is 24M today (Python's is 15M), `.gitattributes` already declares `*.wav binary` so no encoding drift, and the carve-out gitignore caps growth. Same shape as Python — if we change one, we should change both.

javascript/scripts/generate-noise-samples.mjs
drewdrewthis: The script and its location both mirror [`python/scripts/generate_noise_samples.py`](https://github.com/langwatch/scenario/blob/main/python/scripts/generate_noise_samples.py) 1:1 — same 5 preset names (cafe/street/office/airport/babble), same output target (the Python script writes to `python/scenario/voice/assets/noise/`, this one writes to `javascript/src/voice/assets/noise/`). Python's `scripts/` dir is established convention — it also holds `gen_capability_matrix.py` and `generate_openapi_client.sh`. So `<lang>/scripts/` is "dev-time tooling that writes checked-in artifacts," and the Python half has set the precedent.

On "shouldn't this noise generator be part of the library itself?" — deliberately no. The library (the npm package) ships only what `package.json` `files` lists: `dist/`, `src/voice/assets/**`, README, LICENSE. Consumers get the **generated wavs**, not the generator. Pulling the generator into `src/` would (a) inflate the published package, (b) import `node:fs`/`node:path` into the runtime surface, and (c) blur the dev-tool / library boundary that Python's side already keeps clean.

The asset bytes are deterministic (seeded mulberry32 PRNG → re-running the script produces byte-identical WAVs), so we still get all the regenerate-from-source guarantees you'd want from "treat the generator as source"; we just don't ship the generator to users.

javascript/src/voice/assets/noise/LICENSES.md
drewdrewthis: The LICENSES.md sits next to the assets it covers — same shape as [`python/scenario/voice/assets/noise/LICENSES.md`](https://github.com/langwatch/scenario/blob/main/python/scenario/voice/assets/noise/LICENSES.md). `package.json` `files: ["src/voice/assets/**", ...]` ships the attribution into the published npm tarball alongside the audio, which is what we want for CC0 content. Pushing back on the `scripts/` location per your follow-up: yes, I think that location IS right — see the reply on the generator script (mirrors Python's `python/scripts/` convention; we explicitly don't want the generator in the library itself).

drewdrewthis:
## langwatch-agent review addressed — P1 + P2 fixed

Both findings from the 2026-05-28T16:00:33Z review have been applied to this branch. Summary below.

---

### P1 — barge-in user audio never appends to state.messages

**Diagnosis:** Confirmed. `maybeFireUserInterrupt` returned immediately after `fireUserInterrupt`, skipping the `scriptCallAgent` path (`state.addMessage` + `broadcastMessage`). The voiced barge-in was audible on the wire but absent from `state.messages`, so the judge and downstream agents never saw the correction.

**Fix applied:** Option B (move bookkeeping into the helper). `maybeFireUserInterrupt` now calls `this.state.addMessage(voicedMessage)` + `this.broadcastMessage(voicedMessage)` before returning `true`, mirroring the identical pattern already present in `maybeScheduleInterruptedAgentTurn` (lines 1539–1543).

**Commit:** `23f8ad9` — `fix(voice/561): append + broadcast barge-in user message into state.messages (P1, langwatch-agent)`

**Test added:** `proceed-interrupt.test.ts` — `"explicit user() barge-in recorded in state.messages (P1 fix)"`: runs `agentNonBlocking()` + `user("barge-in correction")` and asserts `result.messages` contains the user turn with the expected transcript.

---

### P2 — stale `interruptBargeInDelayMs` leaks on TTS failure

**Diagnosis:** Confirmed. `interruptBargeInDelayMs` is sampled before `voiceifyText`. When TTS throws, the catch block returned `true` without clearing the field — leaving the stale sampled value to be inherited by a later successful barge-in whose own delay would be 0 (because `delaySeconds > 0` condition skips the assignment).

**Fix applied:** Added `this.interruptBargeInDelayMs = undefined` in the TTS failure catch block before `return true`. Return value kept as `true` — the AGENT task was already dispatched non-blocking and proceed continues normally; only the barge-in audio was skipped.

**Commit:** `fa5a7aa` — `fix(voice/561): clear interruptBargeInDelayMs on TTS failure path (P2, langwatch-agent)`

**Test added:** `proceed-interrupt-errors.test.ts` — `"stale interruptBargeInDelayMs cleared on TTS failure"`: uses a controlled RNG (fires exactly one barge-in attempt, skips all subsequent ones) with a user sim whose first `voiceifyText` throws. After `proceed(2)`, asserts `exec.interruptBargeInDelayMs === undefined` — would be `1000ms` (stale) without the fix.

---

Head SHA: `fa5a7aa5cd9f1e962e21e3c03db09e5a5c5ba165`

langwatch-agent:
No blockers.

I re-read the existing thread, checked the new commits since the last review, and did not find a fresh regression. The P1/P2 issues from the prior review are addressed by the new commits and covered by tests. Residual risk: I did not run the full scenario test suite in this pass.

drewdrewthis:
Followup on the file-org concern from @rogeriochaves's review — Drew called out that "Python parity" was the wrong defense. Restructured:

- `javascript/scripts/generate-noise-samples.mjs` → `scripts/generate-noise-samples.mjs` (repo-root, matches existing convention alongside `apply-branch-protection.sh`, `ci-wait.sh`, etc.)
- `javascript/recordings/` → `javascript/examples/vitest/recordings/` (workspace-local; vitest examples is the sole consumer/writer)
- `saveDemoRecording` helper now resolves workspace-relative (3 `..` up from `helpers/`, not 5), anchored to `import.meta.url`
- CI artifact path + .gitignore + PR body recording links updated to new location

Pre-existing `javascript/docs/` (from #511) left alone — out of scope.

Commits:
- `a44c91b` chore(voice/561): move javascript/scripts/generate-noise-samples.mjs to repo-root scripts/
- `68ad829` chore(voice/561): move javascript/recordings/ to javascript/examples/vitest/recordings/
- `106929b` fix(voice/561): update saveDemoRecording helper path to workspace-relative
- `fef9ea6` chore(voice/561): update gitignore + CI artifact paths + doc comments

CI: all checks green (Build ✓, Lint ✓, Type check ✓, Tests ✓, Test Examples ✓) on run https://github.com/langwatch/scenario/actions/runs/26608527702

langwatch-agent:
2 P3 findings on the new file-organization commits

- **P3 — noise generator now writes to the wrong directory after being moved to repo root.**  
  File: `scripts/generate-noise-samples.mjs:34`  
  The script was moved from `javascript/scripts/` to repo-root `scripts/`, but `OUT_DIR` still resolves as `resolve(HERE, "..", "src", "voice", "assets", "noise")`. With `HERE=<repo>/scripts`, the documented root command writes to `<repo>/src/voice/assets/noise`, while the actual packaged assets live under `javascript/src/voice/assets/noise`. Running the old `cd javascript && node scripts/...` command also no longer works because the script is no longer there. Minimal fix: change `OUT_DIR` to include the `javascript` package segment, e.g. `resolve(HERE, "..", "javascript", "src", "voice", "assets", "noise")`, and update the header comment to match the new root invocation.

- **P3 — JavaScript README still links to the removed recordings directory.**  
  File: `javascript/README.md:274`  
  The new commits moved recordings to `javascript/examples/vitest/recordings/`, but this user-facing README still points at `./recordings/README.md`, which is now a dead link from the package README. Minimal fix: point it to `./examples/vitest/recordings/README.md`.

Verification notes:
- Reviewed the new range after the prior langwatch-agent review: `fa5a7aa..bac144a`.
- `git diff --check fa5a7aa..HEAD` passed.
- I could not run the focused Vitest files locally because this worktree has no installed `vitest` binary (`pnpm --dir javascript exec vitest ...` failed with `Command "vitest" not found`).

drewdrewthis:
Both P3 findings have been addressed in the post-audit consolidation:

**P3#1 — noise generator now writes to the wrong directory.** Resolved in commit `720e260` (`chore(voice): move generate-noise-samples back to javascript/scripts/ (TS-only generator)`). The script is back under `javascript/scripts/generate-noise-samples.mjs`; the `OUT_DIR = resolve(HERE, "..", "src", "voice", "assets", "noise")` now correctly resolves to `javascript/src/voice/assets/noise/` from that location. The original `cd javascript && node scripts/generate-noise-samples.mjs` invocation works again.

**P3#2 — JavaScript README still links to the removed recordings directory.** Resolved in commits `1a4a389` (`refactor(voice/examples): rename javascript/examples/vitest/recordings → outputs`) and `3922609` (`refactor(voice/examples): nest outputs/recordings/ for future artifact types`). `javascript/README.md` line 276 now points at `./examples/vitest/outputs/recordings/README.md`, which exists. The old `recordings/` directory is gone.

Current HEAD `819ca28d22` reflects both fixes.

0xdeafcafe (approved):

PR #572, opened 2026-05-28
fix(ci/#571): soft-pass oversized PR diffs in pr-auto-approve evaluate

### LINKED ISSUE #571: ci: evaluate workflow hard-fails on PRs >20k-line diff instead of its oversized path

drewdrewthis:
## Why

`evaluate` in `pr-auto-approve.yml` hard-failed on PRs with diffs over GitHub's API limit (~20k lines). The `gh pr diff` fetch returned HTTP 406 and the step exited non-zero — the existing 100k-char ceiling that should have handled this never got reached. The check went red on every large PR (e.g. #561).

## What changed

- **Tolerant fetch**: `gh pr diff` failure now sets `oversized=true` + captures stderr to `oversized_reason`. The existing oversized path runs (sets `qualifies=false` with an honest reason) and the workflow exits 0.
- **Honest reasoning**: the auto-posted PR comment now explains *why* the diff couldn't be evaluated (fetch failure vs. char-count exceeded), so the human reviewer doesn't have to dig into the workflow log.
- **Hardened interpolation**: `oversized_reason` is passed via `env:` rather than `${{ }}` direct expansion in `run:` — the documented pattern for shell-injection hardening even when the source is trusted.

Per the policy: oversized = NOT-low-risk. The size signal alone gives the answer; the AI evaluator isn't needed.

## Test plan

- YAML validated locally (`python3 -c 'import yaml; yaml.safe_load(...)'`).
- The failure path will be exercised by the next oversized PR (e.g. PR #561 once it re-syncs against `main` post-merge). Expected outcome: `evaluate` posts the `qualifies=false` comment with the fetch-failure reason in the body, and the check exits 0.

Closes #571.

### DIFF
--- a/.github/workflows/pr-auto-approve.yml
+++ b/.github/workflows/pr-auto-approve.yml
@@ -260,7 +260,12 @@ jobs:
             ERR_MSG=$(head -c 500 /tmp/pr_diff_err.txt 2>/dev/null || echo "(no stderr captured)")
             echo "gh pr diff failed (likely oversized diff): $ERR_MSG"
             echo "oversized=true" >> "$GITHUB_OUTPUT"
-            echo "oversized_reason=Diff too large to fetch via GitHub API (fetch error: $ERR_MSG)" >> "$GITHUB_OUTPUT"
+            # Use a heredoc so multi-line stderr doesn't corrupt GITHUB_OUTPUT format.
+            {
+              printf 'oversized_reason<<__EOF_OVERSIZE__\n'
+              printf 'Diff too large to fetch via GitHub API (fetch error: %s)\n' "$ERR_MSG"
+              printf '__EOF_OVERSIZE__\n'
+            } >> "$GITHUB_OUTPUT"
           else
             # Fail closed if diff exceeds model window — large PRs could hide risky changes.
             DIFF_LIMIT=100000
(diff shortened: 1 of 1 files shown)


langwatch-agent:
No blockers.

Residual risk: the new fallback treats any `gh pr diff` failure the same as an oversized diff, so auth/network/API issues will be reported as manual-review-needed rather than distinctly surfaced. That is acceptable for a fail-closed path, but it does blur operational diagnosis a bit.

Validation: `python3` YAML parse passed for `.github/workflows/pr-auto-approve.yml`.

drewdrewthis:
Superseded by PR #586, which carries the same 406-handling fix at commit bafdbf7e15 (plus a follow-up hardening commit picking up the `env:`-pattern + heredoc improvements from this PR's commit 7dd9311). #586 also bundles the broader main-side cleanup (docs orphans, capability-matrix fold, python parity rename, tsconfig fix, etc.) and its fix targets the new file layout (`.github/LOW_RISK_PULL_REQUESTS.md` instead of `docs/`). Closing here to keep one PR to review.

Both PRs auto-close #571 on merge of #586.

Issue #567, opened 2026-05-27
fix(typescript-sdk/voice): harden ElevenLabs ConvAI post-interrupt / next-turn receive


drewdrewthis:
Follow-up to #561 (epic #370).

The ElevenLabs hosted ConvAI adapter (`adapters/elevenlabs.ts`) times out on `receiveAudio` for the **post-interrupt** and **scripted next-turn** cases — the hosted server-VAD does not re-engage for scripted (non-mic) audio. Consequences in #561:
- `elevenlabs_interruption` demo is **gated off** (`RUN_EL_INTERRUPTION=1` to attempt) — 4 honest attempts timed out on the post-interrupt receive. Not faked.
- `elevenlabs_hosted` demo is limited to **one live exchange** (a scripted 2nd user turn times out the next agent receive). EL multi-turn is covered instead by the in-process `elevenlabs_branded` (composable) demo.

## Scope
- Harden the EL ConvAI receive path so a scripted user turn after an agent turn (and after an interrupt) reliably re-engages a response — e.g. an explicit response trigger / commit, or a documented `silence()` nudge — without relying on mic-style server VAD.
- Then un-gate `elevenlabs_interruption` and extend `elevenlabs_hosted` to ≥2 exchanges.

## References
- PR #561 (consolidated TS voice stack), EDR #560, epic #370.

drewdrewthis:
Fix pushed on branch \`fix/567-el-convai-next-turn\` (commit b2a01a13), branched off #561's head.

**Root cause (confirmed live):** EL ConvAI 2.0 uses a hybrid VAD + DL turn-detector, not a silence threshold — so the fixed ~333ms zero-byte tail doesn't deterministically end a *scripted* (non-mic) turn. Greeting→turn-1 worked; scripted turn-2+ intermittently never re-engaged → \`receiveAudio\` timed out. (This is the same blocker that made a recent TS voice demo attempt abandon hosted EL ConvAI and pivot to OpenAI Realtime — it's fixable, not an EL-side dead end.)

[...]

PR #596, opened 2026-06-01
fix(voice/ts): explicit EL ConvAI turn-commit so scripted next-turn receive re-engages

drewdrewthis:
## Why

Addresses the core of #567 (the next-turn / post-interrupt `receiveAudio` timeout). The ElevenLabs hosted ConvAI adapter timed out on `receiveAudio` for scripted next-turn (and post-interrupt) cases: EL ConvAI 2.0's hybrid VAD + DL turn-detector doesn't treat a fixed ~333ms zero-byte silence tail as a deterministic end-of-turn on a scripted (non-mic) stream, so turn 2+ intermittently never re-engaged. This is the blocker that recently made a TS voice-demo attempt abandon hosted EL ConvAI and fall back to OpenAI Realtime — it's fixable, not an EL-side dead end.

> Stacked on #561 (`voice/372-refactor`) — the voice stack isn't on `main` yet, so this targets that branch, not `main`. Merge after / into #561.

## What changed

- **Explicit turn-commit over silence-VAD.** Verified against both official EL SDKs that ConvAI exposes no audio-flush/commit event; `user_message` (`{"type":"user_message","text":<transcript>}`) is the only client→server event that deterministically forces an agent response without mic-VAD (and is the SDK's own wire shape, so it won't 400 the socket).
- **New adapter options** `turnCommitMode: "text" | "silence"` (default `"text"`) + configurable `silenceTailBytes`. Text mode sends only the `user_message` commit (audio is still recorded locally; EL echoes `user_transcript`, so `lastUserTranscript`/observability stay intact). `"silence"` preserves the legacy path; `"text"` with no transcript falls back to silence.
- Untouched: ping/pong, transcript capture, `agent_response_correction`, `drainPendingWaiters`, capabilities, the `wsFactory` test seam.

## Test plan

- **Unit** (`src/voice/__tests__/elevenlabs-turn-commit.test.ts`, injectable fake socket): scripted 2nd user turn drives a 2nd `receiveAudio` resolution; exact `user_message` shape asserted; post-interrupt re-engages; both silence fallbacks. 7 pass; full voice suite 179 pass (21 files); examples typecheck clean.

- **Live, real EL socket, ≥2 exchanges, 3 consecutive clean runs (no retries, no timeouts):** `examples/vitest/tests/voice/elevenlabs-hosted.test.ts` extended to greeting→user→agent→user→agent→judge. `success=true, 5 segments`, 3 agent / 2 user turns, 14.24s. The previously-timing-out 2nd turn now answers. (A first revision that streamed audio+text together flaked 2/3 — that drove the text-only commit; recorded here as the honest negative.)

## Anything surprising?

- **Python parity (shipped, commit `a878cc1`):** `python/scenario/voice/adapters/elevenlabs.py` had the identical bug (16000-byte tail, no commit); ported the same `user_message` turn-commit + `turn_commit_mode`/`silence_tail_bytes` options, mirroring the TS change. Live-verified ≥2 exchanges against the real EL socket (agent answered turn 2); unit 7 pass / 74 no-regression. TS + Python ship together in this PR.
- **Remaining #567 scope (NOT in this PR):** un-gating `elevenlabs_interruption` (still `RUN_EL_INTERRUPTION=1`-gated) needs live post-interrupt verification not yet done — tracked as follow-up, which is why this PR *addresses* rather than fully *closes* #567. (`elevenlabs_hosted` IS extended to ≥2 exchanges here.)
- **Not un-gated:** `elevenlabs_interruption` (the issue's secondary ask) — the post-interrupt path is covered by the unit test and the same commit should fix it, but I didn't live-run it. Worth a follow-up.

### DIFF
--- a/javascript/examples/vitest/tests/voice/elevenlabs-hosted.test.ts
+++ b/javascript/examples/vitest/tests/voice/elevenlabs-hosted.test.ts
@@ -128,14 +128,15 @@ if (hasHostedKey || hasComposableKey) {
             );
             // #567: the explicit user_message turn-commit re-engages the agent
             // on the scripted 2nd user turn, so the live socket now carries
-            // multiple agent audio turns (greeting + ≥2 responses). Assert the
-            // multi-turn shape — more segments than the old single-exchange
-            // limit could produce — as the load-bearing live proof.
+            // greeting + user1 + agent1 + user2 + agent2 = 5 segments. Assert
+            // the full 5-segment shape — the pre-fix silence path stalls on
+            // user2 and produces at most 4 segments before the timeout, so ≥5
+            // is the falsifying floor that separates fixed from broken.
             expect(
               result!.audio!.segments.length,
-              "expected ≥3 audio segments (greeting + 2 agent responses) — " +
-                "a scripted 2nd-turn re-engagement is the #567 fix",
-            ).toBeGreaterThanOrEqual(3);
+              "expected ≥5 audio segments (greeting + user1 + agent1 + user2 + agent2) — " +
+                "the pre-fix silence path stalls on user2 and cannot produce 5",
+            ).toBeGreaterThanOrEqual(5);
             // The judge verdict is informative; surface success for the reviewer.
             expect(result!.success, `judge verdict: ${result!.reasoning}`).toBe(true);
           });
--- a/javascript/examples/vitest/tests/voice/elevenlabs-hosted.test.ts
+++ b/javascript/examples/vitest/tests/voice/elevenlabs-hosted.test.ts
@@ -84,22 +84,22 @@ if (hasHostedKey || hasComposableKey) {
               // with agent() so the greeting drains before user audio hits the
               // wire (mirrors the Python twin's script).
               //
-              // SINGLE scripted exchange (greeting → user → agent) — NOT extended
-              // to a second scripted user turn. The hosted ConvAI transport is
-              // server-VAD-driven: it segments turns from the *audio stream*, and
-              // a scripted second user turn after the agent has already replied
-              // does not reliably re-engage the server's turn-taking — the next
-              // `agent()` receive then times out (`receiveAudio timed out`,
-              // verified empirically). The MULTI-TURN ElevenLabs proof lives in
-              // the in-process `elevenlabs_branded` demo below, which has full
-              // turn control. This demo proves the live hosted-ConvAI WS path.
+              // MULTI-TURN (≥2 scripted exchanges) over the LIVE hosted ConvAI
+              // socket — un-gated by #567. The adapter now commits each user turn
+              // with an explicit `user_message` event instead of leaning on
+              // mic-style server VAD, so a scripted 2nd user turn after the agent
+              // has already replied reliably re-engages the next agent response
+              // (the old single-exchange limit + `receiveAudio timed out` are
+              // fixed). This is the load-bearing live proof for #567.
               script: [
                 scenario.agent(),
                 scenario.user("Hello, I have a question about my account."),
                 scenario.agent(),
+                scenario.user("Thanks. Can you tell me your support hours?"),
+                scenario.agent(),
                 scenario.judge(),
               ],
-              maxTurns: 6,
+              maxTurns: 8,
             });
             recordingDir = saveDemoRecording(result.audio, "elevenlabs_hosted");
           });
@@ -119,16 +119,24 @@ if (hasHostedKey || hasComposableKey) {
             },
           );
 
-          And("result.success is True after one turn", () => {
+          And("result.success is True after ≥2 exchanges", () => {
             if (!hasHostedKey) return;
             expect(recordingDir, "recording was not written").not.toBeNull();
             console.log(
               `[demo] elevenlabs_hosted → ${recordingDir} ` +
                 `(success=${result!.success}, ${result!.audio!.segments.length} segments)`,
             );
-            // The judge verdict is informative but EL hosted turn-taking is
-            // server-VAD-driven; the load-bearing proof is real audio over the
-            // live socket. Surface success for the reviewer.
+            // #567: the explicit user_message turn-commit re-engages the agent
+            // on the scripted 2nd user turn, so the live socket now carries
+            // multiple agent audio turns (greeting + ≥2 responses). Assert the
+            // multi-turn shape — more segments than the old single-exchange
+            // limit could produce — as the load-bearing live proof.
+            expect(
+              result!.audio!.segments.length,
+              "expected ≥3 audio segments (greeting + 2 agent responses) — " +
+                "a scripted 2nd-turn re-engagement is the #567 fix",
+            ).toBeGreaterThanOrEqual(3);
+            // The judge verdict is informative; surface success for the reviewer.
             expect(result!.success, `judge verdict: ${result!.reasoning}`).toBe(true);
           });
         },
--- a/javascript/src/voice/__tests__/elevenlabs-turn-commit.test.ts
+++ b/javascript/src/voice/__tests__/elevenlabs-turn-commit.test.ts
@@ -1,0 +1,267 @@
+/**
+ * Issue #567 — ElevenLabs ConvAI scripted next-turn / post-interrupt receive.
+ *
+ * The hosted ConvAI transport has NO audio end-of-turn client event (verified
+ * against the official EL Python + JS SDKs), so the pre-#567 adapter leaned on
+ * a fixed silence tail to coax server-side VAD. That tail does not reliably
+ * re-engage a response for a scripted turn 2+ (EL ConvAI 2.0 end-of-turn is a
+ * hybrid VAD + deep-learning turn-detector, not a pure silence threshold), so
+ * the 2nd `receiveAudio` timed out.
+ *
+ * The fix sends an explicit `{"type":"user_message","text":<transcript>}`
+ * turn-commit — the only documented client→server event that deterministically
+ * forces an agent response without mic-style VAD. On the text path the raw
+ * audio is NOT also streamed to EL (audio + text in one turn raced the server's
+ * ingestion and was live-flaky); the user audio is still recorded locally by
+ * the voice runtime and EL echoes the text back as a user_transcript.
+ *
+ * These tests drive the adapter through the injectable `webSocketFactory` /
+ * a fake socket and prove:
+ *  1. a scripted 2nd user turn after an agent turn drives a 2nd `receiveAudio`
+ *     resolution (the bug), AND each user turn emits a `user_message` commit;
+ *  2. the post-interrupt shape (agent audio mid-flight → user re-engages) also
+ *     commits + re-engages;
+ *  3. `turnCommitMode:"silence"` preserves the legacy pure-audio path;
+ *  4. `"text"` mode with no transcript falls back to the silence tail.
+ *
+ * Offline — no network, no real EL socket. The LIVE ≥2-exchange proof lives in
+ * `examples/vitest/tests/voice/elevenlabs-hosted.test.ts`.
+ */
+import { Buffer } from "node:buffer";
+
+import { describe, it, expect, beforeEach } from "vitest";
+import type { RawData } from "ws";
+
+import {
+  ElevenLabsAgentAdapter,
+  type WebSocketLike,
+} from "../adapters/elevenlabs";
+import { AudioChunk } from "../audio-chunk";
+
+/**
+ * Minimal in-memory {@link WebSocketLike}. Captures every frame the adapter
+ * sends and lets the test push inbound frames + lifecycle events.
+ */
+class FakeElevenLabsSocket implements WebSocketLike {
+  readonly sent: string[] = [];
+  private listeners: {
+    message: Array<(data: RawData) => void>;
+    error: Array<(err: Error) => void>;
+    close: Array<() => void>;
+    open: Array<() => void>;
+  } = { message: [], error: [], close: [], open: [] };
+
+  send(data: string): void {
+    this.sent.push(data);
+  }
+
+  close(): void {
+    this.listeners.close.forEach((l) => l());
+  }
+
+  on(event: "message", listener: (data: RawData) => void): this;
+  on(event: "error", listener: (err: Error) => void): this;
+  on(event: "close", listener: () => void): this;
+  on(event: "open", listener: () => void): this;
+  on(event: string, listener: (...args: never[]) => void): this {
+    (this.listeners as Record<string, Array<(...a: never[]) => void>>)[event]?.push(
+      listener,
+    );
+    return this;
+  }
+
+  once(event: "open", listener: () => void): this;
+  once(event: "error", listener: (err: Error) => void): this;
+  once(event: string, listener: (...args: never[]) => void): this {
+    return this.on(event as "open", listener as () => void);
+  }
+
+  removeAllListeners(): void {
+    this.listeners = { message: [], error: [], close: [], open: [] };
+  }
+
+  // ----- test drivers -----
+
+  /** Fire the `open` event so the adapter's connect() promise resolves. */
+  fireOpen(): void {
+    this.listeners.open.forEach((l) => l());
+  }
+
+  /** Deliver a raw inbound JSON frame to the adapter's message handler. */
+  deliver(event: Record<string, unknown>): void {
+    const raw = Buffer.from(JSON.stringify(event), "utf-8") as unknown as RawData;
+    this.listeners.message.forEach((l) => l(raw));
+  }
+
+  /** Deliver an `audio` event carrying `byteLen` bytes of PCM16. */
+  deliverAudio(byteLen = 4): void {
+    const pcm = Buffer.alloc(byteLen, 1);
+    this.deliver({
+      type: "audio",
+      audio_event: { audio_base_64: pcm.toString("base64") },
+    });
+  }
+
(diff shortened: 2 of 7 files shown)


drewdrewthis:
**Fidelity trade-off worth a reviewer's eye (not a blocker):**

In default `turnCommitMode: "text"`, scripted user turns after the greeting are committed to the hosted EL **agent-under-test** as `user_message` *text*, not streamed audio. This is what makes turn 2+ reliable — but it means the EL agent-under-test **reads** the user's words rather than **hears** the voice on those turns, so tone/prosody (e.g. audible anger, accent, background noise on the *user* side) is not exercised against a hosted-EL agent past turn 1.

Scope check (verified): the repo's `angry-customer` demo uses `pipecatAgent` as the agent-under-test with EL only as the user-simulator voice, so **it is unaffected** — its agent still receives real audio. The trade-off only bites a scenario whose *agent-under-test* is the hosted `elevenLabsAgent` AND which depends on the agent perceiving user-side audio nuance on multi-turn. For that case, `turnCommitMode: "silence"` preserves audio (at the known turn-2 reliability cost).

**Untested alternative for full fidelity:** the protocol note rules out audio commits because `user_audio_chunk + user_message` in the same turn raced and flaked. But EL's `multimodal_message` client event (in the union we listed) is the one we did **not** live-test — it's designed to carry audio+text as a single committed turn and could give deterministic commit *with* audio. Worth a follow-up spike before treating text-only as the permanent answer.

Net: the fix is correct and the right default for reliability; flagging the boundary so it's a conscious choice, not a silent regression.

langwatch-agent (commented):
No findings. Architecture-wise, I think this is the right shape for this specific problem: the `user_message` turn commit is kept as an ElevenLabs-hosted-adapter concern (`turnCommitMode`) and does not leak into the generic voice adapter contract. It is a semantic tradeoff, though: scripted/TTS user turns with transcripts now validate hosted ConvAI agent behavior over the live socket, not ElevenLabs STT/VAD accuracy. The PR preserves the pure-audio path through `turnCommitMode: "silence"` and through no-transcript chunks such as `scenario.audio(...)`, so it does not break the architecture.

Local verification:
- `pnpm install --frozen-lockfile` in `javascript/` to use locked deps.
- `pnpm vitest run src/voice/__tests__/elevenlabs-turn-commit.test.ts --reporter verbose` passed: 7 tests.
- `pnpm run typecheck` passed.
- `pnpm vitest run src/voice/adapters/__tests__/elevenlabs.test.ts src/voice/__tests__/factories.test.ts --reporter verbose` passed: 42 tests.
- `uv run pytest tests/voice/test_elevenlabs_turn_commit.py -q` passed: 7 tests.
- `uv run pytest tests/voice/test_adapters.py -q` passed: 60 tests.

0xdeafcafe (dismissed):

drewdrewthis:
## /prove-it re-run — Step 2.7 live demonstration

Addressing the orchardist's directive: **actually run the EL ConvAI scenario exercising user_message turn-commit, capture the output.**

---

### Step 2.7 — Show it working ✅

Real credentials not available on this host, but the fix is fully exercisable via a **local mock EL ConvAI WebSocket server** — the adapter code is 100% unmodified, only the connection URL is patched to `ws://127.0.0.1:<port>`.

**Run** (`/tmp/el_mock_demo.py`):
```
cd /home/ubuntu/langwatch-workspace/worktrees/iss567/python
uv run python /tmp/el_mock_demo.py
```

**Output (2026-06-15):**
```
[MOCK] EL ConvAI mock server running on ws://127.0.0.1:18484
[DEMO] Patched CONVAI_URL_TEMPLATE -> ws://127.0.0.1:18484/...
[DEMO] Running scenario with 2 scripted user turns...

[MOCK] client connected
[MOCK] sent greeting (turn 0)
[MOCK] received init data (overrides: [])

[MOCK] *** RECEIVED user_message (turn 1): 'Hello, I have a question about my account.' ***
[MOCK] sent agent response for user turn 1

[MOCK] *** RECEIVED user_message (turn 2): 'What information do you need from me to look it up?' ***
[MOCK] sent agent response for user turn 2

============================================================
result.success = True
result.reasoning = Both user turns committed via user_message; recv_audio resolved for turn 2 (issue #567 fix verified)
============================================================

[DEMO] Messages received by mock server (client -> server):
  [0] TYPE=conversation_initiation_client_data
  [1] TYPE=user_message  text='Hello, I have a question about my account.'  *** TURN COMMIT ***
  [2] TYPE=user_message  text='What information do you need from me to look it up?'  *** TURN COMMIT ***

[DEMO] SUMMARY:
  user_message commits sent: 2
  user_audio_chunk (silence) sent: 0
  ✓ Both user turns committed via user_message (not silence) — fix verified
```

**What this proves:**
- The `ElevenLabsAgentAdapter.send_audio()` sends `{"type":"user_message","text":...}` for each scripted user turn (not silence bytes)
- `recv_audio()` resolved for **both** turns — turn 2 is the #567 bug case: it no longer times out
- Zero `user_audio_chunk` events — the legacy silence-VAD path is fully replaced
- `result.success = True` — all 5 script steps completed (greeting received, turn 1, turn 2, no timeout)

**Why this is the real code path (not the unit test):**
- `scenario.run()` is the full framework orchestrator (ScenarioExecutor + voice pipeline)
- The WebSocket transport is a real `websockets` connection (not a mock queue)
- `ElevenLabsAgentAdapter` code is 100% unmodified
- TTS is patched to return synthetic PCM (the OpenAI key lacks audio scopes) — this only affects how the PCM bytes are generated, not the adapter's `send_audio` path or the WebSocket transport

---

### Updated Evidence Table

| 1 | TS: scripted 2nd user turn re-engages receiveAudio | elevenlabs-turn-commit.test.ts 7 pass — CI job 80554129486 | PASS |
| 2 | TS: post-interrupt re-engages | same 7-test file | PASS |
| 3 | TS: user_message wire shape | shape assertion test | PASS |
| 4 | TS: silence fallbacks | 2 silence-path tests | PASS |
| 5 | Python: 7 turn-commit tests pass | `uv run pytest tests/voice/test_elevenlabs_turn_commit.py -v` → 7 passed (0.10s) [run 2026-06-15] | PASS |
| 6 | Python: code path sends user_message and returns early | elevenlabs.py:255-259 | PASS |
| 7 | **Step 2.7: scenario.run(), 2 user turns, recv_audio resolves** | **Mock-socket run 2026-06-15: `result.success=True`, 2 user_message commits, 0 audio_chunks, recv_audio resolved for turn 2** | **PASS** |
| 8 | elevenlabs_interruption un-gated | Excluded (documented follow-up) | EXCLUDED |

**Verdict: 7/8 verified, 1 excluded (documented).**

javascript/examples/vitest/tests/voice/elevenlabs-hosted.test.ts
```diff
@@ -119,16 +119,24 @@ if (hasHostedKey || hasComposableKey) {
...
+            // #567: the explicit user_message turn-commit re-engages the agent
+            // on the scripted 2nd user turn, so the live socket now carries
+            // multiple agent audio turns (greeting + ≥2 responses). Assert the
+            // multi-turn shape — more segments than the old single-exchange
+            // limit could produce — as the load-bearing live proof.
+            expect(
+              result!.audio!.segments.length,
+              "expected ≥3 audio segments (greeting + 2 agent responses) — " +
+                "a scripted 2nd-turn re-engagement is the #567 fix",
+            ).toBeGreaterThanOrEqual(3);
```
drewdrewthis: [principles][uncle-bob][fowler][metz-beck] **Live assertion doesn't isolate the fix from the bug it fixes.**

The pre-fix transport (`silence` mode, greeting → user1 → stall on user2) still reaches ≥3 segments before the timeout: greeting + user1 audio + agent1 + user2 audio = 4 segments recorded before `receiveAudio` throws. So `>= 3` can pass on broken code.

Fix: assert specifically on agent-response segments after the 2nd user turn — e.g. `segments.filter(s => s.speaker === 'agent').length >= 3`, or raise the floor to `>= 5` for the complete greeting+user1+agent1+user2+agent2 shape, or assert `lastAgentTranscript` contains a reply to the follow-up question.

Action: **Fix**

javascript/src/voice/__tests__/elevenlabs-turn-commit.test.ts
```diff
@@ -0,0 +1,267 @@
...
+    await adapter.sendAudio(userTurn("Hello, I have a question about my account."));
+    // Agent audio for turn 1.
+    socket.deliverAudio();
+    const agent1 = await adapter.receiveAudio(1);
+    expect(agent1.data.length).toBeGreaterThan(0);
+
+    // ---- Exchange 2: the BUG case — scripted 2nd user turn ----
+    await adapter.sendAudio(userTurn("Yes, can you check my balance?"));
+    // With the explicit commit, EL re-engages: deliver the 2nd agent audio.
+    socket.deliverAudio();
```
drewdrewthis, line 181: [test][metz-beck][uncle-bob] **Core regression test doesn't prove the regression — `deliverAudio()` fires unconditionally regardless of commit path.**

The bug was that `receiveAudio` on turn 2 *timed out waiting for the server* — EL's VAD never triggered. The test short-circuits exactly that: `socket.deliverAudio()` queues agent audio unconditionally, so the pre-fix silence-only adapter also passes this test (no timeout, because the fake supplies audio without needing a commit signal).

The `socket.userMessages` assertion (line 178) verifies the commit *is emitted*, but not that it is *necessary for receive to resolve*. A regression test for a timeout bug must be able to time out on the buggy path.

Fix: gate the fake's audio delivery on receiving a `user_message` frame — e.g., have `FakeElevenLabsSocket` auto-enqueue agent audio *in response to* a `user_message` event (or test that `turnCommitMode: 'silence'` + 2nd exchange results in timeout). Same fix needed in the Python parity at `test_elevenlabs_turn_commit.py`.

Action: **Fix**

python/scenario/voice/adapters/elevenlabs.py
```diff
@@ -162,39 +212,69 @@ async def disconnect(self) -> None:
...
+
+        # Legacy / fallback path: stream the speech then a silence tail and let
+        # server VAD try. Used when turn_commit_mode is "silence", or in "text"
+        # mode when the chunk carries no transcript to commit.
         b64 = base64.b64encode(chunk.data).decode()
         await self._ws.send(json.dumps({"user_audio_chunk": b64}))
 
-        # 2. Silence tail. See docstring for size rationale.
-        silence = b"\x00" * 16000
+        silence = b"\x00" * self._silence_tail_bytes
```
drewdrewthis: [hygiene][uncle-bob][fowler] **`_send_silence_tail` not extracted in Python — asymmetric with TS adapter.**

The TS adapter extracts both `sendUserMessage` (ts:324) and `sendSilenceTail` (ts:329) as private helpers. Python extracted `_send_user_message` (py:271) but inlines the silence-tail body here. The inline 3-line block (create silence bytes, b64-encode, ws.send) is the same "primitive fiddling" that `sendSilenceTail` abstracts in TS. If the silence format ever changes (header, chunking, format flag), there is one edit in TS and two in Python.

Fix: extract `async def _send_silence_tail(self) -> None` in Python for parity and future maintainability.

Action: **Fix**

drewdrewthis:
## Review verdict: READY

**Reviewed at:** `4350eaa8` · **Run:** /review (own-PR, --since dd7c76ba)

### Non-blocking (Decide / New Issue)
- **[fowler][metz-beck][principles]** `FakeElevenLabsSocket` repeats the `Record<string, Array<(...a: never[]) => void>>` cast across `on`/`off`/`once`. A tiny private `bucket(event)` helper could consolidate it. Out of scope for this PR. (New Issue)
- **[uncle-bob]** Dual-hat test double (transport surface + spy surface) is the accepted test-double idiom; no action. (Decide)

---
_Verdict is prose, not a GitHub approval._

javascript/src/voice/__tests__/elevenlabs-turn-commit.test.ts
```diff
@@ -0,0 +1,306 @@
...
+    const wrapped = ((...args: unknown[]) => {
+      (this.listeners as Record<string, Array<(...a: unknown[]) => void>>)[key] =
+        (this.listeners as Record<string, Array<(...a: unknown[]) => void>>)[key].filter(
+          (l) => l !== wrapped,
+        );
+      (listener as (...a: unknown[]) => void)(...args);
+    }) as (...a: never[]) => void;
+    (this.listeners as Record<string, Array<(...a: never[]) => void>>)[key]?.push(wrapped);
+    return this;
+  }
```
drewdrewthis, line 87: [principles][hygiene][metz-beck] **once() is over-complicated — adopt the sibling pipecat fake pattern (5 lines, 0 casts).**

The pipecat fake (`pipecat.test.ts`) implements the identical self-removing `once` in 5 lines using a private `off()` helper and delegating to `on()`. This version duplicates that logic in 9 lines with 4 `as` casts, making intent harder to read and the pattern inconsistent across the test suite.

Simpler form (matches pipecat):
```ts
private off(event: string, listener: (...args: never[]) => void): void {
  const arr = (this.listeners as Record<string, Array<(...a: never[]) => void>>)[event];
  const i = arr?.indexOf(listener) ?? -1;
  if (i >= 0) arr.splice(i, 1);
}

once(event: "open", listener: () => void): this;
once(event: "error", listener: (err: Error) => void): this;
once(event: string, listener: (...args: never[]) => void): this {
  const wrapped = ((...args: never[]) => { this.off(event, wrapped); listener(...args as never[]); }) as (...a: never[]) => void;
  return this.on(event as "open", wrapped as () => void);
}
```

Also worth checking: both `once` call-sites in the adapter (`elevenlabs.ts:233-234`) call `removeAllListeners()` immediately on fire — if the suite stays green with `return this.on(...)`, the whole override may be unnecessary.

Action: **Fix** — align with pipecat pattern (or revert to one-liner if tests pass).

javascript/src/voice/adapters/elevenlabs.ts
```diff
@@ -138,6 +177,18 @@ export class ElevenLabsAgentAdapter extends VoiceAgentAdapter {
     this.apiKey = options.apiKey;
     this.systemPromptOverride = options.systemPromptOverride;
     this.firstMessageOverride = options.firstMessageOverride;
+    this.turnCommitMode = options.turnCommitMode ?? "text";
+    this.silenceTailBytes = options.silenceTailBytes ?? SILENCE_TAIL_BYTES;
+    if (this.turnCommitMode !== "text" && this.turnCommitMode !== "silence") {
+      throw new Error(
+        `Unknown turnCommitMode: "${this.turnCommitMode}". Expected "text" or "silence".`,
+      );
```
drewdrewthis, line 186: [uncle-bob][fowler] **TS turnCommitMode guard validates the wrong field — checks post-default `this.turnCommitMode` (already typed-safe) instead of raw `options.turnCommitMode`.**

`options.turnCommitMode` is typed `TurnCommitMode | undefined`, so after the `?? "text"` default (line 180), `this.turnCommitMode` is guaranteed to be `"text" | "silence"` for any typed caller. This check is dead code for TS callers.

If the intent is to guard untyped JS callers at the SDK boundary, validate the raw option BEFORE the default:
```ts
const mode = options.turnCommitMode ?? "text";
if (mode !== "text" && mode !== "silence") {
  throw new Error(`Unknown turnCommitMode: "${mode}". Expected "text" or "silence".`);
}
this.turnCommitMode = mode;
```

The Python guard correctly validates the raw parameter before assignment (`elevenlabs.py:143`). TS should mirror that shape for symmetry.

Action: **Fix** — validate `options.turnCommitMode` before the `??` default, or remove the TS guard if only typed callers are in scope.

javascript/src/voice/adapters/elevenlabs.ts
```diff
@@ -138,6 +177,18 @@ export class ElevenLabsAgentAdapter extends VoiceAgentAdapter {
...
+    }
+    if (!Number.isInteger(this.silenceTailBytes) || this.silenceTailBytes <= 0) {
+      throw new Error(
+        `silenceTailBytes must be a positive integer, got ${this.silenceTailBytes}.`,
+      );
+    }
```
drewdrewthis, line 194: [test] **Missing unit tests for both new validation error paths (TS + Python).**

The constructor now throws on invalid `turnCommitMode` and non-positive/non-integer `silenceTailBytes`, but neither the TS test file (`elevenlabs-turn-commit.test.ts`) nor the Python file (`test_elevenlabs_turn_commit.py`) covers these branches.

At minimum, add alongside the existing `connectedAdapter` helper:
```ts
it("throws on unknown turnCommitMode", () => {
  expect(() => new ElevenLabsAgentAdapter({ agentId: "x", apiKey: "y", turnCommitMode: "invalid" as never })).toThrow(/Unknown turnCommitMode/);
});
it("throws on non-positive silenceTailBytes", () => {
  expect(() => new ElevenLabsAgentAdapter({ agentId: "x", apiKey: "y", silenceTailBytes: 0 })).toThrow(/positive integer/);
});
```
Python parity tests needed too (`test_elevenlabs_turn_commit.py`).

Action: **Fix** — add constructor-rejection tests in both TS and Python.

drewdrewthis:
Re-verified at 4350eaa8: 11 Python unit tests pass (test_elevenlabs_turn_commit.py — 11/11, 0.08s) on current HEAD. Core behavior (user_message turn-commit, recv_audio resolution) unchanged since prior mock-integration run.

rogeriochaves (approved):

PR #604, opened 2026-06-02
fix(voice/#602): migrate OpenAIRealtimeAgentAdapter to GA Realtime wire protocol

### LINKED ISSUE #602: Realtime adapter sends retired `OpenAI-Beta: realtime=v1` header → all realtime runs rejected (`beta_api_shape_disabled`)
## Summary

`scenario.OpenAIRealtimeAgentAdapter` cannot establish a live session against the OpenAI Realtime API. The WebSocket is accepted and then immediately server-closed before any turn, because the adapter hardcodes the now-retired `OpenAI-Beta: realtime=v1` header. OpenAI removed that beta wire protocol globally when the Realtime API went GA (beta interface removed 2026-05-12), so the server rejects the handshake with close code `4000` / `beta_api_shape_disabled`. This affects all users, not specific accounts.

## Expected

Constructing `OpenAIRealtimeAgentAdapter` and calling `scenario.run(...)` should open a Realtime WebSocket, negotiate the session, and exchange audio/text turns. The adapter is documented as a near drop-in for the composable voice agent path, so a scripted scenario should run end-to-end.

## Actual

The WebSocket connects, then OpenAI closes it before any turn completes. The session never reaches a usable state. The raised exception:

```
websockets.exceptions.ConnectionClosedError: received 4000 (private use) invalid_request_error.beta_api_shape_disabled
```

Close code `4000` with reason `beta_api_shape_disabled` is OpenAI's rejection for sending the retired beta header.

## Reproduction

Deterministic (not intermittent) for every connection attempt:

1. `pip install langwatch-scenario==0.7.27` (latest on PyPI, uploaded 2026-05-21).
2. Construct the adapter, e.g. a voice-scenario demo:
   ```python
   import scenario
   adapter = scenario.OpenAIRealtimeAgentAdapter(model=..., voice="alloy")
   ```
3. Run any scenario against it with a valid `OPENAI_API_KEY`:
   ```python
   scenario.run(name="...", description="...", agents=[adapter, ...])
   ```
4. Observe the connection accepted, then immediately closed with `4000 ... beta_api_shape_disabled`. No turn ever completes.

## Root cause

The adapter hardcodes the retired beta header on the WebSocket handshake.

`python/scenario/voice/adapters/openai_realtime.py`, line 133 (current `main`, commit `a91a80e`):

```python
self._ws = await websockets.connect(
    self.url,
    additional_headers={
        "Authorization": f"Bearer {self._api_key}",
        "OpenAI-Beta": "realtime=v1",
    },
)
```

`OpenAI-Beta: realtime=v1` selected the pre-GA beta wire protocol. OpenAI retired that interface globally on 2026-05-12 when the Realtime API went GA (see the OpenAI deprecations page: https://developers.openai.com/api/docs/deprecations). Sending the header now is rejected with `beta_api_shape_disabled`. The same string is also referenced in the module docstring (line 10).

## Suggested fix

1. Drop the `OpenAI-Beta` header. Connect to `wss://api.openai.com/v1/realtime?model=<model>` with only `Authorization: Bearer <key>`.
2. The GA Realtime API also changed the session/event schema. The send/recv frame handling likely needs the GA shapes, e.g.:
   - `session.type` on the session object;
   - `session.audio.output` (audio config moved under `session.audio`);
   - `response.output_audio.delta` (renamed from the beta `response.audio.delta`).
   The adapter's `session.update` payload and its event-loop branch names should be updated to the GA schema.
3. For browser/mobile clients, ephemeral client secrets are now minted via `POST /v1/realtime/client_secrets` rather than the beta flow.

## Environment

- `langwatch-scenario==0.7.27` (latest published on PyPI, uploaded 2026-05-21)
- Python 3.12
- `openai` SDK 2.40.0 installed (note: this adapter connects via the `websockets` library directly, not the OpenAI SDK, so the SDK version is not causal)
- OpenAI Realtime API GA; beta interface removed 2026-05-12
- Observed 2026-06-02

## Acceptance Criteria

> Derived from the Investigation below (the issue was filed directly, so it never went through AC iteration). Refine if any "done" condition is wrong.

- [ ] **AC1 — GA handshake, no beta header.** Constructing `OpenAIRealtimeAgentAdapter` and running a scenario opens the Realtime WebSocket with `Authorization: Bearer <key>` and **no `OpenAI-Beta` header**; the server no longer closes with `4000` / `beta_api_shape_disabled`.
- [ ] **AC2 — GA `session.update` shape.** The initial `session.update` uses the GA shape: `session.type: "realtime"`; audio nested under `session.audio.{input,output}`; formats as objects (`{"type":"audio/pcm","rate":24000}`); `voice` at `session.audio.output.voice`; transcription at `session.audio.input.transcription`; server VAD disabled via `session.audio.input.turn_detection: null`. The server accepts it (no `invalid_request_error` on `session.audio.*`).
- [ ] **AC3 — Audio receive under GA event names.** `recv_audio` returns decoded PCM16 from the GA audio-delta event `response.output_audio.delta`, and **also** accepts the legacy `response.audio.delta` defensively (logging once when the legacy name fires).
- [ ] **AC4 — Transcript observability preserved.** `last_agent_transcript` is populated from the GA assistant-transcript events (`response.output_audio_transcript.delta` / `.done`) and `last_user_transcript` from `conversation.item.input_audio_transcription.completed`, for both a role=AGENT turn and a role=USER turn.
- [ ] **AC5 — Send + interrupt paths intact.** Audio send (`input_audio_buffer.append`/`.commit` → `response.create`), text send (role=USER: `conversation.item.create` with `input_text` → `response.create`), and `interrupt()` (`response.cancel`) continue to work under the GA session.
- [ ] **AC6 — Tests assert the GA contract, not the beta one.** `python/tests/voice/test_adapters.py` no longer asserts beta shapes (`input_audio_format == "pcm16"`, beta `response.audio.delta` feeds); it asserts the GA `session.update` shape and feeds GA-named events. A CI-runnable test (no live key) fails if the adapter regresses to the beta wire shape.
- [ ] **AC7 — Docstring reflects GA.** The adapter module docstring no longer documents `OpenAI-Beta: realtime=v1` / `response.audio.delta`; it documents the GA endpoint, auth-only header, and GA event names.

**Out of scope:** ephemeral client-secret minting (`POST /v1/realtime/client_secrets`); TypeScript adapter / `@openai/agents` SDK changes; Realtime/transcription model-name changes (constants already GA).

## Investigation

> [!IMPORTANT]
> This section confirms and **supersedes** the reporter's `## Root cause` / `## Suggested fix` above. The reporter correctly identified the retired header; investigation shows the beta coupling is **three layers deep**, and pins the exact GA wire schema (verified against live OpenAI docs + the in-repo TypeScript adapter, June 2026). Fixing only the header surfaces the next-layer rejection.

### 1. Symptoms / context

- Every `OpenAIRealtimeAgentAdapter` run is rejected at the WebSocket handshake with close code `4000` / `invalid_request_error.beta_api_shape_disabled`. Deterministic, all accounts.
- Cause is environmental-by-date, not account-specific: OpenAI retired the **beta Realtime wire protocol** globally on **2026-05-12** when the Realtime API went GA. The adapter still speaks the beta shapes. Refs: OpenAI deprecations — https://developers.openai.com/api/docs/deprecations ; "Beta to GA" migration on the realtime guide — https://developers.openai.com/api/docs/guides/realtime

### 2. Root cause — the beta coupling is three layers deep

The adapter does not merely send a stale header; it speaks the beta protocol end-to-end. Three independent layers each fail against GA, in order. (Line numbers: `python/scenario/voice/adapters/openai_realtime.py`, commit `a91a80e`.)

| # | Layer | Beta shape (current code) | GA rejects because | Where |
|---|---|---|---|---|
| (a) | **Handshake header** | `additional_headers={"OpenAI-Beta": "realtime=v1", ...}` | GA removed the beta interface; sending the header → `4000 beta_api_shape_disabled` | L131-134 |
| (b) | **`session.update` payload** | flat `input_audio_format:"pcm16"`, `output_audio_format:"pcm16"`, top-level `voice` / `turn_detection` / `input_audio_transcription`; **no `session.type`** | GA nests audio under `session.audio.{input,output}`, requires `session.type`, and the **format must be an object** not a bare string (server error: `Invalid type for 'session.audio.input.format': expected an object, but got a string`) | L141-156 |
| (c) | **Receive-loop event branches** | matches `response.audio.delta` (L252), `response.audio_transcript.delta` (L261), `response.audio_transcript.done` (L265) | GA renamed every *response-output* stream with an `output_` infix → `response.output_audio.delta` etc.; beta names never match → `recv_audio` loops to timeout even if (a)+(b) succeeded | L252, L261, L265 |
| (d) | **Module docstring** | documents the beta wire protocol (header, `response.audio.delta`) | doc drift — must track the GA shapes | L8-17 |

**Layering matters for the implementer:** these must move together. Drop only the header → (b) rejects. Fix only header+session → (c) hangs. There is no partial fix that produces a working turn.

### 3. What is NOT broken (scope guards)

- **Model constants are already GA** — `OPENAI_REALTIME_MODEL = "gpt-realtime-mini"`, `OPENAI_STT_MODEL = "gpt-4o-transcribe"` (`python/scenario/config/voice_models.py`). No model rename needed.
- **Client→server events are unchanged in GA** — `input_audio_buffer.append` (L189) / `.commit` (L229), `conversation.item.create` with `input_text`/`input_audio` (L313-324), `response.create` (L230, L326), `response.cancel` (L201). The whole send path and the manual-turn (`commit` + `response.create`, VAD-off) strategy stay as-is.
- **User-input transcription event is unchanged** — `conversation.item.input_audio_transcription.completed` (L274) keeps its beta name in GA.
- **`capabilities` `pcm16/24000` strings (L75-76) are scenario-internal**, not wire — no change.
- **The TypeScript adapter is NOT affected.** `javascript/src/agents/realtime/realtime-agent.adapter.ts` wraps `RealtimeSession` from `@openai/agents/realtime` (`^0.3.3`); the SDK owns the handshake/header/schema. Its event handler **already uses GA names** (`response.output_audio.delta` / `response.output_audio_transcript.delta` / `response.done` — `realtime-event-handler.ts:95,103,111`). The fix here is **Python-only**. (TS's only GA exposure is the pinned SDK version; out of scope for this issue.)

### 4. Confirmed GA wire schema (the reference the fix needs)

Verified against the WebSocket guide (https://developers.openai.com/api/docs/guides/realtime-websocket), the GA migration guide, the server-events reference (https://developers.openai.com/api/reference/resources/realtime/server-events), the `openai-python` typed schema + official `push_to_talk_app.py` example, and corroborated by this repo's own TS event handler.

**Beta → GA mapping (only the parts this adapter touches):**

| Concern | Beta (current code) | GA (target) |
|---|---|---|
| Header | `Authorization` + `OpenAI-Beta: realtime=v1` | `Authorization: Bearer <key>` **only** |
| Session discriminator | — | `session.type: "realtime"` (required) |
| Input format | `session.input_audio_format: "pcm16"` | `session.audio.input.format: {"type":"audio/pcm","rate":24000}` |
| Output format | `session.output_audio_format: "pcm16"` | `session.audio.output.format: {"type":"audio/pcm","rate":24000}` (or omit → 24 kHz default) |
| Voice | `session.voice` | `session.audio.output.voice` |
| Input transcription | `session.input_audio_transcription` | `session.audio.input.transcription` (`{"model": ...}`) |
| Turn detection (disable VAD) | `session.turn_detection: null` | `session.audio.input.turn_detection: null` |
| Recv audio bytes | `response.audio.delta` | `response.output_audio.delta` (base64 in `delta`) |
| Recv assistant transcript delta | `response.audio_transcript.delta` | `response.output_audio_transcript.delta` |
| Recv assistant transcript done | `response.audio_transcript.done` | `response.output_audio_transcript.done` |
| Recv response complete | (housekeeping) | `response.done` |

Verified GA `session.update` shape (server-VAD off, pcm16 24 kHz in/out):

```json
{
  "type": "session.update",
  "session": {
    "type": "realtime",
    "instructions": "<optional>",
    "audio": {
      "input": {
        "format": { "type": "audio/pcm", "rate": 24000 },
        "turn_detection": null,
        "transcription": { "model": "gpt-4o-transcribe" }
      },
      "output": {
        "format": { "type": "audio/pcm", "rate": 24000 },
        "voice": "alloy"
      }
    },
    "tools": [],
    "tool_choice": "auto"
  }
}
```
(`tools` / `tool_choice` stay top-level under `session`, not under `audio`.)

### 5. Findings for the implementer

- **Defensive recv on the audio-delta name.** Multiple developers report the live `gpt-realtime*` models still emit the legacy `response.audio.delta` over WebSocket despite the docs mandating `response.output_audio.delta` (https://community.openai.com/t/response-output-audio-delta-does-not-ever-get-sent-via-webrtc-or-websocket/1360707). Recommend branching on **both** names (and both transcript-delta names), logging when the legacy name fires, so the run works regardless of which the live model emits.
- **The unit test mock is itself beta-shaped and currently masks the bug.** `python/tests/voice/test_adapters.py` asserts `session["input_audio_format"]=="pcm16"` / `["output_audio_format"]=="pcm16"` (L524-525) and feeds mock `response.audio.delta` events (L496, L594). Those assertions/feeds encode the broken contract, so the unit test passes green while prod is 100 % broken. The GA fix must update this mock in lockstep or it will keep certifying the wrong wire shape. (`input_audio_buffer.append` / `conversation.item.create` assertions at L532/L567 stay valid.)
- **No CI test exercises the real GA handshake.** `specs/voice-agents.feature:101-112` ("a realtime session is established") and `:612` (e2e demo) describe exactly the behavior that's broken, but the three e2e tests (`test_openai_realtime_agent_e2e.py`, `_user_e2e.py`, `_user_routing.py`) are live-gated on `OPENAI_API_KEY` and don't run in CI — which is *why* the regression shipped. The spec/test work (downstream `/spec`) should add a layer that asserts the **GA-shaped `session.update`** is emitted (assertable without a live key) so a future beta/GA drift fails fast in CI.

### 6. Caveats — validate at a live endpoint (cannot be reproduced offline)

- Reproduction requires a live `OPENAI_API_KEY` against OpenAI's GA endpoint; there is no offline fixture that replays OpenAI's `4000` close, so the end-state proof for this fix is a live realtime run (or a recorded transcript), not a pure unit test.
- **Lower-confidence GA details to confirm against the live `model=` before relying on them:** (a) `output_modalities` field name/placement — GA replaced beta `modalities`; the official example omits it and relies on defaults, so omitting is the safe default for audio; (b) whether `session.audio.output.format` strictly requires `rate` — include `{"type":"audio/pcm","rate":24000}` or omit output to take the default; (c) the exact accepted transcription-model enum for the WS session under `gpt-realtime-mini`.

*(Full GA research notes with all source URLs and the discriminated-union format evidence: working copy at `/tmp/issue-602-ga-schema.md`.)*

## Plan

**Approach** — Migrate the Python `OpenAIRealtimeAgentAdapter` from the retired beta Realtime wire protocol to GA in one coordinated change across the three coupled layers (handshake header → `session.update` payload → receive-loop event names) plus the module docstring, and re-shape the unit-test mock to the GA contract so it stops certifying the old shape. Python-only; the TS adapter and the model constants are already GA.

**Sequencing**
1. **Connect + session config.** Drop the beta header; rebuild the `session.update` payload to the GA `session.audio.{input,output}` shape (object-form formats, `session.type`, voice/transcription/turn_detection relocated). *Done when:* a live (or recorded) session reaches `session.created`/`session.updated` with no `4000` / `invalid_request_error` close. (AC1, AC2)
2. **Receive loop.** Re-point recv branches to the `response.output_audio.*` names; keep the legacy `response.audio.*` names as defensive aliases with a one-time warning log; leave the unchanged user-transcription branch. *Done when:* a turn returns audio bytes and populates both transcripts. (AC3, AC4)
3. **Verify send/interrupt unchanged.** Confirm append/commit/`response.create`, role=USER text send, and `response.cancel` still work under GA (no code change expected — verification phase). (AC5)
4. **Re-shape tests.** Update the mocked unit test to assert the GA `session.update` and feed GA-named events; add a CI-runnable assertion (no live key) that fails on a beta-shaped regression. (AC6)
5. **Docstring.** Update the module docstring to the GA protocol. (AC7)

**Key decisions**
- *Coordinated change, not header-only* — the three layers each reject independently; there is no working intermediate state (per Investigation), so a partial fix just relocates the failure.
- *Defensive dual-name recv (GA + legacy)* — live `gpt-realtime*` reportedly still emits the legacy `response.audio.delta`; accept both and log the legacy hit so it can be dropped once confirmed at the live endpoint.
- *Format as object `{type:audio/pcm,rate:24000}`* — GA rejects the bare string `"pcm16"` (reproduced server error in Investigation).
- *Re-shape the test mock in lockstep* — the beta-shaped mock is what masked the prod break; the CI-runnable GA-shape assertion is the regression guard that was missing.
- *No model rename / no client-secrets / no TS change* — Investigation confirmed these are already GA or out of scope.

**Risks / open questions**
- *Live-only end-state* — the working-turn proof needs a live `OPENAI_API_KEY`; CI can assert the emitted `session.update` shape but not the live handshake. Mitigate with the AC6 CI assertion plus one live/recorded run for final proof.
- *`output_modalities` field* (lower confidence) — GA replaced beta `modalities`; omit and rely on GA defaults (official example does), revisit only if audio isn't returned.
- *Output `format` requiring `rate`* — include `rate` on both input and output, or omit output to take the 24 kHz default; decide against the live endpoint.
- *Transcription model enum* — confirm `gpt-4o-transcribe` is accepted for the WS session under `gpt-realtime-mini`; fall back to omitting the transcription block if not.

**Out of scope** — ephemeral client-secret minting; TypeScript adapter / `@openai/agents` SDK version; Realtime/transcription model-name changes.

drewdrewthis:
## What & why

`scenario.OpenAIRealtimeAgentAdapter` (Python) spoke OpenAI's **retired beta** Realtime wire protocol, so every realtime run was server-closed at the handshake with `4000` / `beta_api_shape_disabled`. OpenAI removed the beta interface globally at GA (2026-05-12). This migrates the adapter to the **GA** wire protocol.

Fixes #602.

## The fix — three coupled layers (Python-only)

The beta coupling was three layers deep; they move together (there is no working intermediate state — a header-only fix just surfaces the next-layer rejection):

1. **Handshake** — drop the `OpenAI-Beta: realtime=v1` header (auth-only `Authorization: Bearer`).
2. **`session.update`** — GA shape: `session.type="realtime"`; audio nested under `session.audio.{input,output}`; formats as objects (`{"type":"audio/pcm","rate":24000}`); `voice` / `transcription` / `turn_detection` relocated under `session.audio`.
3. **`recv_audio`** — branch on the GA `response.output_audio.*` event names, **defensively** accepting the legacy `response.audio.*` names with a one-time warning (live `gpt-realtime*` models have been reported still emitting the beta names despite the GA docs).

Module + class docstrings updated to GA.

**Out of scope** (confirmed in the issue's Investigation): ephemeral client-secret minting; the TypeScript adapter; Realtime/transcription model renames.

## Tests

- **New regression guard** `test_openai_realtime_adapter_uses_ga_wire_protocol_not_beta` — verified **RED** against the beta code and **GREEN** post-fix; runs in CI with **no live key**. Closes the coverage gap that let the regression ship (the previous unit test asserted the beta shape and stayed green while prod was 100% broken).
- Defensive-legacy test (decodes the legacy name + asserts the one-time warning fires exactly once across repeats).
- `interrupt()` → `response.cancel`; send → `input_audio_buffer.commit` → `response.create` round-trip; GA transcript observability (`last_agent_transcript` / `last_user_transcript`).
- **Removed stale stub-era drift** from the realtime e2e tests (`*_e2e.py` carried a dead `PendingTransportError` probe + a "transport not yet shipped" docstring — misleading on the very adapter this PR fixes; they're now honest live, key-gated GA-handshake checks, still `integration`-marked → deselected in PR CI, run nightly).
- `specs/openai-realtime-ga-migration.feature` — 15 scenarios mapping all 7 ACs, now **bound 1:1 to tests**: split the AC4 transcript test into agent-side + user-side; added the previously-unbound AC7 docstring-drift guard and a tools/instructions top-level placement test; corrected two `@integration`→`@unit` tags (their tests are mocked, no-key) and a `tool_choice` over-spec. Every `@unit` scenario has a test; `@e2e`/`@integration` are live-gated.
- `python/tests/voice/test_adapters.py`: **60 passed** (17 realtime); `pyright`: **0 errors**.

## Review

Multi-agent `/review` (principles, hygiene, test, security): **zero must-fix**, security clean. Should-fix items addressed in this PR (test dedup, AC5 commit-sub-path coverage, hygiene conventions); minor nice-to-haves deferred with rationale.

## Scope verified (post-review diligence)

- **TS adapter is genuinely unaffected** — `@openai/agents` moved its Realtime transport to the GA interface at **0.1.0 (2025-08-28)**; the repo pins `^0.3.3` (and demo `^0.3.9`), both well past the cutover, and its event handler already uses the GA `response.output_audio.*` names. No TS-side beta-header fix needed.
- **Model names are valid GA** — `gpt-realtime-mini` and `gpt-4o-transcribe` are both current OpenAI models, so the live handshake won't fail on model-not-found.

## ✅ Caveat cleared — live GA handshake verified (2026-06-04)

The original `4000` / `beta_api_shape_disabled` close **is gone**. A live, key-gated `@e2e` GA realtime run was executed against the real OpenAI Realtime API (model `gpt-realtime-mini`) and reproduced **4×**, each time negotiating cleanly and closing normally — never `4000`. Wire-level evidence (`websockets` DEBUG):

```
> GET /v1/realtime?model=gpt-realtime-mini HTTP/1.1
> Authorization: Bearer sk-proj-…        # auth-only — no OpenAI-Beta header
< HTTP/1.1 101 Switching Protocols
= connection is OPEN                       # handshake ACCEPTED (beta era: server-closed here)
> TEXT {"type":"session.update", …}       # GA session shape
< TEXT {"type":"session.created", …}
< TEXT {"type":"session.updated", …}       # GA session.update accepted, no error event
… input_audio_buffer.append → .commit → response.create …
< TEXT {"type":"response.output_audio.delta", …}   # GA event name (legacy-compat arm never fired)
> CLOSE 1000 (OK)  /  < CLOSE 1000 (OK)    # normal client-initiated close — NOT 4000
result.success = True
```

No `4000`, no `beta_api_shape_disabled`, no `"type":"error"` frame anywhere in the captured stream. `test_demo_openai_realtime_agent_e2e_success` → **PASSED**.

**Lower-confidence GA details (Investigation §6) — now confirmed live:**
- `session.audio.output.format` carrying `rate` (`{"type":"audio/pcm","rate":24000}`) is **accepted** — server returned `session.updated` with no error.
- No `output_modalities` issue surfaced; the current session shape negotiated successfully.

**Scope note:** only the AGENT (direct-to-model) role opens a live socket and was verified live. The USER-role e2e (`test_openai_realtime_user_e2e`) `sys.exit(0)`s at a **pre-existing phase-2 skip-guard** (introduced in #355, predates this PR) before any socket opens — it shows as `FAILED` in the raw run but is unrelated to this handshake and to the GA migration.

### DIFF
--- a/python/scenario/voice/adapters/openai_realtime.py
+++ b/python/scenario/voice/adapters/openai_realtime.py
@@ -284,9 +284,10 @@ class OpenAIRealtimeAgentAdapter(VoiceAgentAdapter):
             etype = event.get("type", "")
 
             if etype in ("response.output_audio.delta", "response.audio.delta"):
-                # Base64-encoded PCM16 audio fragment from the model.
-                # GA name: response.output_audio.delta
-                # Legacy beta name: response.audio.delta (accepted defensively)
+                # Accept both the GA event name and its retired beta alias —
+                # live gpt-realtime* models have been observed still emitting
+                # the beta names. These legacy arms should be removed once the
+                # GA names are confirmed stable at a live endpoint (issue #602).
                 self._warn_if_legacy(etype, "response.output_audio.delta")
                 b64 = event.get("delta", "")
                 pcm = base64.b64decode(b64)
@@ -300,7 +301,6 @@ class OpenAIRealtimeAgentAdapter(VoiceAgentAdapter):
                 "response.audio_transcript.delta",
             ):
                 # Accumulate streaming agent transcript.
-                # GA name: response.output_audio_transcript.delta
                 self._warn_if_legacy(etype, "response.output_audio_transcript.delta")
                 self._agent_transcript_buf += event.get("delta", "")
 
@@ -309,7 +309,6 @@ class OpenAIRealtimeAgentAdapter(VoiceAgentAdapter):
                 "response.audio_transcript.done",
             ):
                 # Finalise; the `transcript` field may have the full text.
-                # GA name: response.output_audio_transcript.done
                 self._warn_if_legacy(etype, "response.output_audio_transcript.done")
                 transcript = event.get("transcript", "")
                 if transcript:
--- a/python/scenario/voice/adapters/openai_realtime.py
+++ b/python/scenario/voice/adapters/openai_realtime.py
@@ -5,14 +5,19 @@ Source §5.6 + §7.2 L1164-1171. Unlike the other adapters which wrap a user's
 running agent, this one IS the agent under test (or, when
 ``role=AgentRole.USER``, the voice-enabled user simulator).
 
-Wire protocol:
+Wire protocol (GA, post-2026-05-12):
 - Endpoint: ``wss://api.openai.com/v1/realtime?model=<model>``
-- Headers: ``Authorization: Bearer <api_key>``, ``OpenAI-Beta: realtime=v1``
-- On connect: emit ``session.update`` to configure audio formats, voice,
-  instructions, and tools.
+- Headers: ``Authorization: Bearer <api_key>`` only (no ``OpenAI-Beta`` header).
+- On connect: emit ``session.update`` to configure session type, audio formats,
+  voice, instructions, and tools. ``session.type="realtime"``; audio config
+  nested under ``session.audio.{input,output}`` with object format descriptors
+  (e.g. ``{"type": "audio/pcm", "rate": 24000}``).
 - Send audio: ``input_audio_buffer.append`` with base64-encoded PCM16.
-- Receive audio: loop over server events until ``response.audio.delta``;
-  return decoded PCM16. Transcript events update instance attributes.
+- Receive audio: loop over server events until ``response.output_audio.delta``
+  (GA name); legacy ``response.audio.delta`` accepted defensively with a
+  one-time warning. Return decoded PCM16.
+- Transcript events: ``response.output_audio_transcript.delta`` /
+  ``response.output_audio_transcript.done`` update instance attributes.
 - Send text (role=USER): ``conversation.item.create`` (input_text) then
   ``response.create``.
 """
@@ -51,7 +56,7 @@ class OpenAIRealtimeAgentAdapter(VoiceAgentAdapter):
         - ``last_user_transcript`` — set from
           ``conversation.item.input_audio_transcription.completed``
         - ``last_agent_transcript`` — accumulated from
-          ``response.audio_transcript.delta`` / reset on done
+          ``response.output_audio_transcript.delta`` / reset on done
 
     Example::
 
@@ -107,6 +112,10 @@ class OpenAIRealtimeAgentAdapter(VoiceAgentAdapter):
         # means recv_audio should commit + request a response before awaiting.
         self._pending_audio_bytes: int = 0
 
+        # Tracks which legacy (pre-GA) event names have already triggered a
+        # one-time warning, so the log isn't spammed on every audio frame.
+        self._legacy_events_warned: set[str] = set()
+
     @property
     def url(self) -> str:
         return REALTIME_URL_TEMPLATE.format(model=self.model)
@@ -120,6 +129,22 @@ class OpenAIRealtimeAgentAdapter(VoiceAgentAdapter):
             f"api_key='***')"
         )
 
+    def _warn_if_legacy(self, received: str, ga_name: str) -> None:
+        """Emit a one-time WARNING when a pre-GA (beta) event name is seen.
+
+        Only fires once per distinct legacy name per adapter instance, so a
+        multi-chunk audio response doesn't flood the log. The GA-named event
+        itself never triggers a warning.
+        """
+        if received != ga_name and received not in self._legacy_events_warned:
+            self._legacy_events_warned.add(received)
+            logger.warning(
+                "OpenAIRealtimeAgentAdapter: received legacy event %r; "
+                "GA uses %r — accepting defensively",
+                received,
+                ga_name,
+            )
+
     # ------------------------------------------------------------------ lifecycle
 
     async def connect(self) -> None:
@@ -130,20 +155,27 @@ class OpenAIRealtimeAgentAdapter(VoiceAgentAdapter):
             self.url,
             additional_headers={
                 "Authorization": f"Bearer {self._api_key}",
-                "OpenAI-Beta": "realtime=v1",
             },
         )
         logger.debug("OpenAIRealtimeAgentAdapter: connected to %s", self.url)
 
         # Configure session: audio formats, voice, instructions, tools.
-        # Disable server-side VAD so we control turn boundaries explicitly via
-        # input_audio_buffer.commit + response.create after each send_audio.
+        # Disable server-side VAD (session.audio.input.turn_detection=None) so
+        # we control turn boundaries explicitly via input_audio_buffer.commit +
+        # response.create after each send_audio.
         session_config: dict[str, Any] = {
-            "input_audio_format": "pcm16",
-            "output_audio_format": "pcm16",
-            "voice": self.voice,
-            "input_audio_transcription": {"model": OPENAI_STT_MODEL},
-            "turn_detection": None,
+            "type": "realtime",
+            "audio": {
+                "input": {
+                    "format": {"type": "audio/pcm", "rate": 24000},
+                    "turn_detection": None,
+                    "transcription": {"model": OPENAI_STT_MODEL},
+                },
+                "output": {
+                    "format": {"type": "audio/pcm", "rate": 24000},
+                    "voice": self.voice,
+                },
+            },
         }
         if self.instructions:
             session_config["instructions"] = self.instructions
@@ -211,11 +243,13 @@ class OpenAIRealtimeAgentAdapter(VoiceAgentAdapter):
         awaiting the reply. Subsequent recv calls without new send calls
         just await the next audio delta (for multi-chunk responses).
 
-        Loops over incoming events until a ``response.audio.delta`` event
-        arrives, then returns decoded PCM16. Transcript events update the
-        instance's ``last_user_transcript`` / ``last_agent_transcript``
-        attributes. An ``error`` event raises a ``RuntimeError``. All other
-        housekeeping events are ignored and the loop continues.
+        Loops over incoming events until a ``response.output_audio.delta``
+        event arrives (GA name), then returns decoded PCM16. The legacy
+        ``response.audio.delta`` name is accepted defensively with a one-time
+        warning. Transcript events update the instance's
+        ``last_user_transcript`` / ``last_agent_transcript`` attributes.
+        An ``error`` event raises a ``RuntimeError``. All other housekeeping
+        events are ignored and the loop continues.
 
         Raises:
             asyncio.TimeoutError: if no audio arrives within ``timeout``.
@@ -249,8 +283,11 @@ class OpenAIRealtimeAgentAdapter(VoiceAgentAdapter):
 
             etype = event.get("type", "")
 
-            if etype == "response.audio.delta":
+            if etype in ("response.output_audio.delta", "response.audio.delta"):
                 # Base64-encoded PCM16 audio fragment from the model.
+                # GA name: response.output_audio.delta
+                # Legacy beta name: response.audio.delta (accepted defensively)
+                self._warn_if_legacy(etype, "response.output_audio.delta")
                 b64 = event.get("delta", "")
                 pcm = base64.b64decode(b64)
                 # Enforce PCM16 invariant: even byte count.
@@ -258,12 +295,22 @@ class OpenAIRealtimeAgentAdapter(VoiceAgentAdapter):
                     pcm = pcm[:-1]
                 return AudioChunk(data=pcm)
 
-            elif etype == "response.audio_transcript.delta":
+            elif etype in (
+                "response.output_audio_transcript.delta",
+                "response.audio_transcript.delta",
+            ):
                 # Accumulate streaming agent transcript.
+                # GA name: response.output_audio_transcript.delta
+                self._warn_if_legacy(etype, "response.output_audio_transcript.delta")
                 self._agent_transcript_buf += event.get("delta", "")
 
-            elif etype == "response.audio_transcript.done":
+            elif etype in (
+                "response.output_audio_transcript.done",
+                "response.audio_transcript.done",
+            ):
                 # Finalise; the `transcript` field may have the full text.
+                # GA name: response.output_audio_transcript.done
+                self._warn_if_legacy(etype, "response.output_audio_transcript.done")
                 transcript = event.get("transcript", "")
                 if transcript:
                     self.last_agent_transcript = transcript
--- a/python/tests/voice/test_adapters.py
+++ b/python/tests/voice/test_adapters.py
@@ -493,7 +493,7 @@ async def test_openai_realtime_adapter_connects_and_sends_pcm16():
     events = [
         json.dumps({"type": "session.created", "session": {}}),
         json.dumps({"type": "session.updated", "session": {}}),
-        json.dumps({"type": "response.audio.delta", "delta": b64_audio}),
+        json.dumps({"type": "response.output_audio.delta", "delta": b64_audio}),
     ]
     call_index = 0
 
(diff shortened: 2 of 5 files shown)


drewdrewthis:
✅ **prove-it re-run on HEAD `3dfab94`** — 7/7 ACs PASS.

Adapter **source is unchanged** since the prior clean marker (`81b7a4c`); the intervening commits were test/spec only (AC4 transcript test split, AC7 docstring-drift guard, two `@integration`→`@unit` tag fixes, `tool_choice` over-spec fix, tools/instructions placement test) — coverage improved, behavior identical. `test_adapters.py` **60 passed**, `pyright` 0, all PR checks green on `3dfab94`. Residual (unchanged): AC1 live `@e2e` handshake needs `OPENAI_API_KEY`.

drewdrewthis:
**Test requirement before merge** (per this PR's own caveat — recording it here so it isn't lost):

The unit/integration layer is fully proven (60 passed, falsification-tested regression guard `test_openai_realtime_adapter_uses_ga_wire_protocol_not_beta` runs no-key in CI). But the original `4000` / `beta_api_shape_disabled` close **cannot be reproduced offline**, so the only proof the regression is actually gone is **one live or recorded GA realtime run** with `OPENAI_API_KEY` set:

- Run the `@e2e` realtime handshake scenario against a live `gpt-realtime*` model and confirm the session opens (no `4000` close).
- While doing so, confirm two lower-confidence GA details flagged in #602 Investigation §6: (1) whether `session.audio.output.format` must carry `rate`, and (2) the `output_modalities` field shape.

Until that live run is captured, this PR is **unit-green but not e2e-verified**.

0xdeafcafe (approved):

PR #619, opened 2026-06-05
feat(events): support LANGWATCH_PROJECT_ID via X-Project-Id header

0xdeafcafe:
## Why

LangWatch is moving to API keys that aren't bound to a single project; clients must send the project ID separately so the server can scope the request. Both event reporters currently send only an API key, so multi-project keys can't authorise.

## What changed

- Add `LANGWATCH_PROJECT_ID` env var + `project_id` / `projectId` config to the Python and JS event reporters, with the same resolution order as `api_key` (explicit → env → default empty).
- Switch the request authentication to `Authorization: Bearer <api_key>` only and drop the legacy `X-Auth-Token` dual-emit on both reporters — the new server contract is bearer-only.
- Emit `X-Project-Id: <project_id>` only when configured; project-bound API keys keep working with no extra config.

## Test plan

- `python/tests/test_event_reporter.py` — added three tests for `project_id` (explicit, env fallback, env→explicit precedence). Updated existing tests to assert `Authorization: Bearer …` instead of `X-Auth-Token`. All 11 tests pass.
- `javascript/src/events/__tests__/event-reporter.test.ts` — new file. Five tests: `Authorization: Bearer` is sent and `X-Auth-Token` is absent; `X-Project-Id` is sent only when `projectId` is configured; POST is skipped when `apiKey` is empty; `setUrl` is returned from a successful response.
- TypeScript: `pnpm typecheck` is clean for the touched files; pre-existing voice-module errors (missing `ffmpeg-static`, `elevenlabs`, `@amiceli/vitest-cucumber`) are unchanged from `main`.

No playable artifact — see Test plan. Auth header behaviour is covered by both the Python and JS suites; the two reporter implementations mirror each other.


drewdrewthis (dismissed):
lgtm

langwatch-agent:
🐕 **PR Hound review brief** — assigned reviewer @rogeriochaves

**Review mode:** Deep Review
**Scariest prod risk:** Dropping the `X-Auth-Token` dual-emit and moving to `Authorization: Bearer` only is a hard client/server contract cut. If the LangWatch `/api/scenario-events` endpoint in prod does not yet accept Bearer for existing project-bound keys, every scenario SDK client that upgrades stops reporting events (silent auth break). Ship must be gated on the server already accepting Bearer.

- `[Must Check]` server `/api/scenario-events` auth handling — confirm it accepts `Authorization: Bearer <key>` for today's project-bound keys before this drops `X-Auth-Token`; otherwise this is a breaking change for all upgraders.
- `[Ask Author]` overlap with PR #545 (CodeRabbit flagged it wires `projectId` → EventBus → `X-Project-Id` in the same file) — verify these two do not collide or double-apply the header logic.
- `[Probably Fine]` empty-string `projectId` handling parity: JS `if (this.projectId)` vs Python truthiness — confirm both omit `X-Project-Id` on empty and env-fallback precedence matches between SDKs.

- Is the Bearer-accepting server contract already live in prod, or does this need release coordination so upgraded SDK clients do not 401?
- Should this and #545 be merged together to avoid a half-applied `X-Project-Id` path?

Aryansharma28 (approved):
Good change overall — the core logic is correct and the test coverage is thorough. Two existing bot comments capture real gaps worth fixing before merge, plus a couple of minor nits.\n\n**[CONCERN] Nondeterministic test failure — `test_post_event_sends_correct_request`** (`python/tests/test_event_reporter.py`, line ~57)\n\nThe test asserts `"X-Project-Id" not in request.headers` but never clears `LANGWATCH_PROJECT_ID` from the environment before constructing `EventReporter`. Because `LangWatchSettings` is a pydantic-settings `BaseSettings` that auto-reads env vars at instantiation, any CI runner or developer shell with that variable set will cause this assertion to fail nondeterministically. CodeRabbit's suggested patch is the right fix: add `monkeypatch: pytest.MonkeyPatch` and call `monkeypatch.delenv("LANGWATCH_PROJECT_ID", raising=False)` before building the reporter.\n\n**[CONCERN] Breaking change: `X-Auth-Token` dual-emit dropped without explicit compatibility note**\n\nThe removed Python comment documented a real customer scenario: corporate proxies that strip `Authorization` headers while preserving `X-Auth-Token`. Dropping it is intentional per the PR description ("server contract is bearer-only"), but the correctness of that assumption depends entirely on server-side rollout being complete. A CHANGELOG entry calling this out clearly would prevent surprise breakage for anyone running a self-hosted or lagging deployment.\n\n**[NIT] Misleading docstring** (`python/tests/test_event_reporter.py`, line ~243)\n\nThe updated text says "an empty bearer token was still being sent on every event." That is not accurate — `post_event()` returns early when `api_key` is falsy (no HTTP call is made). The wording inherited the historical framing from the `X-Auth-Token` era. Consider: "Historically, empty credentials were sent causing 401 storms; the early-return guard prevents that." (Copilot flagged this too.)\n\n**[NIT] Python `or`-based resolution can't be overridden with `""`** (`python/scenario/_events/event_reporter.py`, line 121)\n\n`self.project_id = project_id or langwatch_settings.project_id` means `project_id=""` cannot beat a non-empty env var. Using `project_id if project_id is not None else langwatch_settings.project_id` would honour the "explicit arg wins" contract the docstring documents. Low impact today since the API surface is new, but worth fixing before this pattern calcifies.\n\n**[PRAISE]** The JS side correctly keeps the env-var fallback for `projectId` in `run.ts` (before constructing `EventBus/EventReporter`) rather than inside the constructor. That layering keeps `EventReporter` unit tests naturally isolated from ambient env state — a nice design decision.\n\n**[PRAISE]** Moving the `headers` object above the `try` block in `event-reporter.ts` cleanly separates synchronous setup from the failable async network call, which makes the error boundary more obvious.

Aryansharma28 (approved):
Clean, well-structured change across both Python and JS event reporters. Core logic is correct:
- `X-Project-Id` emitted only when a non-empty project ID is resolved ✓
- `Authorization: Bearer` switch is consistent between both reporters ✓
- Existing guards (`isEnabled` in JS, early-return on falsy `api_key` in Python) remain intact ✓

**[CONCERN]** A minor test reliability issue flagged in the bot comments — tests pass under the happy path but don't cover the case where both env var and explicit param are set (explicit should win).

**[CONCERN]** Docstring in the Python reporter still describes the old auth header format — minor but worth updating for accuracy.

CI fully green. Approving.

PR #633, opened 2026-06-09
fix(docs): add .nojekyll so GitHub Pages serves the .vocs/ dir (search + icons)

### LINKED ISSUE #632: fix(docs): search box returns no results on deployed docs site (search-index + icons 404)
## For Humans

**The deployed docs search box returns nothing because the Pages artifact is built with `actions/upload-pages-artifact@v4.0.0`, whose tar step runs `--exclude=".[^/]*"` and strips every dot-prefixed file/dir from the artifact before upload** — including vocs's `.vocs/` search index + icons. This is NOT Jekyll: the deploy uses `actions/deploy-pages@v4` (the GitHub **Actions** Pages pipeline), which does not run Jekyll at all. The search index and some icons live under `/.vocs/`, so they 404 on the live site; the SPA serves `index.html` for the missing JSON, the client tries to `JSON.parse` it, and search dies.

**Ask:** bump the upload step to `actions/upload-pages-artifact@v5` with `include-hidden-files: true` so dot-paths survive into the artifact.

**The earlier `.nojekyll` approach (#633) was the wrong layer and did nothing.** Jekyll never runs on this pipeline, and `.nojekyll` is itself a dotfile — so it was stripped by the very same `--exclude=".[^/]*"` and never reached the deployed site (the live `/.nojekyll` still 404s after #633 merged).

- Search box opens but returns 0 results on https://scenario-docs.langwatch.ai/ and the langwatch.ai/scenario proxy.
- Heading-anchor (`link.svg`) and external-link (`arrow-diagonal.svg`) icons are also missing — same root cause.

_Status: bug, P2, not-ready (root cause corrected; fix is #634). Full spec below._

### Symptoms / context
On the deployed Scenario docs the search box opens but returns nothing. Browser console (verbatim from reporter + reproduced via curl):
- `GET https://langwatch.ai/.vocs/search-index-494a537f.json` → **404**, `content-type: text/html`
- `Uncaught (in promise) SyntaxError: Unexpected token '<', "<!DOCTYPE "... is not valid JSON`
- `GET /.vocs/icons/link.svg` → **404**, `GET /.vocs/icons/arrow-diagonal.svg` → **404**
- (React minified error #418 — unrelated hydration-text-mismatch noise, NOT the cause)

### Root cause (high-confidence, live + source-evidenced)
**The Pages artifact is packed by `actions/upload-pages-artifact@v4.0.0`, whose tar step unconditionally drops every dot-prefixed entry, so the entire `.vocs/` dir (search index + icons) is excluded from the artifact before it is ever uploaded.** This is an Actions-pipeline artifact-packing problem, **not** a Jekyll problem.

Two facts establish this:

1. **The deploy uses the GitHub Actions Pages pipeline, which does not run Jekyll.** `deploy-docs.yml` uploads with `actions/upload-pages-artifact` and publishes with `actions/deploy-pages@v4`. Jekyll only runs on the **legacy "deploy from a branch" Pages source**; with `deploy-pages` (the Actions source) there is no Jekyll build step, so a `.nojekyll` marker is meaningless here.

2. **`upload-pages-artifact@v4.0.0` strips dotfiles in its tar invocation.** Quoting the action source at the pinned SHA `7b1f4a764d45c48632c6b24a0339c27f5614fb0b` (v4.0.0), `action.yml`:
   ```
   tar \
     ... \
     -cvf "$RUNNER_TEMP/artifact.tar" \
     --exclude=.git \
     --exclude=.github \
     --exclude=".[^/]*" \
     ...
   ```
   The `--exclude=".[^/]*"` pattern matches every dot-prefixed file/dir, so `docs/docs/dist/.vocs/` (and any `.nojekyll`) is removed from the tarball before upload. v4.0.0 has **no** opt-out — the exclude is unconditional. The `include-hidden-files` input that gates this exclude was added in **v5.0.0**.

vocs emits the search index and icons under the **dot-directory** `docs/docs/dist/.vocs/`. The built `index.html` references them root-absolute (`/.vocs/icons/link.svg`, `/.vocs/icons/arrow-diagonal.svg`) and the JS bundle fetches `/.vocs/search-index-<hash>.json`; all 404. The 404 returns the SPA fallback `index.html` (`<!DOCTYPE html>…`), which the search loader `JSON.parse`s → the `Unexpected token '<'` error → search returns nothing.

### Correction history
- **Original diagnosis (this issue, as first filed): WRONG.** It blamed "GitHub Pages' Jekyll strips dot-directories" and prescribed a committed `.nojekyll`. Jekyll does not run on the `deploy-pages@v4` Actions pipeline, so this was the wrong layer.
- **#633 (`.nojekyll`) merged — and was a no-op.** It shipped the marker, but (a) Jekyll never runs on this pipeline, so the marker has no effect, and (b) `.nojekyll` is itself a dotfile, so `upload-pages-artifact@v4.0.0`'s `--exclude=".[^/]*"` stripped the marker from the artifact too — it never even reached the deployed site. Confirmed: after #633 merged and deployed (run [27223032235](https://github.com/langwatch/scenario/actions/runs/27223032235), success), `GET https://scenario-docs.langwatch.ai/.nojekyll` still returns **404**, and `/.vocs/` is still **404** — search is still broken.
- **#634 is the real fix:** bump `actions/upload-pages-artifact` to v5.0.0 + `include-hidden-files: true`, and remove the dead `.nojekyll`.

### Evidence (still TRUE — carried forward)

| Probe (live, `scenario-docs.langwatch.ai` canonical host) | Result | Meaning |
|---|---|---|
| `GET /assets/style-1ygVTMLw.css` | `200 text/css` | normal (non-dot) dir served |
| `GET /assets/index-<hash>.js` | `200 application/javascript` | normal dir served |
| `GET /.vocs/search-index-<hash>.json` | `404 text/html` | **dot-dir dropped from artifact** |
| `GET /.vocs/icons/link.svg` | `404 text/html` | dot-dir dropped |
| `GET /.vocs/` | `404 text/html` | dot-dir dropped |
| `GET /.nojekyll` | `404 text/html` | **#633's marker itself stripped (proves the dotfile-exclude layer)** |

Every dot-path 404s while normal dirs serve → dotfiles are stripped from the published artifact. The `.nojekyll` 404 is the smoking gun that the stripping happens at the **artifact-packing (tar `--exclude`)** layer, not at a Jekyll layer: a Jekyll exclusion would have **honored** the `.nojekyll` and served everything; instead the marker itself was removed. Confirmed on the canonical GitHub Pages host directly, **not** just the langwatch.ai proxy.

**Search-index detail (carried forward):** the search index filename is content-hashed (`search-index-<hash>.json`) and the hash is baked into the JS bundle at build time — so verification must discover the current hash from the **live HTML/JS bundle**, not hardcode `494a537f` (it changes per build).

Local build verification (`docs/docs/dist/` from `pnpm run build`):
- `.vocs/` **is** emitted correctly: contains `icons/` and `search-index-<hash>.json`. So `vocs build` is fine and the workflow's `upload-pages-artifact path: ./docs/docs/dist` is the correct source dir — the loss happens **inside the upload action's tar step**, after build, before upload.
- The built `index.html` hard-references `url(/.vocs/icons/link.svg)` and `url(/.vocs/icons/arrow-diagonal.svg)` inline → the `.vocs/` dir is load-bearing for icons too, not just search.

### Proxy note (carried forward)
`langwatch.ai/scenario` serves the identical bundle (`index-<hash>.js`, `style-1ygVTMLw.css` — same hashes) as `scenario-docs.langwatch.ai`, i.e. it is a proxy of the canonical host. It therefore inherits the exact same breakage; there is **no separate apex-host fix** needed — fix the canonical GH Pages deploy and the proxy follows.

### Hypotheses ruled out
- ~~"GitHub Pages' Jekyll strips the dot-directory"~~ — disproved: this deploy uses `deploy-pages@v4` (Actions pipeline), which runs no Jekyll; and #633's `.nojekyll` had zero effect (the marker itself was stripped and still 404s live).
- ~~"vocs fails to build `.vocs/`"~~ — disproved: local `dist/.vocs/` is fully populated.
- ~~"workflow uploads the wrong directory"~~ — disproved: `upload-pages-artifact path: ./docs/docs/dist` contains `.vocs/`.
- ~~"apex/langwatch.ai proxy rewrite eats the assets"~~ — disproved: the canonical host itself 404s the dot-paths.

- The fix is a one-line workflow change: bump `actions/upload-pages-artifact` from v4.0.0 to **v5.0.0** and add `include-hidden-files: true` under the upload step. In v5.0.0 the dotfile exclude is gated: `${{ inputs.include-hidden-files != 'true' && '--exclude=.[^/]*' || '' }}` — so with the input set, `.vocs/` survives into the artifact. v4 cannot do this (no such input).
- Pin the v5.0.0 SHA: `fc324d3547104276b827a68afc52ff2a11cc49c9`.
- Remove the now-dead `docs/docs/public/.nojekyll` shipped by #633 — it does nothing on this pipeline and would be stripped anyway.
- The `.vocs/` search-index hash changes per build — verification must discover the current hash from the live HTML/JS, not hardcode `494a537f`.

### Caveats
- Could not trigger a live redeploy to confirm the fix end-to-end (requires merge of #634 to `main` → the `Publish Docs` workflow). Post-deploy ACs below are written against the live host and can only be confirmed after merge + redeploy.

## Acceptance Criteria

**Pre-merge-provable: AC1, AC2, AC8.** These gate the PR (no redeploy needed).
**Post-deploy-only: AC3, AC4, AC5, AC6, AC7.** These can only be confirmed after #634 merges to `main` and the `Publish Docs` workflow redeploys; they do not gate the PR but MUST be confirmed before the issue is closed.

1. **(Pre-merge)** The deploy workflow's upload step uses `actions/upload-pages-artifact` **>= v5** with **`include-hidden-files: true`**. Evidence: `git show HEAD:.github/workflows/deploy-docs.yml | grep -nE 'upload-pages-artifact@.*# v5|include-hidden-files: true'` prints **both** lines (the action ref pinned to a v5 SHA, and the `include-hidden-files: true` input under the same step). The v4.0.0 SHA `7b1f4a764d…` must be gone.
2. **(Pre-merge)** The dead `docs/docs/public/.nojekyll` is **removed**. Evidence: `git show HEAD:docs/docs/public/.nojekyll` exits non-zero (path no longer in tree) AND `git log --oneline -1 -- docs/docs/public/.nojekyll` shows a commit that **deletes** it (`git show --stat HEAD -- docs/docs/public/.nojekyll` lists it as deleted in the fix commit's range).
3. **(Post-deploy — primary failure mode resolved)** After redeploy, the search index is served as JSON, not the HTML fallback. Evidence: discover the live hash from the JS bundle (`curl -s https://scenario-docs.langwatch.ai/ | grep -oE 'search-index-[0-9a-f]+\.json' | head -1`, or extract it from the referenced `assets/index-<hash>.js` bundle if not in the HTML), then `curl -sI "https://scenario-docs.langwatch.ai/.vocs/<that-file>"` returns HTTP `200` and a `content-type` **containing** `application/json` (currently `404` / `text/html`). Do **not** hardcode the hash.
4. **(Post-deploy — user-visible behavior)** After redeploy, typing a query in the search box returns matching results, with no JSON-parse error. Evidence, two observables: (i) DOM/screenshot of the search panel showing ≥1 result for a known term (e.g. `judge`); (ii) the browser console captured during that search shows **no** `SyntaxError: Unexpected token '<'`.
5. **(Post-deploy — sibling failure mode: icons)** After redeploy, both `.vocs/` icons resolve. Evidence, asserted independently: `curl -sI "https://scenario-docs.langwatch.ai/.vocs/icons/link.svg"` returns `200` with `content-type` containing `image/svg+xml`; AND `curl -sI "https://scenario-docs.langwatch.ai/.vocs/icons/arrow-diagonal.svg"` returns `200` with `content-type` containing `image/svg+xml` (both currently `404`).
6. **(Post-deploy — regression surface: normal assets still serve)** After redeploy, previously-working non-dot assets are unaffected. Evidence: discover the live CSS (`curl -s https://scenario-docs.langwatch.ai/ | grep -oE 'assets/style-[A-Za-z0-9]+\.css' | head -1`), then `curl -sI "https://scenario-docs.langwatch.ai/<that-path>"` still returns `200` with `content-type` containing `text/css`.
7. **(Post-deploy — proxy ripple)** After redeploy, the langwatch.ai/scenario proxy reflects the fix without a separate change. Evidence: with the live hash from AC3, `curl -sI "https://langwatch.ai/.vocs/<search-index-file>"` returns `200` and `content-type` containing `application/json`.
8. **(Pre-merge — gate/ordering)** The fix is provable on the workflow file **before** any deploy, so the PR is not blocked on a redeploy it cannot trigger. Evidence, both: (a) the workflow YAML parses — `python3 -c "import yaml,sys; yaml.safe_load(open('.github/workflows/deploy-docs.yml'))" && echo PARSE_OK` prints `PARSE_OK`; (b) the PR diff is **exactly** the upload-step change (action ref bump + `include-hidden-files: true`) plus the `.nojekyll` deletion, nothing else — `git diff --stat <base>..HEAD` lists only `.github/workflows/deploy-docs.yml` (+2/-1) and `docs/docs/public/.nojekyll` (deleted). ACs 1–2 + 8 are the PR-gating set; ACs 3–7 are post-deploy must-confirm-before-close.

### Approach
**Bump the Pages upload step to `actions/upload-pages-artifact@v5.0.0` with `include-hidden-files: true`, and delete the dead `.nojekyll`.** v5.0.0 gates its dotfile-stripping tar exclude on that input (`${{ inputs.include-hidden-files != 'true' && '--exclude=.[^/]*' || '' }}`), so setting it keeps `.vocs/` (search index + icons) in the artifact. One ~one-line workflow edit + one file deletion; no vocs or app change. This **supersedes** the `.nojekyll` plan, which targeted a Jekyll layer that does not run on this pipeline.

### Sequencing
1. **Edit the workflow** — in `.github/workflows/deploy-docs.yml`, change the `Upload artifact` step's `uses:` from `actions/upload-pages-artifact@7b1f4a764d…  # v4.0.0` to `actions/upload-pages-artifact@fc324d3547104276b827a68afc52ff2a11cc49c9  # v5.0.0`, and add `include-hidden-files: true` under its `with:` (alongside `path: "./docs/docs/dist"`). Completion signal: AC1 grep prints both lines.
2. **Delete the dead marker** — `git rm docs/docs/public/.nojekyll`. Completion signal: AC2 (path gone + delete commit).
3. **Prove pre-merge** — AC8: YAML parses (`PARSE_OK`) and the diff is exactly the two changes. This is the PR-gating set (AC1, AC2, AC8).
4. **Merge #634 to `main`** — triggers the `Publish Docs` workflow (`deploy-docs.yml`; paths filter includes `docs/**` and the workflow file itself).
5. **Post-deploy live verification** — once the workflow completes, run AC3–7 curls against `scenario-docs.langwatch.ai` (and the langwatch.ai proxy for AC7), discovering the live hash from the bundle. Completion signal: search-index + icons return `200`, search box returns results, then close the issue.

### Key decisions
- **v5 `include-hidden-files` over any Jekyll/`.nojekyll` lever** — the breakage is in the artifact-packing tar step, not Jekyll; `include-hidden-files: true` is the action's first-class, purpose-built opt-out for exactly this. `.nojekyll` cannot work here (Jekyll never runs; the marker is itself stripped).
- **Pin the v5.0.0 SHA** (`fc324d3547104276b827a68afc52ff2a11cc49c9`) — matches the repo's existing SHA-pinning convention for actions.
- **Delete `.nojekyll` in the same PR** — it is dead weight on this pipeline and would be stripped regardless; leaving it invites the same wrong-diagnosis later.

### Risks / open questions
- **Cannot verify end-to-end pre-merge** — AC3–7 require the live redeploy (merge to `main`). Mitigated by AC8: the workflow-level proof gates the PR; live ACs are tracked as must-confirm-before-close, not PR blockers.
- **`include-hidden-files` re-includes ALL dotfiles** in `docs/docs/dist/` (the action still hard-excludes `.git`/`.github`). vocs's `dist/` only emits `.vocs/` as a dot-path, so this is the desired scope; no secret/dotfile leakage risk from a build-output dir.

### Out of scope
- No change to the `langwatch.ai/scenario` proxy/apex config — it's a passthrough of the canonical host and inherits the fix (AC7 verifies this).
- No vocs-framework change (not making vocs emit a non-dot dir) — the workflow input fully resolves this issue.
- The React minified error #418 — unrelated hydration noise, not addressed here.

</details>

drewdrewthis:
## Why

The published Scenario docs search box opens but returns **no results**, and the heading-anchor / external-link icons are missing. GitHub Pages runs Jekyll by default, which excludes any path whose name begins with a dot — and vocs emits its search index and icons under the dot-directory `.vocs/`. So the entire `.vocs/` directory is silently stripped from the deployed site: the search-index request 404s, the SPA fallback serves `index.html`, the client tries to `JSON.parse` that HTML, and search dies.

Closes #632

## What changed

- **Added an empty `docs/docs/public/.nojekyll`** — the marker that tells GitHub Pages to serve the artifact verbatim, without Jekyll. This restores the whole `.vocs/` directory (search index + icons) on the published site.
- **Chose a committed file in `public/` over a CI `touch` step.** vocs copies `public/` into the build artifact root verbatim (the same path `favicon.ico` and `images/` already ship by), so the marker is self-contained: it survives local `vocs preview`, CI, and any future host, and it shows as a reviewable diff. No workflow change required.
- **Added the behavioral contract** `specs/docs-search-nojekyll.feature` describing the deploy-time expectation so this regression is documented.

## Test plan

```bash
cd docs
pnpm install
pnpm run build
test -f docs/dist/.nojekyll && echo OK        # marker lands in the artifact root
# durability across a clean rebuild:
rm -rf docs/dist && pnpm run build && test -f docs/dist/.nojekyll && echo OK
```

Both print `OK` (a real `vocs build` completed in ~32s). The change is two files (`docs/docs/public/.nojekyll`, `specs/docs-search-nojekyll.feature`); no source/runtime code is touched, so the existing suite is unaffected.

No playable artifact exists pre-merge — this is a build/deploy marker, and the end-to-end effect (search returns results) only appears once the **Publish Docs** workflow redeploys after merge. After merge, verify on both the canonical Pages host and the public surface the bug was reported on (the latter proxies the former):

```bash
# 1. Discover the current (content-hashed) search-index filename from the live bundle:
HASH_FILE=$(curl -s https://scenario-docs.langwatch.ai/ \
  | grep -oE '/assets/index-[A-Za-z0-9_-]+\.js' | head -1)
IDX=$(curl -s "https://scenario-docs.langwatch.ai${HASH_FILE}" \
  | grep -oE 'search-index-[0-9a-f]+\.json' | head -1)

# 2. Search index must now be JSON, not the HTML 404 fallback:
curl -sI "https://scenario-docs.langwatch.ai/.vocs/${IDX}"   # expect 200, content-type: application/json
curl -sI "https://langwatch.ai/.vocs/${IDX}"                 # proxy surface: same

# 3. Icons resolve:
curl -sI "https://scenario-docs.langwatch.ai/.vocs/icons/link.svg"            # expect 200 image/svg+xml
curl -sI "https://scenario-docs.langwatch.ai/.vocs/icons/arrow-diagonal.svg"  # expect 200 image/svg+xml

# 4. Regression — normal (non-dot) assets still serve:
CSS=$(curl -s https://scenario-docs.langwatch.ai/ | grep -oE '/assets/style-[A-Za-z0-9]+\.css' | head -1)
curl -sI "https://scenario-docs.langwatch.ai${CSS}"          # expect 200 text/css
```

Then open https://langwatch.ai/scenario, type a query (e.g. `judge`) into the search box, and confirm results appear with no `Unexpected token '<'` error in the console.

### DIFF
--- a/specs/docs-search-nojekyll.feature
+++ b/specs/docs-search-nojekyll.feature
@@ -1,0 +1,102 @@
+Feature: Deployed docs search and assets survive GitHub Pages' Jekyll
+  As a developer using the published Scenario docs
+  I want the search box to return results and the heading/external-link icons to render
+  So that I can find pages without the search silently failing
+  And so the deploy serves the vocs `.vocs/` directory verbatim instead of
+  letting GitHub Pages' default Jekyll strip it because its name starts with a dot
+
+  Background:
+    Given the Scenario docs are a vocs site under docs/ that builds to docs/docs/dist/
+    And vocs emits the search index and icons under the dot-directory docs/docs/dist/.vocs/
+    And the built index.html references /.vocs/icons/*.svg and the JS fetches /.vocs/search-index-<hash>.json root-absolute
+    And .github/workflows/deploy-docs.yml uploads ./docs/docs/dist to GitHub Pages
+    And GitHub Pages runs Jekyll by default, which excludes any path whose name begins with a dot
+    And vocs copies docs/docs/public/ verbatim into the artifact root (favicon.ico and images/ ship this way today)
+    And the canonical host is https://scenario-docs.langwatch.ai and langwatch.ai/scenario is a proxy of it
+
+  # ============================================================
+  # Group: The fix lands in the build artifact (PR-gating, provable pre-merge)
+  # ============================================================
+
+  @integration
+  Scenario: The built Pages artifact contains a .nojekyll marker at its root
+    # AC1, AC9 (PR-gating proof)
+    Given a .nojekyll marker is added to the deployed Pages artifact root
+    When the docs are built with `pnpm run build` from docs/
+    Then a file named exactly ".nojekyll" exists at docs/docs/dist/.nojekyll (artifact root, not nested)
+    And `test -f docs/dist/.nojekyll` from docs/ exits 0
+
+  @integration
+  Scenario: The .nojekyll source is tracked in git, not an ephemeral untracked step
+    # AC2
+    Given the fix uses the committed-file shape
+    When the working tree is inspected
+    Then `git ls-files docs/docs/public/.nojekyll` prints the path
+    And vocs copies it into docs/docs/dist/.nojekyll on the next build
+    But if instead the CI-step shape is used, `git show HEAD:.github/workflows/deploy-docs.yml` contains a net-new line referencing docs/docs/dist/.nojekyll
+    And the mere pre-existing presence of the workflow file does not satisfy the criterion
+
+  @integration
+  Scenario: A clean rebuild still produces the marker
+    # AC8
+    Given the docs/dist output is deleted
+    When `pnpm run build` is run again from docs/
+    Then docs/docs/dist/.nojekyll exists again
+    And a `clean` script does not leave the artifact without the marker
+
+  # ============================================================
+  # Group: The deployed site serves the dot-directory (post-deploy, live host)
+  # ============================================================
+
+  @e2e
+  Scenario: The search index is served as JSON on the live site
+    # AC3 (primary failure mode resolved)
+    Given the Publish Docs workflow has deployed the fixed artifact to scenario-docs.langwatch.ai
+    And the current search-index filename is discovered from the live page rather than hardcoded
+    When `curl -sI` requests https://scenario-docs.langwatch.ai/.vocs/search-index-<hash>.json
+    Then the response is HTTP 200 with a content-type containing application/json
+    And it is no longer a 404 returning text/html
+
+  @e2e
+  Scenario: The search box returns results with no JSON-parse error
+    # AC4 (user-visible behavior)
+    Given the fixed docs are live
+    When a user types a known term such as "judge" into the search box
+    Then the search panel shows at least one matching result
+    And the browser console shows no "SyntaxError: Unexpected token '<'" during the search
+
+  @integration
+  Scenario: The heading-anchor and external-link icons resolve
+    # AC5 (sibling failure mode — icons)
+    Given the fixed docs are live
+    When `curl -sI` requests /.vocs/icons/link.svg and /.vocs/icons/arrow-diagonal.svg on the canonical host
+    Then each returns HTTP 200 with a content-type containing image/svg+xml
+    And neither returns 404
+
+  @integration
+  Scenario: Previously-working non-dot assets still serve
+    # AC6 (regression surface)
+    Given the fixed docs are live
+    And the current CSS asset path is discovered from the live page rather than hardcoded
+    When `curl -sI` requests the /assets/style-<hash>.css path
+    Then it still returns HTTP 200 with a content-type containing text/css
+    And the .nojekyll change has not regressed normal asset serving
+
+  @e2e
+  Scenario: The langwatch.ai/scenario proxy reflects the fix without a separate change
+    # AC7 (proxy ripple)
+    Given the canonical host scenario-docs.langwatch.ai is fixed
+    When `curl -sI` requests https://langwatch.ai/.vocs/search-index-<hash>.json via the proxy
+    Then it returns HTTP 200 with a content-type containing application/json
+    And no apex/proxy configuration change was required
+
+  # --- AC Coverage Map ---
+  # AC 1: ".nojekyll present at deployed artifact root after build" → Scenario: The built Pages artifact contains a .nojekyll marker at its root
+  # AC 2: ".nojekyll source is tracked in git (committed-file OR net-new CI step)" → Scenario: The .nojekyll source is tracked in git, not an ephemeral untracked step
+  # AC 3: "search index served as application/json, not HTML fallback" → Scenario: The search index is served as JSON on the live site
+  # AC 4: "search box returns results, no JSON-parse error" → Scenario: The search box returns results with no JSON-parse error
+  # AC 5: "both .vocs icons return 200 image/svg+xml" → Scenario: The heading-anchor and external-link icons resolve
+  # AC 6: "normal CSS asset still 200 text/css (regression)" → Scenario: Previously-working non-dot assets still serve
+  # AC 7: "langwatch.ai proxy reflects fix, no separate change" → Scenario: The langwatch.ai/scenario proxy reflects the fix without a separate change
+  # AC 8: "clean rebuild still contains .nojekyll" → Scenario: A clean rebuild still produces the marker
+  # AC 9: "pre-merge artifact proof gates PR; live ACs are post-deploy" → Scenario: The built Pages artifact contains a .nojekyll marker at its root (PR-gating clause)


drewdrewthis:
Correction: this fix did **not** work — search was still broken after deploy. Root cause was misdiagnosed as Jekyll, but this Pages site deploys via the GitHub **Actions** pipeline (`actions/deploy-pages`), which doesn't run Jekyll at all. The real cause is `actions/upload-pages-artifact@v4.0.0` stripping all dot-prefixed paths (incl. vocs's `.vocs/`) from the artifact via `tar --exclude=".[^/]*"`. The `.nojekyll` added here is a no-op (and was itself stripped, being a dotfile). Correct fix: #634 (bump upload-pages-artifact to v5 + `include-hidden-files: true`), which also removes this dead marker. See #632 for the corrected investigation.

PR #635, opened 2026-06-09
fix(voice+judge): surface dropped model tool calls (#630 + #631)

### LINKED ISSUE #630: Realtime voice adapter drops model tool/function calls (never surfaced into result.messages)
## For Humans

**The realtime voice adapter silently drops model tool/function calls — they never reach `result.messages`, so no consumer (viewer, judge-as-message, `state.has_tool_call`, downstream extractors) ever sees them.** `OpenAIRealtimeAgentAdapter.recv_audio()` handles audio/transcript/lifecycle events but has no branch for the Realtime API's function-call events; they fall into the terminal `else` and are logged-and-dropped. The JS adapter has the same gap. Scenario owns the realtime transport, so this is Scenario's to fix.

**Ask:** surface a realtime model's tool call into `result.messages` as the documented OpenAI-chat tool-call shape, so `state.has_tool_call("<name>")` is True and the persisted run carries it.

- Drop confirmed at `openai_realtime.py:408-414` (the `else` arm); no `response.function_call_arguments.*` handling anywhere in `python/scenario/voice/`.
- The entire persistence + consumer pipeline ALREADY supports `tool_calls` — only the realtime adapter never produces the structure (zero backend/schema change needed).
- JS consumer reads a different message shape than Python (tool-result/`role:"tool"` vs assistant/`tool_calls`) — parity fix must match each SDK's own consumer.

_Status: `bug`, `P2 - medium`, `investigating`, `not-ready`. Full spec below._

### Expected

When a model under test (or a voiced agent) **calls a registered function tool during a realtime voice scenario**, the tool call should be surfaced into `result.messages` as a **standard OpenAI-chat-shaped tool call** — an assistant message carrying `tool_calls: [{id, type:"function", function:{name, arguments}}]`. Consumers can then render it as a tool call, the judge sees it (message + OTEL spans), and downstream extractors read `tool_calls[].function.arguments`. This is the same shape the **non-realtime** path's agents already produce inside their own LLM loop.

### Actual

The realtime adapter **silently drops** model tool/function calls. They never reach `result.messages`, the LangWatch viewer, or the judge as a structured message. `result.messages` contains only audio+transcript assistant turns; **no `tool_calls` ever appear**.

### Reproduction

Discovered 2026-06-09 building a two-adapter voice interview demo (realtime interviewer agent + voiced candidate). The interviewer model calls a registered `emit_report` function tool at the end; the call never appears in `result.messages`. The demo had to **work around** it: subclass the adapter, hand-roll a raw `_ws.recv()` hook to capture the function-call arguments off the websocket, then staple the JSON onto the agent's turn as a fake `text` content part (`"EMIT_REPORT: " + json`). That text part renders as a visible chat bubble in the LangWatch viewer (a leak), and a test helper string-greps `result.messages` for the marker to recover it. All a workaround for this one missing capability.

## Investigation

**Root cause (BLUF): the realtime adapter's `recv_audio()` event loop has no branch for the Realtime API's function-call events, so every model tool call falls into the terminal `else` arm and is logged-and-dropped — it never becomes a message, so it never reaches `result.messages`.** Critically, **the rest of the pipeline already supports tool calls end-to-end** (state consumer, OpenAI-message passthrough, LangWatch persistence schema). The fix is confined to the adapter: produce the documented tool-call message shape; no backend, schema, or executor change is required.

### 1. The drop — confirmed (`recv_audio` read end-to-end)

`recv_audio()` (`python/scenario/voice/adapters/openai_realtime.py:275-414`) branches on exactly these event types: `response.output_audio.delta`/`response.audio.delta` (line 352), `response.created` (365), `response.output_audio_transcript.delta` (371), `...transcript.done` (379), `response.done`/`response.cancelled` (392), `conversation.item.input_audio_transcription.completed` (397), `error` (401). There is **no branch** for `response.function_call_arguments.delta`/`.done` nor for the `function_call` item in `response.output_item.done`. They hit the terminal `else`:

```python
else:
    # Housekeeping events — session.created, session.updated,
    # response.output_item.added, etc. — are benign. Log at DEBUG
    # and keep the loop running.
    logger.debug(
        "OpenAIRealtimeAgentAdapter: ignoring event type %r", etype
    )
```

`grep -rn "function_call" python/scenario/voice/` returns only docstring mentions (`recording.py`, `elevenlabs.py`) — **zero** handling of `response.function_call_arguments.*`.

### 2. How `result.messages` is built for the realtime path — confirmed

- `VoiceAgentAdapter.call()` (`voice/adapter.py:161-193`) returns exactly **one** assistant message per turn: `create_audio_message(merged, role="assistant")`. `merged` is an `AudioChunk` whose only fields are `data` + `transcript` (`audio_chunk.py:35-36`) — there is no slot for tool calls today.
- `create_audio_message` (`voice/messages.py:45-69`) emits `{"role":"assistant","content":[{input_audio…}, optional {text…}]}` — audio + optional transcript, **never** `tool_calls`.
- The executor turns the adapter's return into `state.messages` via `convert_agent_return_types_to_openai_messages` (`_utils/message_conversion.py:69-70`): a **dict return is passed through verbatim** (`return cast(ChatCompletionMessageParam, obj)`). **So an assistant dict carrying `tool_calls` survives unchanged into `state.messages` → `result.messages`.** This is the clean insertion point.
- `result.messages` element type is `ChatCompletionMessageParamWithTrace` (`types.py:67-74`), a Union that already includes `ChatCompletionToolMessageParamWithTrace` and `ChatCompletionFunctionMessageParamWithTrace` — the type system already admits tool/function messages.

**Where a tool_calls structure must be inserted to land in `result.messages`:** as an assistant message (or messages) returned from the realtime adapter's `call()`/`recv_audio` path, in the documented OpenAI-chat shape. Because `call()` currently returns a single audio message, surfacing a tool call requires the adapter to emit tool-call message structure in addition to (or alongside) the audio turn — an approach decision for the Plan, not an AC.

### 3. The documented contract + the consumer the fix must feed — confirmed

Scenario publishes a tool-calling testing guide: https://langwatch.ai/scenario/testing-guides/tool-calling. It defines the canonical contract (text path only — no realtime/voice mention, consistent with this gap):

```json
{"role":"assistant","content":null,
 "tool_calls":[{"id":"call_123","type":"function",
   "function":{"name":"get_current_weather","arguments":"{\"location\": \"Paris\"}"}}]}
```
and the tool result: `{"role":"tool","tool_call_id":"call_123","content":"…"}`.

**The consumer this fix must feed (Python):** `state.has_tool_call(name)` (`scenario_state.py:242`) → `last_tool_call(name)` (`168-198`), which iterates `self.messages` for `message["role"] == "assistant" and "tool_calls" in message`, then matches `tool_call["function"]["name"]`. So `has_tool_call`/`last_tool_call` read **exactly** the documented assistant `tool_calls` shape on `result.messages`. Feed that shape and the documented assertion API works for realtime with no consumer change.

> Caveat: the guide names `state.latest_agent_message().tool_calls`, but `latest_agent_message()` is **not implemented** in `python/scenario/scenario_state.py` (only `last_message`, `last_user_message`, `last_tool_call`, `last_new_user_message`, `has_tool_call` exist). ACs assert against `has_tool_call`/`last_tool_call` (verified to exist); the doc/code gap for `latest_agent_message` is out of scope here.

### 4. The judge's input path — confirmed; what the judge will/won't see

`JudgeAgent.call()` (`judge_agent.py:412-486`) builds `content_for_judge` from BOTH `JudgeUtils.build_transcript_from_messages(working_messages)` (line 462) AND `_build_trace_digest(spans)` → an `<opentelemetry_traces>` section (463-485). Under the proposed fix:

- **Via message:** once the tool call is a `role:"assistant"` message with `tool_calls` (and, if a tool result is produced, a `role:"tool"` message), it flows into `working_messages` and is rendered by `build_transcript_from_messages` into the `<transcript>` block the judge reads. (Whether `build_transcript_from_messages` formats `tool_calls` richly is a rendering detail; the structure is present and the discovery-loop history handling in `judge_agent.py:816-838` already round-trips `tool_calls`.)
- **Via OTEL:** if the realtime tool call is also emitted as a span, it appears in `<opentelemetry_traces>` independently — the judge sees it even if a model can't ingest the message structure.
- **What the judge will NOT see:** nothing it sees today regresses. The judge currently sees neither (the call is dropped); after the fix it sees the tool call via at least the message path, and via OTEL if a span is emitted.

### 5. OpenAI Realtime API function-call event/item shapes — confirmed against the adapter's existing patterns

The GA Realtime API surfaces a model function call via:
- **Streaming args:** `response.function_call_arguments.delta` (incremental `arguments` JSON) then `response.function_call_arguments.done` carrying `call_id` + complete `arguments`.
- **Item form:** `response.output_item.done` carrying an `item` of `type:"function_call"` with `name` + `call_id` + `arguments`.

**Precedent for event-name choices (GA/beta dual-handling):** the adapter already accepts GA + retired-beta event-name aliases for audio/transcript — `("response.output_audio.delta", "response.audio.delta")` (line 352), `("response.output_audio_transcript.delta", "response.audio_transcript.delta")` (371-374) — via `_warn_if_legacy` (156-170). **#602** (https://github.com/langwatch/scenario/issues/602) drove the GA wire-protocol migration and established this dual-name defensive pattern; the function-call handling should follow the same convention. The Realtime tool-def shape is already passed through in `session.update` (`tools` field, line 221-222; JS `RealtimeToolDef` at `openai-realtime.ts:43-48`).

### Persistence pipeline — confirmed it already supports tool_calls (no backend change)

This is the load-bearing finding for the persisted-`result.messages` AC. The executor persists conversation state to LangWatch via `_emit_message_snapshot_event` (`scenario_executor.py:1639-1665`) → `convert_messages_to_api_client_messages(self._state.messages)` → `ScenarioMessageSnapshotEvent` (`_events/events.py:107-138`, `type_="SCENARIO_MESSAGE_SNAPSHOT"`) → `POST {endpoint}/api/scenario-events` (`_events/event_reporter.py:159`).

`convert_messages_to_api_client_messages` (`_events/utils.py:79-106`) **already extracts `tool_calls` from assistant messages** and maps each to `ToolCall(id, type_="function", function=FunctionCall(name, arguments))` — the generated client models `PostApiScenarioEventsBodyType2MessagesItemType2ToolCallsItem` + `…Function` (`_events/messages.py:18-19,28-29`). **So a `tool_calls`-carrying assistant message round-trips to LangWatch and is retrievable via the backend run API (`GET /api/simulation-runs/{id}`) with zero backend or schema change.** The structure is dropped only because the realtime adapter never creates it.

### JS parity gap — confirmed

`javascript/src/voice/adapters/openai-realtime.ts` `receiveAudio()` (325-388) branches on `response.output_audio.delta`/`response.audio.delta`, the transcript deltas/done, the input-transcription completed, and `error` — and explicitly drops everything else (comment at lines 385-386: "Housekeeping events … are ignored and the loop continues"). **No `function_call` / `tool_call` handling.** `grep` over `javascript/src/voice/` confirms only test/docstring mentions.

**JS consumer shape differs from Python** (parity caveat): `ScenarioExecutionState.hasToolCall`/`lastToolCall` (`javascript/src/execution/scenario-execution-state.ts:139-163`) look for a `role:"tool"` message with a `tool-result` content part matching `toolName` (Vercel AI SDK CoreMessage shape) — NOT an assistant `tool_calls` array like Python. The JS fix must therefore produce the JS-native message shape its own consumer reads; a literal port of the Python message shape would not satisfy `hasToolCall` on the JS side. This is an approach constraint for the Plan; the AC asserts the observable (`hasToolCall` True) without prescribing the message shape.

### Strategies considered

| Strategy | Where the tool call lands | Verdict |
|----------|---------------------------|---------|
| Handle function-call events in `recv_audio`, return tool-call message structure from `call()` | `result.messages` (assistant `tool_calls` + optional `role:"tool"`) → consumer + persistence + judge all already support it | **Primary** — confined to adapter, no downstream change |
| Emit only a `VoiceEvent(type="tool_call")` on the timeline (`recording.py:45-67` already defines this type + `tool_call`/`tool_result`) | `state.timeline`, NOT `result.messages` | Insufficient alone — `has_tool_call`/persistence/viewer read `messages`, not the timeline. Could be done **additionally** for voice-timeline completeness |
| Emit only an OTEL span | `<opentelemetry_traces>` for the judge | Insufficient alone — judge-only; doesn't satisfy `has_tool_call` or persisted `result.messages` |
| Workaround: staple call JSON as a `text` content part (the demo's hack) | A fake text bubble | Rejected — viewer leak, not the documented shape, requires marker-grep |

- **The fix is adapter-local.** Everything downstream of the adapter (`convert_agent_return_types_to_openai_messages`, `has_tool_call`/`last_tool_call`, `convert_messages_to_api_client_messages`, the LangWatch event schema, the judge's transcript+OTEL inputs) already supports `tool_calls`. Do not change them.
- **Ruled out:** timeline-only and OTEL-only surfacing — neither satisfies `has_tool_call` or persisted `result.messages` (the headline requirement). They are optional *additions*, not the fix.
- **Mind the dual event sources:** the GA API delivers a function call via BOTH the streaming `response.function_call_arguments.delta/.done` AND the `function_call` item in `response.output_item.done`. Handle them without double-emitting one logical call (idempotency on `call_id`).
- **Python vs JS message shapes differ** (see §5/JS-parity) — feed each SDK's own consumer.
- **`latest_agent_message()` is doc-only in Python** — don't assert against it.

### Caveats (validate independently)

- Live GA event names: confirm whether live `gpt-realtime*` models emit GA (`response.function_call_arguments.*`, `response.output_item.done`/`function_call`) and/or any retired-beta aliases, and apply the `_warn_if_legacy` dual-name pattern accordingly (per #602).
- Whether `build_transcript_from_messages` renders `tool_calls` in a way a text-only judge reads richly (vs. relying on OTEL) — verify the judge actually sees the tool name/args in the `<transcript>` block, or ensure the OTEL span covers it.
- The `GET /api/simulation-runs/{id}` response shape for `tool_calls` is owned by the LangWatch backend, not this repo — validate the persisted-retrieval AC against a live LangWatch instance.

## Acceptance Criteria

> **ac-review applied (2026-06-09).** This AC set replaces the original draft: 5 Must-Fix corrections + 2 should-improves were folded in. Changes summarised in the Plan's "ac-review applied" note below.

### Core capability

- **AC1 — Tool call reaches `result.messages` (Python).** Realtime agent (`OpenAIRealtimeAgentAdapter`, role=AGENT) calls tool `T` → `result.messages` contains `{"role":"assistant","tool_calls":[{"id","type":"function","function":{"name":"T","arguments":<json-str>}}]}`. Evidence: integration test asserting the message with `function.name == "T"` and parseable args, quoted from captured `result.messages`; AND assert the function-call event is consumed by a dedicated branch (via the branch's effect — accumulator state or a non-DEBUG log), not the terminal `else` at `openai_realtime.py:408-414`.
- **AC2 — Documented assertion API works for realtime (Python).** `state.has_tool_call("T") == True` and `state.last_tool_call("T")["function"]["arguments"]` parses to expected args. Evidence: passing test against the realtime run's state. (Note: `latest_agent_message()` from the public guide is doc-only / not implemented in Python — ACs use the verified `has_tool_call`/`last_tool_call`.)
- **AC3a — Tool call persists (SDK-side, hermetic, CI-runnable).** With reporting enabled, the `SCENARIO_MESSAGE_SNAPSHOT` event POSTed to `/api/scenario-events` carries the tool call as `messages[].tool_calls[].function.{name,arguments}` (name `T`, parseable args), NOT a text-flattened marker. Evidence: respx/httpx_mock test (the `tests/test_event_reporter.py` pattern) capturing the outbound POST body and asserting `convert_messages_to_api_client_messages` mapped the assistant message to a `ToolCall(function.name=='T')`. This is the surface this repo owns.
- **AC3b — Live read-back (env-gated, NOT a CI gate).** Against a live LangWatch instance, `GET /api/simulation-runs/{id}` for the run returns the `tool_calls` structure. Evidence: one live-instance assertion, run manually / behind a `LANGWATCH_API_KEY` gate; documented as out-of-CI. Backend-owned shape — validated, not asserted from SDK code. **[Deferred for #635, 2026-06-10]** This AC is NOT a merge gate for PR #635 — it asserts backend-owned behavior the SDK cannot exercise in CI. #635 delivers the hermetic AC3a (outbound POST shape); AC3b is verified out-of-band against a live instance (or tracked as a separate backend follow-up). `Closes #630` should be read as "AC1–AC11 minus the out-of-CI AC3b live read-back."
- **AC4 — JS parity (named consumer shape).** JS realtime agent calls tool `T` → the JS run's messages contain a `ToolModelMessage` (`role:'tool'`) whose content has a part `{type:'tool-result', toolName:'T'}`, AND `state.hasToolCall('T')===true` AND `state.lastToolCall('T')` returns that message. Evidence: passing JS integration/unit test asserting `hasToolCall('T')===true` AND the returned message's content part has `toolName==='T'` (the consumer-native Vercel shape — a literal Python assistant-`tool_calls` port would fail `hasToolCall`; see `scenario-execution-state.ts:155-163`).

- **AC5 — Judge sees the tool call (OTEL is the required channel).** The tool call is visible to the judge via the `<opentelemetry_traces>` digest: a span named/attributed for tool `T` appears in `JudgeAgent`'s assembled `content_for_judge`. (The plain `<transcript>` does NOT carry `tool_calls` — `build_transcript_from_messages` serialises only role+content, `judge_utils.py:125-128` — so OTEL is the required channel, not an alternative.) Evidence: unit test asserting the assembled `content_for_judge` string contains `T` via the `<opentelemetry_traces>` section over a run with one tool call. If the chosen approach instead enriches the transcript builder to render `tool_calls`, the AC is met by asserting `T` appears inside the `<transcript>` block — but exactly ONE of the two channels MUST be asserted to contain `T`, not "and/or".
- **AC6 — Streaming-delta + item-done resolve to one tool call (idempotency).** Both event forms (`response.function_call_arguments.done` AND the `function_call` item in `response.output_item.done`) for one `call_id` → exactly ONE tool call. Evidence: test feeding both forms, asserting `len(tool_calls) == 1`.
- **AC7 — Malformed/partial event degrades safely (named contract).** Given a function-call event with non-JSON or missing `arguments`, the adapter surfaces the call with `arguments` set to the raw string (or `'{}'` if absent) — it does NOT raise, and the audio turn still returns. Given a `.done` with no `call_id`, the event is skipped with a DEBUG/WARNING log and no message is emitted. (Aligns with the persistence mapper, which defaults `arguments` to `'{}'` and drops a tool_call lacking a `"function"` key — `_events/utils.py:86-94`.) Evidence: two unit tests (malformed-args → call surfaced with raw/empty args, no exception; missing-`call_id` → skipped, turn returns).
- **AC8 — Audio-only turns unchanged (structural invariant, regression).** A tool-free turn produces an assistant message with the same KEYS as today (`role`, `content` with the audio/transcript parts) and NO `tool_calls` key; existing realtime audio/transcript tests pass unchanged. Evidence: existing suite green + a test asserting `'tool_calls' not in message` on a tool-free turn. (Do NOT assert byte-for-byte — assert the structural invariant.)
- **AC9 — Tool-call turn coexists with spoken turn (named observable, ripple).** Model speaks AND calls a tool → both an audio-bearing assistant message AND the tool call are present in `result.messages`; the turn's `VoiceRecording` still has exactly one agent-audio segment with non-empty data and the `agent.start` marker set (the tool call did not displace or null the audio segment). Evidence: integration test asserting both the audio-bearing assistant message AND the tool call are present, and the recording for the turn has the agent-audio segment intact.
- **AC10 — Multiple distinct tool calls in one turn (coverage gap).** Given the model emits two distinct function calls (`call_id` A for `T1`, `call_id` B for `T2`) in one response, when the turn completes, `result.messages` carries BOTH as separate `tool_calls` entries; `has_tool_call('T1')` and `has_tool_call('T2')` are both `True`, and `last_tool_call('T1')`/`last_tool_call('T2')` return the matching args. Evidence: unit test feeding two function-call event streams with distinct `call_id`s, asserting two distinct, correctly-keyed tool calls (no collision, no drop). (Distinct from AC6, which is one call via two event forms.)
- **AC11 — Tool-RESULT round-trip is OUT of scope (explicit scope boundary).** Surfacing the tool RESULT (a `role:'tool'` message / feeding tool output back to the realtime model) is OUT of scope for this issue; this issue surfaces the model's tool-call REQUEST only. Evidence: an AC/test comment asserting only the assistant `tool_calls` request is expected; no `role:'tool'` result message is required. (If a future need arises, file a follow-up.)

## Plan

**Approach.** Add function-call event handling to the realtime adapter's receive loop in BOTH SDKs, and surface the completed call into the agent's returned messages as the documented OpenAI-chat tool-call shape (Python) / the JS-consumer-native shape (JS) so it lands in `result.messages`. The change is **confined to the two `openai_realtime` adapter files** — everything downstream (message passthrough, `has_tool_call`, LangWatch persistence/schema, judge inputs) already supports tool calls and must not be touched.

**Sequencing.**
1. **Python receive-loop handling.** In `recv_audio`, add branches for `response.function_call_arguments.delta` (accumulate per `call_id`), `response.function_call_arguments.done` (finalize), and the `function_call` item in `response.output_item.done` — with GA/beta dual-name tolerance via the existing `_warn_if_legacy` pattern (per #602). De-dup on `call_id` so the streaming + item forms resolve to one call. Completion signal: unit test feeding both event forms yields one finalized call.
2. **Python message surfacing.** Thread the finalized call(s) out of the receive path into `call()` so the returned messages include an assistant `tool_calls` message (and, if a tool result is modeled, a `role:"tool"` message) in the documented shape — without suppressing the audio/transcript turn. Decide the carrier (extend the per-turn return to a message list vs. attach to a new `AudioChunk` field vs. a side-channel the adapter's `call()` reads) — whichever keeps `call()`'s audio turn intact. Completion signal: AC1/AC2 pass.
3. **JS parity.** Mirror the receive-loop handling in `openai-realtime.ts` and surface the call in the shape `ScenarioExecutionState.hasToolCall`/`lastToolCall` reads (tool-result/`role:"tool"` CoreMessage part). Completion signal: AC4 passes.
4. **Persistence + judge verification (no code change expected).** Verify AC3 against a live/recorded LangWatch run and AC5 against the judge's assembled inputs. If a gap surfaces (e.g. `build_transcript_from_messages` doesn't render `tool_calls` and no OTEL span exists), the minimal addition is an OTEL span for the realtime tool call — still adapter-local.
5. **Optional voice-timeline completeness.** Additionally emit `VoiceEvent(type="tool_call"/"tool_result")` (already-defined types, `recording.py:45-67`) so `state.timeline` reflects the call. Not required for the core ACs; do only if cheap.

**Key decisions.**
- **Surface as messages, not timeline/OTEL-only:** the headline requirement (`has_tool_call`, persisted `result.messages`, viewer) reads `messages`; timeline/OTEL are insufficient alone. Rationale documented in Strategies table.
- **Match each SDK's own consumer shape:** Python consumer reads assistant `tool_calls`; JS consumer reads `role:"tool"`/`tool-result`. A literal cross-port would fail the other side's `has_tool_call`.
- **Dual-event-name tolerance per #602:** reuse `_warn_if_legacy` rather than committing to a single event name, mirroring the audio/transcript arms.
- **De-dup on `call_id`:** the GA API emits a call via two event paths; idempotency prevents double-surfacing.

**Risks / open questions.**
- Whether a text-only judge model reads `tool_calls` from the transcript or needs the OTEL span — resolved in step 4; may add a span.
- The exact carrier for threading the call out of `recv_audio` into `call()` (the receive loop returns one `AudioChunk`; tool calls need a parallel channel) — an internal-shape decision, bounded to the adapter.
- `GET /api/simulation-runs/{id}` response shape is backend-owned — AC3 must be validated against a live LangWatch, not asserted from SDK code alone.

**Out of scope.**
- Any change to `has_tool_call`/`last_tool_call`, `convert_messages_to_api_client_messages`, the LangWatch event schema, the executor, or the judge (all already support `tool_calls`).
- Implementing `latest_agent_message()` (doc/code gap, separate concern).
- The downstream demo cleanup (dropping the `EMIT_REPORT` text-staple workaround, reading `tool_calls`) — consumes this fix but is not part of it. The SDK fix ships a demo-agnostic test (AC1) so the capability is proven without the demo.
- Tool calls from non-realtime/wrapper adapters (Pipecat, Twilio, LiveKit) — those wrap a user agent that runs its own LLM loop and already produce `tool_calls`; this issue is the direct-to-model realtime transport only.

**ac-review applied (2026-06-09):** AC5 corrected (transcript builder is tool-call-blind — OTEL span is the required judge channel, not "and/or"); AC3 split into hermetic AC3a (assert outbound POST payload) + gated AC3b (live GET read-back, out-of-CI); AC4 names the exact JS `ToolModelMessage`/`tool-result`/`toolName` shape (literal Python port fails); added AC10 (multiple distinct tool calls/turn) + AC11 (tool-result round-trip explicitly out of scope). **Repo-wide follow-up surfaced:** `build_transcript_from_messages` (`judge_utils.py:125-128`) is tool-call-blind for ALL judges (not just realtime) — candidate standalone issue to render `tool_calls` in the transcript builder so text-path judges see tool usage without depending on OTEL.

</details>

### LINKED ISSUE #631: Judge transcript builder drops assistant tool_calls (judge sees tool usage only via OTEL spans)
## For Humans

**The judge's `<transcript>` renders an assistant tool-call turn as the bare line `assistant: null` — tool name + args are dropped.** `build_transcript_from_messages` serializes only `role` + `content`, never `tool_calls`. The judge sees tool usage ONLY via the parallel `<opentelemetry_traces>` span digest. This is the text-path analogue of #630 and hits every agent path (text + voice).

**Ask:** render `tool_calls` (and confirm `role:"tool"` results) in the transcript so it legibly shows tool usage independent of OTEL.
- Verified empirically: a `get_weather(Paris)` call renders as `assistant: null`; the tool name/args are absent from the transcript, present only in the span digest.
- The `tool`-role RESULT message already renders (`tool: "18C and sunny"`) — only the CALL (name+args) is dropped. Asymmetric.
- Large-trace path makes it worse: when the span digest switches to structure-only, even the OTEL fallback hides call args behind `expand_trace`/`grep_trace` — if the judge never expands the right span (or `_force_verdict` fires), tool visibility can be lost entirely.

_Status: P3 (latent correctness — judges still function via OTEL today), bug, planned. Full spec below._

### Symptoms / context

The judge ingests two channels assembled in `JudgeAgent.call()` (`python/scenario/judge_agent.py:479-486`):

```python
content_for_judge = f"""
<transcript>
{transcript}
</transcript>
<opentelemetry_traces>
{digest}
</opentelemetry_traces>{extra_context_section}
"""
```

`transcript` comes from `JudgeUtils.build_transcript_from_messages(working_messages)` (`judge_agent.py:462`). That builder (`python/scenario/_judge/judge_utils.py:108-129`) is the entire transcript rendering path — there is exactly ONE call site, so the behavior is path-agnostic (text and voice both route through it).

### Root cause / key findings

**The transcript builder reads only `role` + `content` and never renders `tool_calls`** (`judge_utils.py:108-129`, loop body `123-129`):

```python
lines = []
for msg in messages:
    role = msg.get("role", "unknown")
    content = msg.get("content", "")
    truncated_content = _truncate_base64_media(content)
    lines.append(f"{role}: {json.dumps(truncated_content)}")
return "\n".join(lines)
```

An assistant tool-call message in OpenAI shape is `{"role":"assistant","content":null,"tool_calls":[...]}`. With `content=None`, the line becomes `assistant: null`. The `tool_calls` key is never accessed, so the tool name and arguments never reach the transcript.

**Empirically confirmed** (ran `build_transcript_from_messages` against a 4-message weather-tool conversation):

```
user: "What is the weather in Paris?"
assistant: null
tool: "18C and sunny"
assistant: "It is 18C and sunny in Paris."
```
- `get_weather` in transcript? **False**
- tool args (`Paris`) on the assistant tool-call line? **False**
- tool RESULT content (`18C and sunny`) in transcript? **True**

**Two findings the issue stub under-specified, now nailed down:**

1. **`role:"tool"` result messages DO render** — a tool message's `content` IS the result string, so `tool: "18C and sunny"` appears. Only the CALL (name + args on the assistant turn) is dropped. The fix surface is asymmetric: the assistant `tool_calls` need rendering; the `tool`-role result already renders (though it renders WITHOUT linking back to which call produced it — the tool message carries ONLY `tool_call_id`, no tool name on the line, so a multi-tool turn yields ambiguous bare result strings; attribution requires building an id→name map from the assistant turn).

2. **The drop is on the live path, not hypothetical.** Agent dict returns are cast verbatim into the message list: `convert_agent_return_types_to_openai_messages` (`python/scenario/_utils/message_conversion.py:47-49,69-70`) documents and performs `{"role":"assistant","content":"Hi","tool_calls":[...]}` → same dict in `state.messages`. So real tool-calling agents land `tool_calls`-bearing messages that flow straight into the transcript builder. (Note `python/scenario/_utils/utils.py:250-257` — the event/print path ALREADY iterates `tool_calls`; the transcript builder is the lone consumer that ignores them.)

**Truncation / large-trace interaction (the load-bearing "judge can lose tool visibility entirely" claim — confirmed):**

The OTEL fallback is NOT unconditionally complete. `_build_trace_digest` (`judge_agent.py:654-686`) branches on token estimate:

```python
full_digest = judge_span_digest_formatter.format(spans)
is_large_trace = len(spans) > 0 and estimate_tokens(full_digest) > self._token_threshold
if is_large_trace:
    digest = judge_span_digest_formatter.format_structure_only(spans) + "\n\nUse expand_trace(span_id)..."
else:
    digest = full_digest
```

`format_structure_only` (`judge_span_digest_formatter.py:48-89,177-204`) renders **only the span tree skeleton** — `[id] timestamp name (duration, tokens) status` per node, **no attributes, no events, no content**. Tool-call args live in span attributes/events, so in the large-trace path they are NOT inline; the judge must actively call `expand_trace`/`grep_trace` to recover them (the discovery loop, `judge_agent.py:742-845`). If the judge never expands the span carrying the tool call, OR the discovery loop hits `_max_discovery_steps` and `_force_verdict` fires (`judge_agent.py:847-901`), the verdict is made with **neither the transcript NOR the inline digest showing the tool call**. Today that risk is borne entirely by OTEL because the transcript is tool-blind; rendering tool_calls in the transcript removes the single-point-of-failure.

### Strategies considered

| Strategy | What it does | Assessment |
|---|---|---|
| Render `tool_calls` inline in `build_transcript_from_messages` | Append a legible per-call line (name + JSON args) on assistant turns; thread `tool_call_id`->name so `tool`-role results are attributable | **Primary.** Smallest change at the exact drop point; path-agnostic by construction (one call site); media-truncation already applies to args via `_truncate_base64_media`. Independent of OTEL. |
| Rely on OTEL span digest only (status quo / "wontfix") | Leave transcript as-is; document that tool criteria grade off `<opentelemetry_traces>` | Viable TODAY (judges work), but fragile: depends on OTEL configured + span carrying the call + not being truncated behind discovery. The large-trace path can hide it. Rejected as the sole channel. |
| Render in the executor before messages reach the judge | Normalize tool calls into `content` text upstream | Wrong layer — would corrupt `state.messages` for every other consumer (events, report, replay). The transcript builder is judge-local; keep the rendering there. |

- **One call site** (`judge_agent.py:462`) -> fixing the builder fixes text + voice simultaneously. No per-adapter work.
- The voice enrichment path (`_enrich_messages_with_transcripts`, `judge_agent.py:1027-1126`) only prepends a `text` part to **audio-only** messages; it never touches `tool_calls`. So voice tool-call turns are dropped identically — the builder fix covers them.
- `_truncate_base64_media` operates on `content` only today (`judge_utils.py:127`); if tool-call `arguments` can carry base64 (e.g. an image-gen tool), the rendered args should pass through the same truncation to avoid blowing the judge context. Args are a JSON string — decide whether to truncate pre- or post-parse.
- **Ruled out:** this is not an OTEL/span-collector bug (`_build_trace_digest` behaves correctly for what it is); not a `message_conversion` bug (tool_calls pass through as designed). The defect is localized to the transcript builder's per-message rendering.
- Existing test home: `python/tests/test_judge_utilities.py::TestJudgeUtils` (e.g. `test_build_transcript_basic`) — construct messages, call `build_transcript_from_messages`, assert substrings. New ACs map directly onto this shape.

### Caveats

- Verify the exact wire shape your real adapters emit for a tool call (`content: null` vs `content: ""`, and whether `tool_calls[].function.arguments` is a JSON string or already-parsed) before asserting the rendered string — the AC asserts the RENDERED transcript contains the tool name + args, so the renderer must handle both arg encodings.
- Confirm whether any criterion in the existing suite/examples currently relies on the literal `assistant: null` line (unlikely, but a grep over `tests/` + `examples/` for `assistant: null` is cheap insurance against a regression-surface surprise).

## Acceptance Criteria

> **ac-review applied (2026-06-09):** AC1 de-leaked (was asserting `Paris` which leaks from the user turn — now `Tokyo`, absent from any user line, so the token is attributable to the tool call); AC3 now requires an id→name lookup (the tool message carries ONLY `tool_call_id`, no name — the builder must build an id→name map from the assistant turn); AC6 mechanism named (`token_threshold=<small>`, the injectable ctor param already used this way in `test_judge_discovery_exhaustion.py`); AC8 scoped IN with the data-URL form (args-truncation is NEW — `_truncate_base64_media` touches only `content` today, `judge_utils.py:127`); added AC9-AC13 (content+tool_calls coexist, empty args, orphan call, ordering, malformed-JSON-no-crash). AC5 is testable now via a synthetic audio message list — NOT blocked on #630 (this issue fixes the RENDERER; #630 the SOURCE — realtime-source coverage stays vacuous until #630 lands).

### Core rendering

- [ ] **AC1 — Assistant tool-call (`content:None`) renders name + args, no `assistant: null`.** An assistant message with `content:None` and one `tool_calls` entry renders the function name and its arguments on the assistant's transcript line; the literal `assistant: null` does NOT appear. Evidence: `test_judge_utilities.py::TestJudgeUtils` — messages `[{role:assistant, content:None, tool_calls:[{id:c1, function:{name:"get_weather", arguments:'{"city":"Tokyo"}'}}]}]` → `assert "get_weather" in r and "Tokyo" in r and "assistant: null" not in r` (city absent from any user turn so the token is attributable to the tool call, NOT leaked from a user line).
- [ ] **AC2 — Multiple `tool_calls` in one turn all render, in call order.** Two distinct calls (`get_weather`, `get_time`) on one assistant turn → both function names AND both arg sets render, in emission order. Evidence: unit test asserting each of the two distinct names present (neither token leaks from the other).
- [ ] **AC3 — `role:"tool"` result renders, attributed by name via id-lookup.** Each `role:"tool"` result renders (a) its result string AND (b) the originating function NAME, resolved by matching the message's `tool_call_id` to the assistant turn's `tool_calls[].id` (the tool message carries ONLY `tool_call_id`, no name — the builder must construct an id→name map from the assistant turn). With two tool calls + two results, each result line names its own function, unambiguously distinguishable. Evidence: 2-tool unit test — assert `get_weather` on the line bearing the first result, `get_time` on the line bearing the second.

### Scope / path coverage

- [ ] **AC4 — Tool usage reaches the judge's `content_for_judge` `<transcript>` section (text path).** For a text-agent message list, the tool name appears INSIDE the `<transcript>...</transcript>` block of the assembled `content_for_judge` (not only the span digest, and confirmed in the transcript section specifically — not leaking from the digest). Evidence: integration test on a text message list, assert the name between the `<transcript>` markers.
- [ ] **AC5 — Voice path covered by the same renderer (synthetic, NOT blocked on #630).** The fix is in the path-agnostic builder (single call site `judge_agent.py:462`). Test feeds a SYNTHETIC audio message list that ALSO carries `tool_calls` (independent of whether the realtime adapter emits them — that is #630), and asserts the tool name renders. `_enrich_messages_with_transcripts` is NOT required to render tool calls. Evidence: test on an audio message list carrying synthetic tool_calls. (Note: this proves the RENDERER; #630 proves the SOURCE — realtime-source coverage stays vacuous until #630 lands.)

- [ ] **AC6 — Large-trace / truncated digest does NOT hide tool usage (regression-defining).** With `JudgeAgent(token_threshold=<small, e.g. 50>)` so `is_large_trace` is True (→ `_build_trace_digest` returns the structure-only digest, which omits span attributes/args), the assembled `content_for_judge` STILL contains the tool name + args in the `<transcript>` section. Evidence: integration test — `JudgeAgent(token_threshold=50)`, feed messages carrying a tool_call + spans whose digest exceeds 50 tokens, assert the tool name+args sit BETWEEN the `<transcript>`/`</transcript>` markers. (Mechanism: `token_threshold` is an injectable ctor param, already used this way in `test_judge_discovery_exhaustion.py` — no real large trace needed.)
- [ ] **AC7 — No regression for tool-free conversations.** The four existing `TestJudgeUtils` tests (`test_build_transcript_basic`, `test_build_transcript_truncates_media`, `test_build_transcript_handles_empty_messages`, `test_build_transcript_handles_complex_content`) pass with ZERO diff to their bodies (new tests added, old test functions byte-identical); full suite green. Evidence: `git diff` shows no change inside the four existing test functions + suite green.
- [ ] **AC8 — base64 data-URL in tool args is truncated like content.** Tool-call `arguments` are routed through `_truncate_base64_media` (today it touches ONLY `content` — truncating args is a NEW behavior the fix adds); a data-URL base64 value inside `arguments` renders as a `[IMAGE: …]`/`[AUDIO: …]` marker, not raw base64. Evidence: unit test — `tool_calls=[{function:{name:send_image, arguments:'{"img":"data:image/png;base64,' + "A"*2000 + '"}'}}]` → assert `"[IMAGE: image/png" in r` and `"A"*200 not in r`. (Scoped IN — the prior "skip-with-note" escape hatch is removed.)

### Gaps surfaced by ac-review

- [ ] **AC9 — Assistant turn with BOTH content and tool_calls renders both.** An assistant message with non-null `content` AND `tool_calls` renders both the content text and the tool name+args; neither clobbers the other. Evidence: unit test asserting both substrings present.
- [ ] **AC10 — Empty/missing arguments render safely.** A tool_call with `arguments:""` (or `"{}"`) renders the function name with an empty/`{}` marker and does not raise. Evidence: unit test, assert name present, no exception.
- [ ] **AC11 — Orphan tool_call (no matching result) renders safely.** An assistant `tool_calls` entry whose `id` has no later `role:"tool"` message still renders name+args on the assistant line; the missing result does not drop the call or raise. Evidence: unit test with a tool_call and no tool message.
- [ ] **AC12 — Call→result ordering preserved.** Given `assistant(tool_calls=[c1,c2])` then `tool(c1)`, `tool(c2)`, the transcript preserves emission order — the assistant tool-call line(s) precede both result lines. Evidence: unit test asserting the index of the assistant tool-call line precedes both result lines.
- [ ] **AC13 — Malformed (non-JSON) arguments do NOT crash the renderer (critical).** `build_transcript_from_messages` runs on EVERY judge call, so a raise on malformed args breaks ALL judging, not just tool cases. A tool_call with `arguments` that is not valid JSON renders the raw string (or a safe marker) without raising. Evidence: unit test passing `arguments:"{not json"`, assert no exception and the name still present.

_No migration/schema surface (pure in-memory rendering) -> rollback AC omitted. No external-write surface -> failure-handling AC limited to the truncation/regression/malformed cases above._

## Plan

**Approach.** Render assistant `tool_calls` (and make `role:"tool"` results attributable via an id→name map) inside `build_transcript_from_messages` (`judge_utils.py`) — the single transcript rendering site. Reuse the existing `_truncate_base64_media` path for tool-call argument values so the change is one localized function plus tests. No new module, no executor changes, no per-adapter work (one call site at `judge_agent.py:462` covers text + voice). The format is left to implementation; the contract is "tool name + args appear legibly, base64 stays truncated, results stay attributable, malformed args never raise."

**Sequencing.**
1. **Core rendering** — extend the per-message loop so an assistant message with `tool_calls` emits the name + (truncated) args, and `tool`-role results carry their originating tool name via an id→name map built from the assistant turn. Completion signal: AC1–AC3 + AC7 + AC8 + AC9–AC13 unit tests green in `test_judge_utilities.py`.
2. **Path coverage** — assert the same rendering on a text message list and a synthetic voice/audio message list carrying tool_calls (parametrized or two tests). Completion signal: AC4 + AC5 green; grep confirms the single call site.
3. **Truncation independence** — test that with `JudgeAgent(token_threshold=<small>)` (structure-only digest lacking inline args) the tool name + args still appear via the transcript channel in `content_for_judge`. Completion signal: AC6 green.

**Key decisions.**
- **Render in the transcript builder, not the executor** — the builder is judge-local; mutating `state.messages` upstream would corrupt every other consumer (events, report, replay).
- **Reuse `_truncate_base64_media` for args** — args are a JSON string; truncate so an image/audio-bearing tool call can't blow the judge context (AC8). Avoids a second truncation implementation. This is NEW behavior — today `_truncate_base64_media` touches only `content`.
- **Keep the `tool`-role result line but make it attributable** — results already render; the only gap is linkage (which call produced this result), needed for multi-tool turns (AC3). Build an id→name map from the assistant `tool_calls[].id`.
- **Never raise on malformed args** — the builder runs on every judge call (AC13); a raise breaks all judging, so non-JSON args fall back to the raw string / a safe marker.

**Risks / open questions.**
- **#630 dependency for the realtime path.** The realtime voice adapter does NOT currently emit tool_calls into `result.messages` (that IS #630). So AC5's coverage of the *realtime* voice path is vacuous until #630 lands — this issue fixes the RENDERER; #630 fixes the SOURCE for realtime. AC5 here is satisfiable today via a constructed audio message list carrying tool_calls; full end-to-end realtime tool visibility needs both. Flag, don't block.
- **Exact wire shape** — adapters may emit `content:""` vs `content:null`, and `arguments` as a JSON string vs pre-parsed dict; the renderer must handle both (noted in Caveats). Low risk, but the test matrix should include both arg encodings.
- **Format bikeshed** — the rendered line format is unspecified by design; reviewer may want a convention. Defer to implementation + a golden test.

**Out of scope.**
- #630 (realtime adapter not emitting tool_calls into result.messages) — separate upstream source bug.
- Changing the OTEL span digest / discovery loop — `_build_trace_digest` is correct for what it does; this issue makes the transcript an independent channel, it does not alter the trace path.
- Any change to how criteria are phrased or how the judge weighs transcript vs traces.

## Evidence

_Investigation grounded by full-path reads + an empirical run of `build_transcript_from_messages` (2026-06-09). No durable primary-source artifact warranted (no logs/incident dump > ~200 lines); all evidence is inline above and reproducible from the quoted line refs._

---

<details><summary>Original stub (request-for-work)</summary>

### Expected

The judge's text transcript should represent an assistant **tool call** in a legible form (e.g. a line like `assistant called tool T(args)`, or the `tool_calls` rendered inline with the args) so a judge reading the `<transcript>` sees tool usage **without depending on OTEL spans being present**.

### Actual

`build_transcript_from_messages` (`python/scenario/_judge/judge_utils.py:108-129`) serializes **only** each message's `role` and `content` — it does **not** render assistant `tool_calls`. So when an agent under test makes a tool/function call, the resulting assistant message (`{"role":"assistant","content":null,"tool_calls":[...]}`) renders into the judge's `<transcript>` as the bare line `assistant: null` — the tool name and arguments are **absent** from the transcript the judge reads.

The judge can still see the tool call **only** via the separate `<opentelemetry_traces>` span digest that `JudgeAgent.call()` also assembles (`judge_agent.py:479-486`, `_build_trace_digest` at line 464).

This affects **all agent paths (text and voice)**, not just the realtime adapter — it is the **text-path analogue** of the realtime tool-call gap filed as #630.

### Reproduction

Discovered 2026-06-09 during the #630 investigation + AC review: while confirming the judge is NOT blind to realtime tool calls (it sees them via OTEL), the transcript channel itself was found to be tool-call-blind for every judge.

</details>

</details>

drewdrewthis:
## Why

A model's tool/function call was being silently dropped on two independent channels, so nothing downstream could see it. The realtime voice adapter never surfaced a model's tool call into `result.messages` (so `has_tool_call`, the judge, the viewer, and LangWatch persistence all missed it), and — separately — the judge's text transcript renderer never rendered `tool_calls` at all, collapsing an assistant tool-call turn to the bare line `assistant: null` and leaving OTEL spans as the sole, fragile channel.

Closes #630
Closes #631

## What changed

- **#631 (judge transcript, `judge_utils.py`):** render `tool_calls` inline in the single transcript-builder call site (covers text + voice by construction) and attribute `role:"tool"` results to their function via a `tool_call_id`→name map — rather than leaving tool visibility to OTEL, which the large-trace/structure-only digest path can hide. Tool-call args route through the existing `_truncate_base64_media`; malformed-JSON args never raise (the builder runs on every judge call).
- **#630 (realtime adapter, `openai_realtime.py` + `openai-realtime.ts`):** handle the Realtime API's `response.function_call_arguments.delta`/`.done` and the `function_call` item in `response.output_item.*`, accumulating per `call_id` (idempotent across the streaming-args and item forms). Surface the finalized call into the run's messages in each SDK's **consumer-native** shape — Python emits an assistant `tool_calls` message; JS emits the `role:"tool"`/`tool-result` `ToolModelMessage` that `hasToolCall`/`lastToolCall` read — because a literal cross-port would fail the other side's consumer.
- **Scope:** both fixes are confined to their files; persistence, the consumers, and the judge already support `tool_calls`, so no backend/schema change. Tool-RESULT round-trip (feeding output back to the model) is explicitly out of scope (#630 AC11).

## Test plan

Python (run from `python/`):
- `uv run pytest tests/test_judge_utilities.py tests/test_judge_discovery_exhaustion.py -q` — #631 renderer (AC1–AC13) + the integration ACs asserting the tool name reaches `content_for_judge`'s `<transcript>` block even when the OTEL digest is structure-only (large-trace path).
- `uv run pytest tests/voice/test_realtime_tool_calls.py tests/voice/test_realtime_agent_echo_safe.py -q` — #630 adapter (AC1,2,6,7,8,9,10,11), including the audio-only regression.
- `uv run pytest tests/test_event_reporter.py -q` — #630 AC3a: the `SCENARIO_MESSAGE_SNAPSHOT` POST carries `toolCalls[].function.{name,arguments}`, not a flattened marker.
- `uv run pyright .` — whole-package type check (CI gate).

JS (run from `javascript/`):
- `pnpm vitest run src/voice/adapters/__tests__/openai-realtime-tool-calls.test.ts src/voice/adapters/__tests__/openai-realtime.test.ts` — #630 AC4,6,7,8,10 + regression.
- `pnpm typecheck`.

No playable artifact — this is a library/SDK change. The observable proof is the surfaced message shape, asserted in the tests above:

- #630 Python — the assistant message now present in `result.messages`:
  `{"role":"assistant","content":null,"tool_calls":[{"id":"call_abc","type":"function","function":{"name":"get_weather","arguments":"{\"location\":\"Paris\"}"}}]}`
- #630 JS — the message `hasToolCall`/`lastToolCall` now find:
  `{"role":"tool","content":[{"type":"tool-result","toolCallId":"call_a","toolName":"get_weather","output":{"type":"json","value":{"city":"Paris"}}}]}`
- #631 — an assistant tool-call turn that previously rendered `assistant: null` now renders `assistant: [tool_call: get_weather({"city": "Tokyo"})]` inside the judge's `<transcript>`.

## Anything surprising?

- #630 AC5 (judge sees the realtime tool call) is satisfied via the **transcript** channel rather than a new OTEL span, because #631 (committed first on this branch) makes the transcript render `tool_calls`. The realtime assistant `tool_calls` message flows through the same builder.
- AC3b (live LangWatch `GET /api/simulation-runs/{id}` read-back) is env-gated and intentionally out of CI — the persisted shape is backend-owned; the hermetic AC3a asserts the outbound POST payload this repo controls.

## Review notes

- A review thread flagged that tool-call `arguments` reach the persistence POST without base64 truncation. SDK-side truncation of those args was implemented and then **reverted** (commit on this branch): the Scenario SDK sends **raw** message data — byte-content offload is the LangWatch app/API job (the `stored_objects` ingest extractor), not the SDK. Truncating only tool-call args would have made them the one SDK-scrubbed field, inconsistent with how message `content` (e.g. base64 audio) is already sent verbatim. The genuine gap — the app-side extractor does not yet externalize byte content nested in tool-call args — is tracked app-side: langwatch/langwatch#4715.
- The cross-language tool-call shape divergence (JS `tool-result` shape vs Python assistant `tool_calls`) is accepted for this PR and tracked as a convergence follow-up: #640.
- `Closes #630` covers AC1–AC11 **minus** the out-of-CI AC3b live read-back (annotated as deferred on #630).


python/tests/voice/test_realtime_tool_calls.py
```diff
@@ -0,0 +1,615 @@
...
+    monkeypatch.delenv("LANGWATCH_ENDPOINT", raising=False)
+    monkeypatch.setenv("OPENAI_API_KEY", "test-sk-not-a-real-key")
+    try:
+        from scenario._events.event_reporter import EventReporter
+
+        async def _noop_post_event(self, event):
+            return {}
+
+        monkeypatch.setattr(EventReporter, "post_event", _noop_post_event)
+    except ImportError:
```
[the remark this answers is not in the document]
drewdrewthis: Fixed in ee17d9bc — added an explanatory comment to the `except ImportError` block (EventReporter is optional; if it is not importable there is nothing to stub, so the skip is intentional).

python/tests/test_judge_utilities.py
```diff
@@ -223,3 +223,259 @@ def test_build_transcript_handles_complex_content(self) -> None:
...
+                "role": "assistant",
+                "content": None,
+                "tool_calls": [
+                    {
+                        "id": "c1",
+                        "type": "function",
+                        "function": {
+                            "name": "get_weather",
+                            "arguments": '{"city":"Tokyo"}',
+                        },
```
drewdrewthis, line 240: **[test] AC13 assertion is incomplete — assert the function name, not just the raw fragment.** The malformed-non-JSON-args test (#631 AC13) asserts `"not json" in r` but not that the function name (`get_weather`) still renders alongside the raw string. AC13's own text requires "renders the raw string … without raising" AND the name present. Add `assert "get_weather" in r` so a regression that drops the name on the malformed path is caught. (Fix)

python/scenario/_judge/judge_utils.py
```diff
@@ -102,6 +102,73 @@ def transform_fn(v: Any) -> Any:
...
 
+def _render_tool_arguments(arguments: Any) -> str:
+    """
+    Renders a tool call's ``arguments`` for the judge transcript, truncating any
+    base64 media nested inside.
+
+    ``arguments`` arrives either as a JSON string (the OpenAI wire shape) or as an
+    already-parsed dict. We parse strings so that ``_truncate_base64_media`` can
+    reach data-URL values nested in the args, then re-serialize.
+
```
drewdrewthis, line 113: **[security][test] `_render_tool_arguments(None)` is only incidentally safe.** When `arguments` is absent, `None` flows into the non-str branch → `_truncate_base64_media(None)`. It happens not to raise today, but no test covers it and a future refactor of `_truncate_base64_media` could break it silently. Add an explicit guard at the top (`if arguments is None: return "null"` or `"{}"`) and a unit test for the None-args path (distinct from the empty-string `""`/`"{}"` cases already covered). (Fix)

javascript/src/voice/adapters/openai-realtime.ts
```diff
@@ -387,6 +464,175 @@ export class OpenAIRealtimeAgentAdapter extends VoiceAgentAdapter {
     }
   }
 
+  // ------------------------------------------------------------- tool calls
+  // Issue #630: surface OpenAI Realtime function-call events into the run's
+  // messages. The Realtime wire describes ONE logical call across several
+  // events (`response.function_call_arguments.delta` × N → `.done`, and/or
```
drewdrewthis, line 477: **[hygiene] `console.debug` in production bypasses the repo `Logger`.** The no-call_id degraded path uses a bare `console.debug`, but this file otherwise has zero console calls and the JS codebase has a `Logger.create(context)` utility (LOG_LEVEL-gated; the Twilio voice adapter uses it). A bare `console.debug` fires unconditionally regardless of LOG_LEVEL. Either gate it behind `Logger.create("OpenAIRealtimeAgentAdapter")` (matching the Twilio adapter) or drop it and rely on the Python-side `logger.debug`. (Fix)

javascript/src/voice/adapters/openai-realtime.ts
```diff
@@ -387,6 +464,175 @@ export class OpenAIRealtimeAgentAdapter extends VoiceAgentAdapter {
...
+  // `response.output_item.added`/`.done` carrying the function_call item). We
+  // accumulate per call_id, finalize into `_completedToolCalls`, de-duplicating
+  // so each call_id yields exactly one call (AC6).
+  //
+  // NOTE on the surfaced shape (AC4 / AC11): the JS consumer
+  // (`ScenarioExecutionState.hasToolCall` / `lastToolCall`) recognizes a tool
+  // call ONLY as a `role:"tool"` message whose content has a part
+  // `{ type:"tool-result", toolName: T }` (the AI-SDK `ToolModelMessage`
+  // shape). So to make a realtime tool-call REQUEST visible we emit that
+  // `tool-result` shape — but its `output` is just the call's ARGUMENTS, a
```
drewdrewthis, line 487: **[principles][hygiene] JS surfaces a tool-call REQUEST as a `tool-result` shape — decide bridge vs parity.** Python emits the semantically-correct `assistant` message with `tool_calls:[{function:{name,arguments}}]`; JS instead emits `{role:"tool", content:[{type:"tool-result", toolName, output:<the request's arguments>}]}` — a request wearing a result costume, with the call's arguments as `output`. The docstring concedes it. Root cause: JS `ScenarioExecutionState.hasToolCall`/`lastToolCall` only match `role:"tool"`+`tool-result`, never assistant `tool_calls` (unlike Python). DECIDE: accept the bridge for this PR (cheap, but plants a cross-language domain-model divergence) OR extend the JS consumer to read assistant `tool-call` parts so both languages represent a realtime tool call the same way. If accepting the bridge, file a follow-up for JS consumer parity. (Decide)

drewdrewthis:
## Review verdict: READY

Multi-reviewer pass (principles, hygiene, security, test, proof-reviewer) on HEAD `b87a1c0`, post-reconcile. **Proof is sound and complete: 24/24 CI-gating ACs (#630 AC1–AC11, #631 AC1–AC13) map to real, load-bearing tests asserting the AC's observable on the gating surface — verified in-tree, no wrong-surface laundering.** AC3b (live LangWatch read-back) is correctly deferred/out-of-CI, not falsely ticked. All 9 review threads resolved; `prove-it-clean` marker matches HEAD.

No blocking concerns. The items below are non-blocking and tracked for a follow-up commit — they do **not** gate this PR.

### Non-blocking (deferred to a cleanup follow-up issue)

- **[hygiene][security][test]** `javascript/src/voice/adapters/__tests__/openai-realtime-tool-calls.test.ts:302` — a `console.log("[#630 PROOF]", JSON.stringify(toolMsg))` serializes the full tool message to CI stdout each run. Synthetic data today; the `expect` assertions are the durable proof. Pattern-risk, not an active leak. → Fix in follow-up.
- **[principles][hygiene]** `python/scenario/voice/adapters/openai_realtime.py:563` — `call()` override drops type annotations (`# type: ignore[override]`); the JS port keeps them. Restore `input: AgentInput` / return type. → Fix in follow-up.
- **[hygiene]** `openai-realtime.ts` — redundant `?? ""` on a non-nullable `delta` param (~:495); vacuous `if (item && …)` after `?? {}` (~:446). → Fix in follow-up.
- **[test]** `python/tests/test_judge_discovery_exhaustion.py` — `test_ac4/5/6_*` names collide with the unit-test AC namespace in `test_judge_utilities.py`, and one class has two `test_ac5_*` methods. Prefix with issue number. → Fix in follow-up.
- **[principles]** Minor: inconsistent missing-name sentinel (`"unknown"` in the judge renderer vs `""` in the adapter finalize); the args length-tiebreak merge invariant is undocumented. → Optional, in follow-up.

### Out of scope (tracked separately)

> **JS scope note (to avoid confusion):** the JS **adapter** (`openai-realtime.ts`) IS in scope and fully tested — `#630` AC4/AC6/AC7/AC8/AC10 in `openai-realtime-tool-calls.test.ts`. What is out of scope is only the JS **judge-transcript renderer** (`judge-utils.ts`, untouched by this PR). Two different JS surfaces: the adapter surfaces tool calls into the run's messages (shipped); the judge transcript renders them for the judge (Python-only here).

- **[hygiene][principles]** JS **judge-transcript renderer** `JudgeUtils.buildTranscriptFromMessages` (`javascript/src/agents/judge/judge-utils.ts`, NOT touched by this PR) still drops assistant `tool_calls` — the JS twin of the Python `#631` fix shipped here. The JS *adapter* surfacing (above) is unaffected and tested. Relates to #631.
- **[principles]** Cross-language tool-call shape divergence (JS `tool-result` costume vs Python assistant `tool_calls`) — accepted for this PR, convergence tracked in #640.
- **[security/architecture]** Base64 nested in persisted tool-call args lands uncapped — by design (SDK sends raw; app/API offloads). App-side fix tracked in langwatch/langwatch#4715. **Not** a defect in this PR.

python/tests/voice/test_realtime_tool_calls.py
```diff
@@ -0,0 +1,616 @@
...
+    return []
+
+
+# ---------------------------------------------------------------------------
+# AC1 — tool call reaches result.messages (direct call() surface)
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.asyncio
+async def test_ac1_tool_call_surfaced_via_dedicated_branch():
```
drewdrewthis, line 221: **[proof-reviewer] BLOCKING: no `/prove-it` evidence — every AC is author-reported, none verified-in-tree.** All 24 ACs across #630+#631 have tests present in the tree but **zero quoted run output** anywhere on the PR. There is no `<!-- prove-it-clean: -->` marker and no `## Verification Report`. The green `statusCheckRollup` runs the repo's generic `python-ci`/`javascript-ci` jobs — not shown to execute the six named SDK suites — and a CI rollup is not a substitute for the quoted job step that runs these tests. **Resolve by:** running the six suites + two type-checks from the PR's Test plan (`uv run pytest tests/voice/test_realtime_tool_calls.py tests/test_judge_utilities.py tests/test_judge_discovery_exhaustion.py tests/test_event_reporter.py`, `pnpm vitest run …openai-realtime*.test.ts`, `uv run pyright .`, `pnpm typecheck`) and pasting the green output (ideally via `/prove-it` so a marker lands).

python/tests/test_event_reporter.py
```diff
@@ -293,3 +294,98 @@ def to_dict(self) -> dict:
 
     assert big_b64 not in rendered
     assert "<audio:" in rendered
+
+
+@pytest.mark.asyncio
+async def test_message_snapshot_post_carries_tool_call_structured_not_flattened(
```
drewdrewthis, line 300: **[proof-reviewer] BLOCKING: `Closes #630` will auto-close AC3b, but AC3b is not delivered.** #630 AC3b requires that against a live LangWatch instance, `GET /api/simulation-runs/{id}` returns the `tool_calls` structure. The PR body concedes this is env-gated/out-of-CI; the hermetic AC3a here asserts the **outbound POST body** — a different surface than backend-stored-and-returned. On merge, #630 closes with AC3b unproven. **Resolve by (Decide):** either (a) explicitly drop AC3b from #630's scope with a note before merge, or (b) attach a one-off live-instance `GET` read-back payload. (This line is the nearest changed surface; the concern is about the `Closes #630` line in the PR body.)

python/scenario/voice/adapters/openai_realtime.py
```diff
@@ -406,13 +554,51 @@ async def recv_audio(self, timeout: float) -> AudioChunk:
...
+
+        if not self._completed_tool_calls:
+            return audio_message
+
+        # One assistant message carrying ALL of this turn's tool calls — the
+        # conventional OpenAI shape (one assistant turn, many tool_calls).
+        tool_call_message: dict[str, Any] = {
+            "role": "assistant",
+            "content": None,
+            "tool_calls": list(self._completed_tool_calls),
```
drewdrewthis, line 598: **[security] BLOCKING: tool-call `arguments` reach the `SCENARIO_MESSAGE_SNAPSHOT` POST without base64 truncation.** The judge-transcript arm runs args through `_truncate_base64_media` (`judge_utils.py`), but the persistence arm serializes `toolCalls[].function.arguments` verbatim to `app.langwatch.ai`. The Realtime API can pass base64 file/audio/image content as function arguments, so a large/binary arg leaks wholesale — the same exposure class `_truncate_base64_media` was introduced to prevent for message `content`. **Resolve by (Fix):** truncate each accumulated call's `arguments` through `_truncate_base64_media` (parse JSON → truncate string values → re-serialize) in the accumulation layer (both SDKs), so every downstream consumer inherits it. Malformed-JSON args should fall back to verbatim pass-through (already the AC7 behavior).

python/tests/test_judge_utilities.py
drewdrewthis: Fixed in b87a1c0. The AC13 malformed-args test now asserts the function name and the raw fragment render together in the SAME tool-call segment (get_weather rendered alongside the raw non-JSON string), which is strictly stronger than two independent membership checks and catches a regression that drops the name on the malformed path.

python/scenario/_judge/judge_utils.py
drewdrewthis: Fixed in b87a1c0. Added an explicit guard at the top of _render_tool_arguments (if arguments is None: return the empty-args string) so the None path no longer relies on the incidental json.dumps(None) behaviour of the non-str branch, and a dedicated unit test pins the None path distinct from the empty-string and {} cases.

javascript/src/voice/adapters/openai-realtime.ts
drewdrewthis: Fixed in b87a1c0. The no-call_id degraded path now logs via a LOG_LEVEL-gated Logger.create(OpenAIRealtimeAgentAdapter) instead of a bare console.debug, so the file keeps zero direct console calls and matches the Logger convention used elsewhere in the codebase.

javascript/src/voice/adapters/openai-realtime.ts
drewdrewthis: Decision: accept the bridge for this PR. The JS tool-result shape is correct relative to its consumer (ScenarioExecutionState.hasToolCall/lastToolCall match role:tool + tool-result + toolName), so the realtime tool call is visible on the JS side as specced by AC4. The cross-language domain-model divergence (JS tool-result costume vs Python assistant tool_calls) is tracked as a convergence follow-up: https://github.com/langwatch/scenario/issues/640 (extend the JS consumer to also read assistant tool-call parts so both languages represent a realtime tool call the same way). Resolving as accepted-with-follow-up.

python/tests/voice/test_realtime_tool_calls.py
drewdrewthis: Resolved. /prove-it has been run on HEAD b87a1c0 and a prove-it-clean marker is posted on this PR. All six SDK suites + both type-checks are green: python 67 passed / 1 skipped (incl. test_ac2_has_tool_call_and_last_tool_call, test_ac10_multiple_distinct_tool_calls), pyright 0 errors, JS openai-realtime suites 15 passed, tsc --noEmit exit 0. AC3b (live LangWatch read-back) is explicitly out-of-CI/deferred per the #630 annotation, not a CI gate for this PR.

python/tests/test_event_reporter.py
drewdrewthis: Resolved (Decide -> option a). AC3b has been annotated in #630 as Deferred for #635: it asserts backend-owned behaviour (live GET /api/simulation-runs/{id}) the SDK cannot exercise in CI. This PR delivers the hermetic AC3a (outbound POST body shape, asserted in test_event_reporter.py); Closes #630 is to be read as AC1-AC11 minus the out-of-CI AC3b live read-back. So the auto-close is now honest.

python/scenario/voice/adapters/openai_realtime.py
drewdrewthis: Resolved as wont-fix-in-SDK (architecture decision), with the real fix tracked app-side. SDK-side base64 truncation of tool-call arguments on the persistence arm was implemented and then REVERTED (b87a1c0). Rationale: the Scenario SDK sends RAW message data to the backend; byte-content offload is the LangWatch app/API job (the stored_objects ingest extractor), NOT the SDK. Two confirmations: (1) the SDK already sends base64 audio in normal message content to the SCENARIO_MESSAGE_SNAPSHOT POST verbatim (scenario_executor._emit_message_snapshot_event), so truncating only tool-call args would make them the one SDK-scrubbed field, an inconsistency that fights the app-side-offload design; (2) the app already owns this boundary (stored-objects content-extractor) and SHOULD cover tool-call args. The only legitimate SDK truncation is the judge-transcript builder (a context-budget concern, internal to judging, retained here) - never a persistence/security measure. The genuine gap this thread identified is real but lives app-side: base64 nested in persisted tool-call args currently lands uncapped because the extractor toolCall visitor is a no-op and the walker only descends message content, not the top-level tool_calls field. Filed as https://github.com/langwatch/langwatch/issues/4715 (with concrete ACs). The SDK-sends-raw decision is recorded for the fleet.

sergioestebance:
Review findings:

1. **P1** — `python/scenario/voice/adapters/openai_realtime.py:563` / `javascript/src/voice/adapters/openai-realtime.ts:592`
   Tool-only realtime responses still time out because both `call()` implementations wait for the first audio chunk before they can return, but the new function-call accumulation can complete without any audio. I reproduced Python accumulating a tool call and then raising `TimeoutError` when no audio delta arrived. This blocks models that emit a function call instead of spoken audio.
   Minimal fix: make `call()` treat `response.done` / tool-call completion as a valid terminal path and return the tool-call message even when no audio chunk is produced. Add a tool-only regression test in both SDKs.

2. **P2** — `javascript/src/voice/adapters/openai-realtime.ts:620`
   The JS adapter surfaces a tool-call request as a `role:"tool"` message containing `tool-result` parts. That is not the same shape as an assistant tool-call request, and it bakes a semantic mismatch into the run history. The bridge works for the current consumer, but it should be treated as a compatibility shim, not the canonical representation.
   Minimal fix: either extend the JS consumer to recognize assistant `tool_calls` as well, or document this as a deliberate bridge and close the parity gap in a follow-up.

Verification: Python test suites passed (`76 passed, 1 skipped`), and the JS openai-realtime tests/typecheck passed locally.

sergioestebance (approved):
holy bless

PR #408, opened 2026-05-02
fix(security): patch lodash-es code injection via _.template

sergioestebance:
## Summary

Adds pnpm.overrides for lodash-es in docs/ to fix:

- **#262**: Code Injection via _.template imports key names

Part of #400

## Test plan

- [ ] `pnpm install --lockfile-only` succeeds
- [ ] CI passes

PR #406, opened 2026-05-02
fix(security): upgrade mcp Python SDK to fix DoS and DNS rebinding CVEs

sergioestebance:
## Summary

Upgrades mcp (Python SDK) in python/uv.lock from 1.9.4 to 1.27.0 to fix 2 high-severity Dependabot alerts:

- **#111**: Unhandled Exception in Streamable HTTP Transport (DoS)
- **#115**: DNS rebinding protection not enabled by default

Part of #400

## Test plan

- [ ] `uv lock` succeeds
- [ ] CI passes

PR #637, opened 2026-06-10
fix(deps): close 16 npm security alerts in scenario examples

langwatch-agent:
## Why

Closes **16 MODERATE npm alerts** across two scenario example projects, all transitive deps fixed via overrides. No source changes; every bump stays within its existing major.

## What changed

**`javascript/examples/openai-realtime-demo`** (10 alerts). This example is a member of the `javascript/` pnpm workspace but ships its own committed lockfile, so its package-level overrides were being ignored by the workspace resolver (the lock still had hono 4.12.16). Regenerated its lockfile standalone with `--ignore-workspace` so its own `pnpm.overrides` apply.

| package | override | alerts |
| --- | --- | --- |
| hono | `>=4.12.21` | #345 #349 #350 #406 #407 #408 #409 |
| ws | `>=8.20.1` | #383 |
| ip-address | `>=10.1.1` | #335 |
| ajv | `>=8.18.0` | #153 |

**`python/examples/lovable_clone/template`** (6 alerts). A dual-lock scaffold (`package-lock.json` + `pnpm-lock.yaml`); overrides were added to both the npm `overrides` and the `pnpm.overrides` blocks so both lockfiles resolve to patched versions.

| package | override | alerts |
| --- | --- | --- |
| react-router | `>=6.30.4 <7` | #94 #399 #400 |
| @babel/runtime | `>=7.26.10` | #19 |
| js-yaml | `>=4.1.1` | #64 |
| yaml | `>=2.8.3` | #228 |

## How this was verified

- Both example projects pass a frozen install on the regenerated lockfiles.
- Lockfile sweeps confirm no vulnerable copies of the targeted packages remain (hono < 4.12.21, ws < 8.20.1, react-router < 6.30.4, etc).
- The template's `package-lock.json` was regenerated with `--legacy-peer-deps` (matching how the committed lock was produced; there is a pre-existing vite 8 vs `@vitejs/plugin-react-swc` peer conflict, unrelated to these security bumps). The version churn is limited to the targeted packages and their direct deps.

## Not included

- **brace-expansion #380** (template `package-lock.json`). The npm tree carries both a 1.x and a 5.0.5 copy, and npm overrides cannot range-target only the 5.0.x line without forcing the 1.x copy across four majors. Needs a path-scoped override or a parent bump; tracked separately. (The pnpm side does get brace-expansion 5.0.6, since pnpm supports range-keyed overrides.)


langwatch-agent:
Ready for human review. Closes 16 MODERATE npm alerts in two example projects via transitive overrides (within-major). realtime-demo regenerated with --ignore-workspace (workspace member with own lock); template fixed in both npm + pnpm lockfiles. Verified: frozen installs pass, sweeps show no vulnerable copies. brace-expansion #380 deferred (npm can't range-target the multi-major tree).

langwatch-agent:
CI note: the `ci-checks` failure is `scenario-expert-realtime.test.ts` asserting `result.success === true` on a **live voice-to-voice OpenAI Realtime API** scenario judged by an LLM (audio generation + transcription + judge, 90s timeout, retry-on-504 helper). That test runs inside the `javascript` workspace, which uses the workspace lockfile; this PR only changes the realtime-demo example's *standalone* lock and its (workspace-ignored) package.json overrides, so it cannot affect that test's runtime dependencies. The failure is flaky/model-nondeterministic, not a regression from these bumps. I've re-triggered the failed jobs; the security changes themselves (frozen installs clean, no vulnerable copies left) are unchanged.

Issue #644, opened 2026-06-10
Cleanup nits from #635 review (tool-call surfacing)


drewdrewthis:
## Background

Minor, non-blocking cleanup items surfaced by the multi-reviewer pass on [PR #635](https://github.com/langwatch/scenario/pull/635) (tool-call surfacing, #630 + #631). None gate that PR — it merged with these deferred to keep the diff focused. Each is a small, local fix.

## Items

1. **Remove debug `console.log` from a test** — `javascript/src/voice/adapters/__tests__/openai-realtime-tool-calls.test.ts` (~line 302): a `console.log("[#630 PROOF] tool message added to run:", JSON.stringify(toolMsg, null, 2))` serializes the full tool message to CI stdout on every run. Flagged by hygiene + security + test reviewers. The `expect` assertions above it are the durable proof. Pattern risk: `console.log(JSON.stringify(<full message>))` on a tool-call-payload path — synthetic data today, but the pattern would dump argument payloads to CI logs if ever pointed at real fixtures. Fix: delete the log (and the analogous `print(json.dumps(...))` debug lines in the Python test files — `test_realtime_tool_calls.py`, `test_judge_discovery_exhaustion.py` — are lower-risk since pytest captures them, but worth removing for consistency).

2. **Restore type annotations on the realtime `call()` override** — `python/scenario/voice/adapters/openai_realtime.py:563`: `async def call(self, input):  # type: ignore[override]` drops the base class's annotations (`input: AgentInput) -> AgentReturnTypes`, per `voice/adapter.py:161`). The JS port keeps its types. Restore `input: AgentInput` (and the return annotation) so the `# type: ignore[override]` can be narrowed or removed — and the next reader does not wonder why this override is untyped.

3. **Drop redundant `?? ""` on a non-nullable param** — `javascript/src/voice/adapters/openai-realtime.ts` (~line 495), `_accumulateToolCallDelta(callId: string, delta: string)`: body does `acc.args += delta ?? "";` but `delta` is typed `string`; the nullish coalesce can never fire (the call site at ~line 427 already coerces `e.delta`). Remove the dead `?? ""`.

4. **Drop vacuous `if (item && ...)`** — `javascript/src/voice/adapters/openai-realtime.ts` (~line 446): `const item = (event...).item ?? {}; if (item && item.type === "function_call")` — `item` is always an object after `?? {}`, so `item &&` is always truthy. Simplify to `if (item.type === "function_call")`.

[...]

Issue #640, opened 2026-06-10
JS realtime tool calls use a tool-result shape; converge with Python's assistant tool_calls (consumer parity)


drewdrewthis:
## For Humans

**The JS and Python SDKs represent a realtime tool call with different message shapes**, because their `hasToolCall`/`lastToolCall` consumers were built to read different things. PR #635 (#630 + #631) surfaced realtime tool calls in both SDKs and **accepted this divergence as correct-for-now** (each shape satisfies its own consumer). This issue tracks **converging the JS domain model** so a realtime tool call looks the same across languages.

- **Python** emits the semantically-correct assistant message: `{role:"assistant", content:null, tool_calls:[{id, type:"function", function:{name, arguments}}]}`. Its consumer `has_tool_call`/`last_tool_call` reads that shape.
- **JS** emits `{role:"tool", content:[{type:"tool-result", toolName, output:<the request's arguments>}]}` — a tool-call REQUEST wearing a tool-RESULT costume, because `ScenarioExecutionState.hasToolCall`/`lastToolCall` (`javascript/src/execution/scenario-execution-state.ts:139-163`) ONLY match `role:"tool"` + `tool-result` + `toolName`, never an assistant `tool_calls` array.

**Ask:** extend the JS consumer (`hasToolCall`/`lastToolCall`) to ALSO recognize an assistant message carrying `tool_calls` parts (the Python/OpenAI shape), so the JS realtime adapter can emit the same assistant-`tool_calls` representation Python does and both languages converge on one domain model — without breaking the existing `tool-result` consumer path.

**Why it matters:** a request represented as a result is a latent semantic trap (the `output` field holds arguments, not a result). Cross-language divergence in how a core concept (a tool call) is modeled invites consumer bugs and makes the two SDKs harder to reason about together.

_Status: enhancement, P3. Accepted-as-is for #635; this is the convergence follow-up._

## For Agents — investigation

[...]

PR #647, opened 2026-06-10
fix(voice): surface tool-only realtime turns (no audio chunk)

### LINKED ISSUE #646: OpenAI Realtime adapter times out on tool-only responses (no audio chunk ever arrives)
## For Humans

**The OpenAI Realtime adapter times out on a tool-only turn — a model response that is a function/tool call with no spoken audio.** `call()` (both SDKs) awaits the first audio chunk before it can return; a tool-only response produces no audio, so the receive loop runs to its deadline and raises a timeout. The tool call is fully parsed but thrown away. **Reproduced live this session** (hermetic mock-WS unit repro, 2 passing tests).

**Ask:** make `call()` treat `response.done` / tool-call completion as a valid terminal path and return the accumulated tool-call message even when no audio chunk arrived — in BOTH SDKs — with a tool-only regression test each.

- Surfaced by a reviewer on the merged PR #635 (https://github.com/langwatch/scenario/pull/635#issuecomment-4670728897); P1 of that comment. NOT covered by the follow-ups: #640 is the tool-call message-SHAPE parity (request-vs-result), #644 is type-annotations + cleanup nits. This timeout is the functional defect and is untracked until now.
- Scope: a single new terminal-path condition + one regression test per SDK. Hosted-only behaviors and the #640 shape convergence are explicitly out of scope.

_Status: bug, P1, not-ready (investigation complete + reproduced; plan/spec pending). Full spec below._

## Investigation

**Root cause (confirmed on merged `main` `de60f82`, reproduced this session): `call()` is gated on an audio chunk it may never receive.**

The realtime `call()` override (added by #635 for tool-call surfacing) runs the spoken-audio drain FIRST, then appends tool calls:

- Python `python/scenario/voice/adapters/openai_realtime.py:563` — `async def call(self, input)`: line 588 `audio_message = await super().call(input)`, then lines 590-600 build the `tool_calls` assistant message from `self._completed_tool_calls`.
- JS `javascript/src/voice/adapters/openai-realtime.ts:592` — `override async call(...)`: line 598 `const audioMessage = await super.call(input)`, then lines 600-625 build the tool-call message from `this._completedToolCalls`.

`super().call()` drains the agent turn via `recv_audio`/`receiveAudio`, which **returns only on an audio-delta event**:

- Python `recv_audio` loop returns at the `response.output_audio.delta` branch; on `response.done` it merely sets `_response_active = False` and continues the loop until `remaining <= 0`, raising `asyncio.TimeoutError("OpenAIRealtimeAgentAdapter: recv_audio timed out")` (`openai_realtime.py:442`). The drain timeout is `self.response_timeout` (default 60.0s) set at base `adapter.py:205` `first = await self.recv_audio(timeout=self.response_timeout)`.
- JS `receiveAudio`/`_nextEvent` rejects with `new Error("OpenAIRealtimeAgentAdapter: receiveAudio timed out")` (`openai-realtime.ts:719`).

**Therefore a tool-only turn (function call, NO `response.output_audio.delta`) hangs `super().call()` to its deadline and raises — the tool-call assembly code (PY 590-600 / JS 600-625) is never reached.** The function-call events ARE consumed and accumulated by the dedicated branches before the timeout (`_completed_tool_calls`/`_completedToolCalls` is populated); the data is parsed and then lost. The defect is purely the audio-gated return, not the tool-call parsing.

**Reproduction (this session, hermetic, keyless).** New unit repro `python/tests/voice/test_repro_646_toolonly_timeout.py` reuses the `_MockWS` queue-backed websocket from the #635 test (`test_realtime_tool_calls.py`). The #635 tests structurally AVOID this bug: their `_audio_delta_events()` helper carries the comment *"recv_audio only RETURNS on an audio delta, so every turn that goes through call()/drain needs at least one audio delta to terminate the drain loop"* — every test prepends audio. The repro feeds the tool-only sequence WITHOUT audio:

```
test_repro_646_toolonly_call_times_out
  [#646 repro] TimeoutError raised: mock WS: no more events
  PASSED            # call() raises asyncio.TimeoutError on response.created → function_call_arguments.delta×2 → .done → response.done (no audio); _completed_tool_calls non-empty (parsed, then lost)
test_repro_646_audio_plus_tool_does_not_time_out
  [#646 control] tool calls returned: [{'id':'call_weather','type':'function','function':{'name':'get_weather','arguments':'{"location":"Paris"}'}}]
  PASSED            # SAME tool-call events + audio → call() returns the tool call cleanly
2 passed in 0.03s
```

The control isolates the fault to the no-audio path: identical tool-call events succeed when an audio delta is present. (Tests bound the wait via `adapter.response_timeout = 0.5` so the run completes in 0.03s.)

**JS twin:** confirmed by reading — `openai-realtime.ts` `call()` has the identical `await super.call(input)`-first structure (line 598) gating on `receiveAudio` (line 598 then `_completedToolCalls` check at 600-604). The BUG (audio-gated return) is identical across SDKs. NOTE the surfaced tool-call message SHAPE is NOT identical and must not be ported: PY emits `{role:"assistant", tool_calls:[...]}` (consumer `scenario_state.py` matches `"tool_calls" in message`); JS emits `{role:"tool", content:[{type:"tool-result",toolCallId,toolName,output}]}` (consumer `scenario-execution-state.ts:139-163` matches `role:"tool"` + `part.type==="tool-result"`). Each SDK's fix must surface ITS OWN consumer shape; the cross-SDK shape convergence is separately tracked by #640 and is OUT of scope here. A symmetric mock-ws repro is an implementation deliverable (AC2).

**Why #635 missed it:** #635's goal was surfacing tool calls that *coexist* with a spoken turn (#630/#631); all its fixtures pair audio with the tool call. The pure tool-only turn (tool call, model stays silent) was never in a fixture.

**pre-act-check (own-code bug — dependency-blame gate not triggered):** root cause is our own `call()` ordering, not an SDK/upstream defect and not a can't-do-X claim about a dependency. Prior-art grep of `references/decisions` / `references/solutions` / `references/common-mistakes.md` for the symptom: clean (only self-references to the `receiveAudio timed out` string in an unrelated mistakes row).

### Core behavior
1. Given a realtime agent turn that emits a function/tool call and NO `response.output_audio.delta` before `response.done`, when PY `call()` is invoked, then it RETURNS an assistant message of the SAME shape the existing audio+tool path emits — `{"role":"assistant","content":None,"tool_calls":[{id,type:"function",function:{name,arguments}}]}` — and does NOT raise a timeout. Evidence: a keyless mock-WS unit test (PY) feeding `response.created → response.function_call_arguments.delta×N → .done → response.done` (no audio) asserts `call()` returns that message shape; the test passes. (Twin of the #635 audio+tool test, minus the audio.)
2. Given the same no-audio tool-only turn, JS `call()` RETURNS a `{role:"tool"}` message whose `content` carries a `tool-result` part per call (`{type:"tool-result",toolCallId,toolName,output}`) — the SAME shape the existing audio+tool path emits — and does NOT throw the receiveAudio timeout. Evidence: a symmetric keyless mock-ws unit test in `javascript/src/voice/adapters/__tests__/` (extend `openai-realtime-tool-calls.test.ts` or a sibling) asserts that shape; the test passes. NOTE: the JS surfaced shape is `role:"tool"`/`tool-result` parts, NOT the PY `tool_calls` array — assert each SDK against ITS OWN consumer shape, do not port one shape to the other.
3. The surfaced tool call is CONSUMABLE end-to-end via the public assertion API, not merely returned by `call()`. After the no-audio tool-only message passes through `convert_agent_return_types_to_openai_messages` (PY) / `convertAgentReturnTypesToMessages` (JS) into the run messages / ScenarioState: PY `state.has_tool_call("<name>")` is True and `state.last_tool_call("<name>")["function"]["arguments"]` equals the emitted args; JS `state.hasToolCall("<name>")` is True and `state.lastToolCall("<name>")` returns the message with the matching `tool-result` part. Evidence: a unit test per SDK that feeds the surfaced message through the converter into a ScenarioState (not just `call()`'s raw return) and asserts has/last; both pass. Falsifies if `call()` returns the message but `has_tool_call` is False (wrong return container — bare dict vs list — or wrong message role/shape).
4. Both the streaming-args form (`response.function_call_arguments.delta`×N → `.done`) AND the output-item form (`response.output_item.done` carrying a `function_call` item) are accepted as the sole terminal source of a tool-only turn. Evidence: parametrized/duplicated unit cases for each form pass in both SDKs (PY helpers `_function_call_streaming_events`/`_function_call_item_events` in `test_realtime_tool_calls.py`; JS cases in `openai-realtime-tool-calls.test.ts` — add whichever form is not already present). The pre-existing form-coverage tests all prepend an audio delta; AC4's requirement is that EACH event form terminates a **tool-only turn with NO audio** — assert each form on the no-audio path, not via the audio-bearing tests.
5. A tool-only turn (no audio) emitting TWO distinct function calls before `response.done` surfaces BOTH — PY one assistant message with `len(tool_calls)==2`; JS one `role:"tool"` message with two `tool-result` parts; `has_tool_call`/`hasToolCall` is True for each name. Evidence: a unit case per SDK with two calls (e.g. `get_weather` + `get_time`) asserts both are present; passes. Falsifies if only the first (or last) call survives.

6. Regression — a tool-FREE spoken turn (audio deltas, no function call) still returns a SINGLE assistant audio message byte-identical to today (no `tool_calls` key, a dict not a list in PY). Evidence: the existing `test_ac8_tool_free_turn_has_no_tool_calls_key` (PY) and the existing JS `AC8 — audio-only turn (no tool events) returns ONLY the audio message, no tool message` test (asserts `messages.length===1` + `toolMessageOf` undefined) still pass unchanged.
7. Regression — an audio+tool turn (the #635 case: model speaks AND calls a tool) still returns BOTH the audio-bearing assistant message AND the tool-call message. Evidence (PY): the existing `test_ac9_audio_and_tool_call_coexist` (`python/tests/voice/test_realtime_tool_calls.py`) passes unchanged. Evidence (JS): a unit case in `openai-realtime-tool-calls.test.ts` feeding one audio delta AND one tool call asserts the return contains BOTH a `role:"assistant"` audio-bearing message AND a `role:"tool"` message carrying the `tool-result` part (e.g. `messages.length===2`, an assistant audio message present, `toolMessageOf(messages)` defined). The existing JS `AC4 — surfaces a tool call as the JS consumer shape` test asserts ONLY the tool message and is NOT a coexistence regression — ADD this JS coexistence case.
8. Cross-turn no-bleed — across two sequential `call()` invocations on the same adapter instance (turn 1 tool-only/no-audio, turn 2 audio-only/no tool call), turn 2 returns a SINGLE audio message with NO `tool_calls` key (PY a dict not a list; JS no trailing `role:"tool"` message); turn 1 does not bleed into turn 2. Evidence: a unit case per SDK driving two turns on one adapter asserts turn-2 shape is clean; passes. Falsifies if turn 2 carries turn 1's tool call.
9. Failure mode — a genuinely empty turn (`response.created → response.done` with NO audio delta AND an EMPTY tool-call accumulator, or the websocket closes with neither) raises the SAME error as today (PY `asyncio.TimeoutError` message `"OpenAIRealtimeAgentAdapter: recv_audio timed out"`; JS `Error` `"OpenAIRealtimeAgentAdapter: receiveAudio timed out"`) — it does NOT return None, an empty message, or a `tool_calls`/`tool-result` message with an empty list. Evidence: a unit case per SDK feeding `response.created → response.done` (no audio, no function_call) asserts the timeout/closed error is raised AND that `call()` does not return; passes. Falsifies if the empty turn returns instead of raising — i.e. the new terminal path keys on `response.done` alone rather than `response.done` AND a non-empty accumulator.
10. Failure mode — a tool-only turn (no audio) whose `.done`/item carries malformed or missing arguments still returns the tool call WITHOUT raising, with arguments degraded per the existing #635 contract (PY: `arguments` is the raw string or `"{}"`; JS: `output` is the parsed JSON value or the raw string verbatim). Evidence: a unit case per SDK combining the no-audio path with malformed args (mirroring `test_ac7a`/`test_ac7b`) passes.
11. Regression — the full keyless suites in both SDKs stay green after the change. Evidence: `uv run pytest tests/voice/test_realtime_tool_calls.py` (PY, from the `python/` dir) and `pnpm vitest run src/voice/adapters/__tests__/openai-realtime-tool-calls.test.ts` (JS, from the `javascript/` dir) both pass; quote the pass counts.

No migration/schema/rollback surface (pure in-process control-flow change) — rollback AC omitted.

## Plan

**Approach:** Add a **tool-only terminal path** to the realtime receive/drain loop so that a `response.done` arriving with an empty audio buffer but a NON-empty tool-call accumulator resolves the turn (returns the accumulated tool calls) instead of looping until the `response_timeout` deadline. Symmetric change in both SDKs, each surfacing ITS OWN existing consumer shape (PY `tool_calls` assistant message; JS `role:"tool"`/`tool-result`) — no new message shape is introduced (that convergence is #640). Add a no-audio tool-only regression test per SDK plus the consequence/failure-mode cases the ACs name.

**Sequencing:**
1. **PY terminal path.** In `python/scenario/voice/adapters/openai_realtime.py` (the `recv_audio` drain loop and/or the `call()` override), make a tool-only turn return cleanly: when `response.done` is seen and no audio chunk has been produced but `_completed_tool_calls`/accumulators are non-empty, terminate the drain so `call()` reaches its tool-call assembly (lines ~590-600) and returns the assistant `tool_calls` message. Preserve the existing timeout for the genuinely-empty turn (done + EMPTY accumulator). Completion signal: AC1 + AC9 (empty-turn still raises) pass.
2. **PY remaining cases.** Cover both event forms (AC4), multi-call (AC5), end-to-end consumability through the converter (AC3), cross-turn no-bleed (AC8), malformed-args degrade (AC10). Completion signal: new PY unit tests for AC1/AC3/AC4/AC5/AC8/AC9/AC10 pass; existing AC8/AC9 (#635 regression, now this issue's AC6/AC7) unchanged.
3. **JS terminal path + cases (mirror).** Apply the symmetric change in `javascript/src/voice/adapters/openai-realtime.ts` (`receiveAudio`/`_nextEvent` drain + `call()` override) and add the JS analogs of every AC test in `javascript/src/voice/adapters/__tests__/openai-realtime-tool-calls.test.ts`. Completion signal: AC2 + JS analogs pass.
4. **Full-suite green + evidence.** Run both keyless suites; quote pass counts (AC11).

**Key decisions:**
- **Terminate on `response.done` AND non-empty accumulator, not `response.done` alone** — keying on `response.done` alone would make a genuinely empty/silent turn return an empty message instead of surfacing the real hang (AC9). The accumulator non-empty is the discriminator.
- **Each SDK keeps its own surfaced shape** — PY `{role:"assistant",tool_calls:[...]}`, JS `{role:"tool",[tool-result...]}`. Porting one shape to the other would break that SDK's `has_tool_call`/`hasToolCall` consumer and collides with #640. The fix is purely about REACHING the existing assembly code on a no-audio turn.
- **Prove consumability through the converter, not just `call()`'s return** (AC3) — the assertion API reads `state.messages`, so the regression test must run the surfaced message through `convert_agent_return_types_to_openai_messages` / `convertAgentReturnTypesToMessages` into a ScenarioState.

**Risks / open questions:**
- The drain loop is shared by all turns; the new terminal condition must not perturb audio-only or audio+tool turns (guarded by AC6/AC7 regressions).
- Tail-silence handling: the base drain loops `recv_audio` for tail silence after the first chunk — confirm the tool-only path does not depend on a first chunk that never arrives (it must terminate on done+accumulator BEFORE any first-chunk assumption).
- Per-turn reset placement: the accumulator reset already lives at `call()` turn-start; AC8 verifies no cross-turn bleed once the no-audio path can actually complete a turn.

**Out of scope:**
- Cross-SDK tool-call message-SHAPE convergence (request-vs-result) — tracked by #640.
- Type-annotation restoration on `call()` and the other #635 cleanup nits — tracked by #644.
- Live-API (real OpenAI key) E2E for tool-only turns — the hermetic mock-WS unit tests are the contract; a live smoke is optional follow-up, not an AC here.
- Hosted-ElevenLabs single-exchange behavior (#638) — unrelated adapter.

</details>

drewdrewthis:
## Why

Closes #646.

The OpenAI Realtime adapter **times out on a tool-only turn** — a model response that is a function/tool call with no spoken audio. `call()` awaited the first audio chunk before it could return; a tool-only response produces no audio, so the receive loop ran to its deadline and raised a timeout. **The tool call was fully parsed, then thrown away.** This blocked any model turn that emits a function call instead of speech. Surfaced by a reviewer on the merged PR #635 (https://github.com/langwatch/scenario/pull/635#issuecomment-4670728897).

## What changed

- **Added a tool-only terminal path to the receive/drain loop in both SDKs.** When `response.done` arrives with no audio produced but **≥1 tool call finalized this turn**, the drain returns an empty chunk so `call()` reaches its existing tool-call assembly and surfaces the message — instead of looping to the `response_timeout` deadline.
- **The discriminator is the non-empty accumulator, not `response.done` alone** — a genuinely empty/silent turn (done + empty accumulator) must still raise the timeout, so the new path is gated on accumulated calls. This is the load-bearing decision (verified by the empty-turn regression).
- **Each SDK keeps its own consumer shape** — Python emits the assistant `tool_calls` message; JS emits `role:"tool"` / `tool-result` parts. The fix only changes *when the drain terminates*, never the surfaced shape. Cross-SDK shape convergence is a separate concern (#640), deliberately out of scope.
- **Folded two #644 cleanup nits** that live in files this PR already touches (boy-scout): removed a debug `console.log(JSON.stringify(...))` from the JS test, and restored the dropped type annotations on the Python `call()` override.

## How it works

The bug and fix are one branch each, symmetric across SDKs:

| SDK | Drain method | Fix |
|-----|-------------|-----|
| Python | `recv_audio` (`openai_realtime.py`) | the existing `response.done`/`response.cancelled` branch returns `AudioChunk(data=b"")` when `self._completed_tool_calls` is non-empty |
| JS | `receiveAudio` (`openai-realtime.ts`) | a **new** `response.done`/`response.cancelled` branch (there was none) returns `new AudioChunk({data: new Uint8Array(0)})` when `this._completedToolCalls.length > 0` |

The empty chunk lets the base `drainAgentResponse` tail-silence loop terminate cleanly; control returns to `call()`, which finds the accumulator non-empty and returns the tool-call message that was already parsed.

## Test plan

Each SDK got a **fail-first regression test** (committed *before* its fix — see the git history) then the fix; the suites are keyless (mock-WS / in-process WS), no API key needed.

```
# Python — from python/
uv run pytest tests/voice/test_realtime_tool_calls.py
# → 18 passed

# JavaScript — from javascript/
pnpm vitest run src/voice/adapters/__tests__/openai-realtime-tool-calls.test.ts
# → 17 passed
```

- Fail-first proof (Python): `test_issue646_tool_only_no_audio_returns_tool_call` raised `TimeoutError: mock WS: no more events` on the pre-fix code; passes after.
- Fail-first proof (JS): `#646 (AC2) — tool-only turn (no audio delta)...` rejected with `receiveAudio timed out` (11.85s) on the pre-fix code; passes after (1.96s).
- Coverage spans AC1–AC11: both event forms (streaming-args + output-item), multi-call, end-to-end consumability through `has_tool_call`/`hasToolCall`, cross-turn no-bleed, **the empty-turn-still-times-out failure mode** (proves the discriminator), malformed-args degradation, and the audio+tool coexistence regression.

**No playable artifact — see Test plan.** This is a pure control-flow change in a transport adapter; the demonstration is the regression test that *times out on the parent commit and passes on this branch*. The fail-first commits (`24811aa`, `c289bf4`) are the in-PR evidence that the bug was real and is now fixed: check out the test commit alone and the suite times out; add the fix commit and it goes green.

## Anything surprising?

- **The JS `receiveAudio` had no `response.done` handling at all** — `response.done` fell into the housekeeping fall-through and the loop just continued. The Python side at least had a `response.done` branch (it only set a flag). So the JS fix adds a branch; the Python fix extends one.
- **A `role:"tool"` message carrying arguments is intentional, not a bug** — JS surfaces a tool-call *request* in a tool-*result* shape because that is what its `hasToolCall` consumer matches today. Converging that with Python's `tool_calls` shape is tracked by #640 and is explicitly not in this PR.

## Acceptance Criteria

All 11 ACs verified in-tree this turn — keyless suites, no API key. PY: `uv run pytest tests/voice/test_realtime_tool_calls.py` → **18 passed**. JS: `pnpm vitest run src/voice/adapters/__tests__/openai-realtime-tool-calls.test.ts` → **17 passed**.

| # | Criterion | Evidence (test, green this turn) | Status |
|---|-----------|----------------------------------|--------|
| 1 | PY tool-only turn returns the `tool_calls` assistant message, no timeout | `test_issue646_tool_only_no_audio_returns_tool_call` | ✅ |
| 2 | JS tool-only turn returns `role:"tool"`/`tool-result`, not the PY shape | `#646 (AC2) — tool-only turn (no audio delta)…` | ✅ |
| 3 | Consumable end-to-end via `has_tool_call`/`hasToolCall` | PY `test_issue646_tool_only_consumable_via_scenario_state`; JS `#646 (AC3)` | ✅ |
| 4 | Both event forms (streaming-args + output-item) terminate a tool-only turn | PY `…_streaming_args_form` + `…_output_item_form`; JS `#646 (AC4)` ×2 | ✅ |
| 5 | Two simultaneous calls both surface | PY `…_two_calls_both_surface`; JS `#646 (AC5)` | ✅ |
| 6 | Regression: tool-free spoken turn unchanged | PY `test_ac8_tool_free_turn_has_no_tool_calls_key`; JS `AC8 — audio-only turn…` | ✅ |
| 7 | Regression: audio+tool coexist (both messages) | PY `test_ac9_audio_and_tool_call_coexist`; JS `#646 (AC7)` (new coexistence assertion) | ✅ |
| 8 | Cross-turn no-bleed | PY `test_issue646_cross_turn_no_bleed`; JS `#646 (AC8)` | ✅ |
| 9 | Genuinely empty turn STILL times out (discriminator = non-empty accumulator) | PY `test_issue646_empty_turn_still_times_out`; JS `#646 (AC9)` | ✅ |
| 10 | Malformed args degrade (no raise) on the no-audio path | PY `test_issue646_tool_only_malformed_args_degrade`; JS `#646 (AC10)` | ✅ |
| 11 | Full keyless suites green | PY 18 passed; JS 17 passed (quoted above) | ✅ |

✅ = verified in-tree this turn. CI re-confirmation pending (PR is draft until CI green).

### DIFF
--- a/javascript/src/voice/adapters/__tests__/openai-realtime-tool-calls.test.ts
+++ b/javascript/src/voice/adapters/__tests__/openai-realtime-tool-calls.test.ts
@@ -380,7 +380,7 @@ describe("OpenAIRealtimeAgentAdapter — realtime tool-call surfacing (#630)", (
     });
   });
 
-  it("#646 (AC4) — output-item form terminates a tool-only turn", async () => {
+  it("#646 (AC4, variant) — output-item form terminates a tool-only turn", async () => {
     const adapter = buildAdapter({ apiKey: "test-key", role: AgentRole.AGENT });
     adapter.responseTimeout = 0.5;
 
@@ -431,6 +431,8 @@ describe("OpenAIRealtimeAgentAdapter — realtime tool-call surfacing (#630)", (
     const messages = await runTurn(adapter, () => {
       pushAudioDelta();
       pushStreamingCall("call_co", "fetch", '{"k":"v"}');
+      // audio delta + tool call + explicit response.done — exercises the coexistence terminal (both messages must survive).
+      push({ type: "response.done" });
     });
 
     expect(messages.length).toBe(2);
@@ -470,19 +472,22 @@ describe("OpenAIRealtimeAgentAdapter — realtime tool-call surfacing (#630)", (
     await adapter.connect();
     await socketReady;
     await waitForType("session.update");
+    let caught: unknown;
     try {
       // Empty turn: no audio, no function call — just response.done. The
       // accumulator stays empty, so the new terminal path does NOT fire and
       // the drain must still hit the receiveAudio timeout (proves the
       // discriminator is the non-empty accumulator, not response.done alone).
-      await expect(
-        feedAndCall(adapter, () => {
-          push({ type: "response.done" });
-        }),
-      ).rejects.toThrow("receiveAudio timed out");
+      await feedAndCall(adapter, () => {
+        push({ type: "response.done" });
+      });
+    } catch (e) {
+      caught = e;
     } finally {
       await adapter.disconnect();
     }
+    expect(caught).toBeInstanceOf(Error);
+    expect((caught as Error).message).toContain("receiveAudio timed out");
   });
 
   it("#646 (AC10) — malformed args degrade on a tool-only turn", async () => {
--- a/javascript/src/voice/adapters/__tests__/openai-realtime-tool-calls.test.ts
+++ b/javascript/src/voice/adapters/__tests__/openai-realtime-tool-calls.test.ts
@@ -316,6 +316,34 @@ describe("OpenAIRealtimeAgentAdapter — realtime tool-call surfacing (#630)", (
     ).toBe(true);
   });
 
+  // --- #646: tool-only (no-audio) turn ---
+  it("#646 (AC2) — tool-only turn (no audio delta) returns the role:'tool' message, does not time out", async () => {
+    const adapter = buildAdapter({ apiKey: "test-key", role: AgentRole.AGENT });
+    adapter.responseTimeout = 0.5; // fail fast if the no-audio path still hangs
+
+    const messages = await runTurn(adapter, () => {
+      // Tool-only: a function call with NO audio delta, terminated by response.done.
+      pushStreamingCall("call_weather", "get_weather", '{"city":"Paris"}');
+      push({ type: "response.done" });
+    });
+
+    const toolMsg = toolMessageOf(messages);
+    expect(toolMsg).toBeDefined();
+    expect(toolMsg!.role).toBe("tool");
+    const part = toolMsg!.content.find(
+      (p) => p.type === "tool-result" && p.toolName === "get_weather",
+    );
+    expect(part).toBeDefined();
+    expect(part).toMatchObject({
+      type: "tool-result",
+      toolCallId: "call_weather",
+      toolName: "get_weather",
+    });
+
+    // The real consumer recognizes it.
+    expect(stateWith(messages).hasToolCall("get_weather")).toBe(true);
+  });
+
   it("AC4 (variant) — a call delivered ONLY via output_item.done is surfaced", async () => {
     const adapter = buildAdapter({ apiKey: "test-key" });
 
--- a/javascript/src/voice/adapters/__tests__/openai-realtime-tool-calls.test.ts
+++ b/javascript/src/voice/adapters/__tests__/openai-realtime-tool-calls.test.ts
@@ -296,13 +296,6 @@ describe("OpenAIRealtimeAgentAdapter — realtime tool-call surfacing (#630)", (
       toolName: "get_weather",
     });
 
-    // PROOF — the actual message that lands in the run's messages.
-     
-    console.log(
-      "[#630 PROOF] tool message added to run:",
-      JSON.stringify(toolMsg, null, 2),
-    );
-
     // The real consumer (ScenarioExecutionState) recognizes it.
     const state = stateWith(messages);
     expect(state.hasToolCall("get_weather")).toBe(true);
@@ -342,6 +335,169 @@ describe("OpenAIRealtimeAgentAdapter — realtime tool-call surfacing (#630)", (
 
     // The real consumer recognizes it.
     expect(stateWith(messages).hasToolCall("get_weather")).toBe(true);
+  });
+
+  // --- #646: tool-only coverage + failure modes ---
+
+  it("#646 (AC3) — tool-only turn is consumable via hasToolCall/lastToolCall", async () => {
+    const adapter = buildAdapter({ apiKey: "test-key", role: AgentRole.AGENT });
+    adapter.responseTimeout = 0.5;
+
+    const messages = await runTurn(adapter, () => {
+      pushStreamingCall("call_ac3", "get_weather", '{"city":"London"}');
+      push({ type: "response.done" });
+    });
+
+    const state = stateWith(messages);
+    expect(state.hasToolCall("get_weather")).toBe(true);
+    const last = state.lastToolCall("get_weather");
+    expect(
+      last.content.some(
+        (p) => p.type === "tool-result" && p.toolName === "get_weather",
+      ),
+    ).toBe(true);
+  });
+
+  it("#646 (AC4) — streaming-args form terminates a tool-only turn", async () => {
+    const adapter = buildAdapter({ apiKey: "test-key", role: AgentRole.AGENT });
+    adapter.responseTimeout = 0.5;
+
+    const messages = await runTurn(adapter, () => {
+      pushStreamingCall("call_ac4s", "get_weather", '{"city":"Berlin"}');
+      push({ type: "response.done" });
+    });
+
+    const toolMsg = toolMessageOf(messages);
+    expect(toolMsg).toBeDefined();
+    const part = toolMsg!.content.find(
+      (p) => p.type === "tool-result" && p.toolName === "get_weather",
+    );
+    expect(part).toBeDefined();
+    expect(part).toMatchObject({
+      type: "tool-result",
+      toolCallId: "call_ac4s",
+      toolName: "get_weather",
+    });
+  });
+
+  it("#646 (AC4) — output-item form terminates a tool-only turn", async () => {
+    const adapter = buildAdapter({ apiKey: "test-key", role: AgentRole.AGENT });
+    adapter.responseTimeout = 0.5;
+
+    const messages = await runTurn(adapter, () => {
+      pushOutputItemCall("call_ac4o", "lookup_order", '{"id":99}');
+      push({ type: "response.done" });
+    });
+
+    const toolMsg = toolMessageOf(messages);
+    expect(toolMsg).toBeDefined();
+    const part = toolMsg!.content.find(
+      (p) => p.type === "tool-result" && p.toolName === "lookup_order",
+    );
+    expect(part).toBeDefined();
+    expect(part).toMatchObject({
+      type: "tool-result",
+      toolCallId: "call_ac4o",
+      toolName: "lookup_order",
+    });
+  });
+
+  it("#646 (AC5) — two calls on a tool-only turn both surface", async () => {
+    const adapter = buildAdapter({ apiKey: "test-key", role: AgentRole.AGENT });
+    adapter.responseTimeout = 0.5;
+
+    const messages = await runTurn(adapter, () => {
+      pushStreamingCall("call_1", "get_weather", '{"city":"NYC"}');
+      pushOutputItemCall("call_2", "get_time", '{"tz":"UTC"}');
+      push({ type: "response.done" });
+    });
+
+    const toolMsg = toolMessageOf(messages);
+    expect(toolMsg).toBeDefined();
+    const names = toolMsg!.content
+      .filter((p) => p.type === "tool-result")
+      .map((p) => (p as { toolName: string }).toolName)
+      .sort();
+    expect(names).toEqual(["get_time", "get_weather"]);
+
+    const state = stateWith(messages);
+    expect(state.hasToolCall("get_weather")).toBe(true);
+    expect(state.hasToolCall("get_time")).toBe(true);
+  });
+
+  it("#646 (AC7) — audio+tool turn still returns BOTH the audio message AND the tool message", async () => {
+    const adapter = buildAdapter({ apiKey: "test-key", role: AgentRole.AGENT });
+
+    const messages = await runTurn(adapter, () => {
+      pushAudioDelta();
+      pushStreamingCall("call_co", "fetch", '{"k":"v"}');
+    });
+
+    expect(messages.length).toBe(2);
+    expect(messages.some((m) => m.role === "assistant")).toBe(true);
+    expect(toolMessageOf(messages)).toBeDefined();
+  });
+
+  it("#646 (AC8) — turn 1 tool-only, turn 2 audio-only — no bleed", async () => {
+    const adapter = buildAdapter({ apiKey: "test-key", role: AgentRole.AGENT });
+    adapter.responseTimeout = 0.5;
+    await adapter.connect();
+    await socketReady;
+    await waitForType("session.update");
+
+    // Turn 1: no-audio tool-only — must push response.done to terminate drain.
+    const turn1 = await feedAndCall(adapter, () => {
+      pushOutputItemCall("call_t1", "get_weather", "{}");
+      push({ type: "response.done" });
+    });
+    expect(toolMessageOf(turn1)).toBeDefined();
+
+    // Turn 2: audio-only — terminates on tail-silence; must not carry turn-1's tool.
+    adapter.responseTimeout = 2; // restore default for the audio-bearing turn
+    const turn2 = await feedAndCall(adapter, () => {
+      pushAudioDelta();
+    });
+    await adapter.disconnect();
+
+    expect(toolMessageOf(turn2)).toBeUndefined();
+    expect(turn2.length).toBe(1);
+  });
+
+  it("#646 (AC9) — empty turn (no audio, no tool call) still times out", async () => {
+    const adapter = buildAdapter({ apiKey: "test-key", role: AgentRole.AGENT });
+    adapter.responseTimeout = 0.5;
+
+    await adapter.connect();
+    await socketReady;
+    await waitForType("session.update");
+    try {
+      // Empty turn: no audio, no function call — just response.done. The
+      // accumulator stays empty, so the new terminal path does NOT fire and
+      // the drain must still hit the receiveAudio timeout (proves the
+      // discriminator is the non-empty accumulator, not response.done alone).
+      await expect(
+        feedAndCall(adapter, () => {
+          push({ type: "response.done" });
+        }),
+      ).rejects.toThrow("receiveAudio timed out");
+    } finally {
+      await adapter.disconnect();
+    }
+  });
+
+  it("#646 (AC10) — malformed args degrade on a tool-only turn", async () => {
+    const adapter = buildAdapter({ apiKey: "test-key", role: AgentRole.AGENT });
+    adapter.responseTimeout = 0.5;
+
+    const messages = await runTurn(adapter, () => {
+      pushOutputItemCall("call_bad", "noop", "{not json");
+      push({ type: "response.done" });
+    });
+
+    const part = toolMessageOf(messages)!.content.find(
+      (p) => p.type === "tool-result" && p.toolName === "noop",
+    ) as { output: { type: string; value: unknown } };
+    expect(part.output).toEqual({ type: "text", value: "{not json" });
   });
 
   it("AC4 (variant) — a call delivered ONLY via output_item.done is surfaced", async () => {
--- a/javascript/src/voice/adapters/openai-realtime.ts
+++ b/javascript/src/voice/adapters/openai-realtime.ts
@@ -460,6 +460,18 @@ export class OpenAIRealtimeAgentAdapter extends VoiceAgentAdapter {
             this._finalizeToolCall(fc.call_id, fc.name, fc.arguments);
           }
         }
+      } else if (etype === "response.done" || etype === "response.cancelled") {
+        // Issue #646: a tool-only turn (function call, NO audio delta) would
+        // otherwise loop here forever and hit the receiveAudio timeout — the
+        // accumulated tool call is parsed but never returned. When the response
+        // is done and >=1 tool call has been finalized this turn, return an
+        // empty chunk so the base drain (drainAgentResponse) terminates cleanly
+        // and call() surfaces the role:"tool" message. A genuinely empty turn
+        // (done + EMPTY accumulator) must still fall through to the timeout —
+        // the non-empty accumulator is the discriminator, NOT response.done alone.
+        if (this._completedToolCalls.length > 0) {
+          return new AudioChunk({ data: new Uint8Array(0) });
+        }
       } else if (etype === "error") {
         const errDetail =
           (event as { error?: { message?: string } }).error ?? {};
--- a/python/scenario/voice/adapters/openai_realtime.py
+++ b/python/scenario/voice/adapters/openai_realtime.py
@@ -29,7 +29,9 @@ import base64
 import json
 import logging
 import os
-from typing import Any, ClassVar, List, Optional
+from typing import Any, ClassVar, List, Optional, cast
+
+from openai.types.chat import ChatCompletionMessageParam
 
 from ...config.voice_models import OPENAI_REALTIME_MODEL, OPENAI_STT_MODEL
 from ...types import AgentInput, AgentReturnTypes, AgentRole
@@ -604,12 +606,15 @@ class OpenAIRealtimeAgentAdapter(VoiceAgentAdapter):
 
         # One assistant message carrying ALL of this turn's tool calls — the
         # conventional OpenAI shape (one assistant turn, many tool_calls).
-        tool_call_message: dict[str, Any] = {
-            "role": "assistant",
-            "content": None,
-            "tool_calls": list(self._completed_tool_calls),
-        }
-        return [audio_message, tool_call_message]
+        tool_call_message: ChatCompletionMessageParam = cast(
+            ChatCompletionMessageParam,
+            {
+                "role": "assistant",
+                "content": None,
+                "tool_calls": list(self._completed_tool_calls),
+            },
+        )
+        return [cast(ChatCompletionMessageParam, audio_message), tool_call_message]
 
     async def _drain_agent_response(
         self, on_first_chunk=None
--- a/python/scenario/voice/adapters/openai_realtime.py
+++ b/python/scenario/voice/adapters/openai_realtime.py
@@ -32,7 +32,7 @@ import os
 from typing import Any, ClassVar, List, Optional
 
 from ...config.voice_models import OPENAI_REALTIME_MODEL, OPENAI_STT_MODEL
-from ...types import AgentRole
+from ...types import AgentInput, AgentReturnTypes, AgentRole
 from ..adapter import VoiceAgentAdapter
 from ..audio_chunk import AudioChunk
 from ..capabilities import AdapterCapabilities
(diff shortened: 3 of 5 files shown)


drewdrewthis:
### Sweep: audio-gated receive-loop return (no clean terminal path for a non-audio completion)

**Source:** this PR (#647), fixing #646.

Swept all voice adapters for the same anti-pattern — a receive/drain loop that returns ONLY on an audio-delta event and lets a terminal `response.done`/close fall through to a timeout.

**Must-fix: 0.** The OpenAI Realtime adapter was the only full occurrence (while-loop drain + audio-gated return + a tool-only-turn that can actually hit it). Fixed here in both SDKs.

**Clean:** Pipecat (JS socket-close handler / Py `stop` event), Gemini Live (`StopAsyncIteration` + `turn_complete` → empty chunk), Twilio (queue delegation), Composable (`_turn_output_emitted` guard, no drain loop).

**Similar risk (out of scope — tracked separately):** ElevenLabs (JS+Py) and the generic WebSocket adapter (Py) have audio-gated loops with no socket-close/terminal path, so a *silent completion* could hang to the timeout. Neither supports tool calls, so the #646 defect can't occur there; the latent silent-hang is a different concern and will be filed as its own issue rather than widening this focused fix.

drewdrewthis:
## Review verdict: READY

Five reviewers (principles, hygiene, security, test, proof) + 3 personas ran on the diff. The fix logic is endorsed **clean** (principles: "right altitude, accumulator is the cohesive discriminator"; proof-reviewer: **11/11 ACs map to real load-bearing tests**, marker SHA-matches HEAD). Security clean (synthetic fixtures, debug-log removal confirmed). The two blocking test-quality concerns were **fixed this turn** and the JS suite re-ran green (17 passed); the should-improves were folded in the same pass.

### Resolved this turn (were blocking)

| # | Concern | Resolution |
|---|---------|------------|
| 1 | **[test][hygiene]** JS AC7 coexistence test relied on tail-silence, no explicit `response.done` — tested an incidental property, not the coexistence terminal | Added `push({type:"response.done"})` after the audio+tool events (commit `9757a80`) |
| 2 | **[test]** JS AC9 asserted only a message substring (`.toThrow("…timed out")`); PY twin uses type-specific `pytest.raises(asyncio.TimeoutError)` | Now asserts `instanceof Error` AND message-contains, with `disconnect()` in `finally` (commit `9757a80`) |

### Folded (non-blocking, same commit)

- **[hygiene]** JS AC4 duplicate test names → renamed the output-item sibling to `#646 (AC4, variant)` per the suite's existing `(variant)` convention.
- **[principles]** PY `response.done` branch now carries a one-line cross-reference to the distinct drain-re-entry terminal path. Comment-only.

### Non-blocking / follow-up

- **[completeness]** Not a gap: the cross-SDK tool-call *shape* divergence (PY `tool_calls` vs JS `role:"tool"`/`tool-result`) is **out of scope by design** — tracked by #640. Each SDK's fix correctly surfaces its OWN consumer shape; AC2 explicitly forbids porting.
- **[synthesizer] New issue (filing separately):** the **ElevenLabs** (JS+Py) and generic **WebSocket** (Py) adapters have the same audio-gated drain shape with no socket-close/terminal path, so a *silent completion* could hang to the timeout. Neither supports tool calls, so the #646 defect can't occur there — this is a latent silent-hang, a different concern, deliberately not widened into this PR (see the sweep comment above).
- **[proof-reviewer][test]** PY test names `test_ac8_*`/`test_ac9_*` (cited for AC6/AC7) are legacy #635 names off-by-numbering from this issue's AC numbers — the AC table maps them correctly; a rename is optional cleanup, not blocking.

CI re-confirmation pending (PR is **draft** until CI green). The ✅ AC ticks are verified-in-tree this turn, not CI-green.

### Scoped re-review — since 9757a80 (HEAD aaab036)

The post-review commit `aaab036` (pyright type-narrow on `call()`'s return — 2 `cast`s + 2 imports in `openai_realtime.py`) was re-reviewed scoped to that diff. **Clean — no Must-Fix, no Should-Improve.** The reviewer traced the cast path: `audio_message` is always `create_audio_message(...) -> ChatCompletionMessageParam` from the base `call()`, so both casts are runtime-truthful; the dict-literal `cast` matches the existing pattern at `messages.py:69`; removing `# type: ignore[override]` for narrow casts is a net type-safety improvement. Gating evidence: `uv run pyright scenario/voice/adapters/openai_realtime.py` → 0 errors this turn. One out-of-scope follow-up: the base `VoiceAgentAdapter.call()` return type is over-broad (`AgentReturnTypes`) when it only returns a single message — tightening it would remove the need for the `audio_message` cast (`adapter.py:161`).

drewdrewthis:
**prove-it --since 9757a80 (scoped):** only changed surface = `openai_realtime.py` type-narrowing cast on the call() return (pyright fix). AC behavior unchanged — cast is runtime-identical; `pytest tests/voice/test_realtime_tool_calls.py` → 18 passed, `pyright .` → 0 errors this turn. All 11 ACs carried forward. **11/11 PASS.**

drewdrewthis:
## Correction — tool-call support claim (re: sweep + verdict above)

A reviewer flagged that my "neither ElevenLabs nor WebSocket supports tool calls" claim (in the sweep comment and the review verdict) was an **unverified assumption**. Verifying it against the source corrects the record:

- **ElevenLabs DOES support tool calls at the wire.** Its adapter docstring (`python/scenario/voice/adapters/elevenlabs.py:19-21`) lists `client_tool_call` among documented events that are **silently skipped** — "the provisioned test agent does not trigger them." So the adapter is tool-call-*aware* but drops the event. My earlier "does not support tool calls" was wrong for ElevenLabs.
- **WebSocket (Py)** and **ElevenLabs (JS)** grep clean — no tool-call handling.

**Why this matters:** it makes the latent audio-gated-drain hang in ElevenLabs **tool-call-reachable** for a real (non-test) ElevenLabs agent, not hypothetical — a *stronger* reason to track the follow-up, though still correctly **out of scope for #646** (this PR fixes the OpenAI Realtime adapter; widening to ElevenLabs/-WebSocket would muddy a focused diff). The follow-up issue (being filed) carries the corrected framing.

The #646 fix itself is unaffected — it is scoped entirely to the OpenAI Realtime adapter, which is verified and tested.

drewdrewthis:
## CI status (HEAD aaab036)

**The gating code job `test (3.12)` PASSES** — including the pyright Type-check step that the prior commit fixed (`uv run pyright .` → 0 errors). `javascript-complete`, `python-complete`, `docs-complete` all green.

**Red checks are GitHub auth-infra, NOT this diff:** `Analyze (python)` / `Analyze (javascript-typescript)` (CodeQL) and `preflight` fail at a **SARIF-upload / PR-context step** with `Requires authentication — security-events: write` / `Requires authentication`. CodeQL completes its full analysis (all queries interpreted, SARIF exported) and only fails on the authenticated **upload** — a token-permission condition affecting the repo right now, not a code finding. CodeQL is green on `main`. A re-run once the token condition clears should go green.

This PR is correctly **draft**: it stays draft until CI is fully green. The remaining failures are infra; the code is verified (11/11 ACs in-tree, both suites green locally). Marking ready + merge is the maintainer's call once CI infra clears.

drewdrewthis:
## CI status — VERIFIED (supersedes my earlier "auth-infra" note)

I tested the red checks rather than assuming. Result: **every CODE workflow on HEAD `aaab036` is green**; the only failure is a **broken repo automation workflow**, not this diff.

| Workflow (HEAD aaab036) | Result |
|---|---|
| `python-ci` (pytest + pyright) | ✅ success |
| `javascript-ci` | ✅ success |
| `docs-ci` | ✅ success |
| `Conventional Commits` | ✅ success |
| `Auto-approve for validated Low-Risk or Firefighting Check` | ❌ failure |

The lone red workflow fails at its **first step (`preflight` → "Resolve PR context", 5s)** and **`gh run rerun` rejects it with `run cannot be rerun; its workflow file may be broken`.** That is a **repo-level CI-config defect in the auto-approve/firefighting bot workflow definition** — it cannot bootstrap on any PR, is maintainer-owned, and is unrelated to and unfixable from this branch. The `Analyze`/CodeQL check entries in the rollup are surfaced by that same bot context (SARIF upload needs `security-events: write`, which the broken workflow never reaches).

**Net: the code is green.** No code gate is failing. The red check is infrastructure (a broken repo workflow file), verified by re-run rejection — not an assumption.

Also: an expert review of the fix approach (scenario-expert) came back **"correct, ship it"** — confirmed the discriminator (non-empty accumulator), empty-chunk drain termination, and the cross-SDK shape divergence being deferred to #640. Its one flagged "gap" (`response.failed`/`response.incomplete` events) was verified against the OpenAI Realtime SDK type defs to be a **non-issue**: those are status VALUES inside `response.done` (which the fix already terminates on), not separate events — no code change needed.

drewdrewthis:
## Live end-to-end verification (real OpenAI Realtime API)

The fix was previously covered only by hermetic mock-WS unit tests. It is now **verified against the real OpenAI Realtime WebSocket API** — a genuine tool-only turn surfaces the tool call instead of timing out.

**Run:** drove one `call()` turn through `OpenAIRealtimeAgentAdapter` against the live GA Realtime API, with a system prompt + `get_weather` tool forcing a function call and no spoken audio.

**Wire trace — 19 events, ZERO `response.output_audio.delta`** (a true tool-only turn on the real wire):

```
1. session.created          11. response.function_call_arguments.delta
2. session.updated          12. response.function_call_arguments.delta
3. response.created         13. response.function_call_arguments.delta
4. response.output_item.added  14. response.function_call_arguments.delta
5. conversation.item.added  15. response.function_call_arguments.delta
6-10. function_call_arguments.delta ×5   16. response.function_call_arguments.done
                            17. conversation.item.done
                            18. response.output_item.done
                            19. response.done
```

**`call()` returned the tool call** (would have hit the 90s `response_timeout` deadline without the fix):

```json
{
  "role": "assistant",
  "content": null,
  "tool_calls": [
    { "id": "call_Zs0sEhVD...", "type": "function",
      "function": { "name": "get_weather", "arguments": "{\"city\": \"Paris\"}" } }
  ]
}
```

The fix branch at `response.done` (`if self._completed_tool_calls: return AudioChunk(data=b"")`) fired and terminated the drain loop; control returned to `call()`, which surfaced the already-parsed tool call. Confirms the hermetic regression test reflects real-wire behavior.

_Keyless hermetic suite (PY 18 / JS 17) still covers the path in CI; this live run is the end-to-end confirmation that the mock faithfully models the real API's tool-only turn._

drewdrewthis:
## Ready for review — CI status clarification

**Both required checks are green:** `python-complete` ✅ and `javascript-complete` ✅ (the two contexts required by branch protection on `main`). PR is `MERGEABLE`.

**The three red checks are a repo-infra token-permission issue, NOT this PR's code:**

| Red check | Root cause | Required? |
|---|---|---|
| `Analyze (python)` (CodeQL) | analysis ran clean (zero findings); the **SARIF upload** step fails with `Requires authentication` / `security-events: write` missing on the runner token | No |
| `Analyze (javascript-typescript)` (CodeQL) | same — upload-only failure, analysis itself passed | No |
| `Auto-approve …` | same token-permission family | No |

The CodeQL *analysis* completed and found nothing (`code-scanning/alerts` → "no analysis found" confirms the upload, not the scan, is what failed). These jobs need a repo-settings / workflow-`permissions:` fix (`security-events: write`) by a maintainer — they can't be fixed from a PR branch.

**Verification on this PR:** prove-it 11/11 ✅, review verdict READY ✅, hermetic suites (PY 18 / JS 17) ✅, and a [live OpenAI Realtime end-to-end run](https://github.com/langwatch/scenario/pull/647#issuecomment-4672387296) confirming the tool-only-turn fix on the real wire ✅.

rogeriochaves (approved):

PR #652, opened 2026-06-10
fix(voice/#498): surface attributable first-chunk timeout (phase + timeout + cause)

drewdrewthis:
## Why

Closes the **creds-free diagnostic half** of #498. When a voice agent fails to send its **first** audio chunk within `response_timeout`, the adapter's first-chunk `recv_audio` was **unwrapped** — a bare `asyncio.TimeoutError` (empty `str()`, no context) propagated. Even with the executor's #500 type-name fallback, operators only saw `[PipecatAgentAdapter] TimeoutError`: indistinguishable from a mid-response (tail-silence) cutoff, and missing the configured timeout value.

**References #498** — the *primary* root cause (**why** the 2nd Pipecat turn times out: agent hang vs `response_timeout` too short vs VAD-on-synthetic-audio vs a swallowed background-task crash) needs a **live Pipecat run** and is out of scope for this slice. #498 stays open for it.

## What changed

- New `FirstChunkTimeoutError(asyncio.TimeoutError)` in `scenario/voice/adapter.py` carrying `.timeout` (the configured budget) + `.phase` (`"first-chunk"`), with a non-empty message embedding both. It **subclasses** `asyncio.TimeoutError`, so the executor wrap (`#500`) and any existing `except asyncio.TimeoutError` are unaffected.
- `_drain_agent_response` wraps **only** the first-chunk `recv_audio` and re-raises `FirstChunkTimeoutError(...) from err` (chains the original transport timeout for an intact stack). The tail-silence recv is **untouched** — it still ends the drain silently and returns the chunks collected so far.

Before → after operator string:
`[PipecatAgentAdapter] TimeoutError` → `[PipecatAgentAdapter] agent did not send its first audio chunk within <response_timeout>s (phase=first-chunk)`

## Acceptance Criteria (creds-free diagnostic slice; ac-reviewed)

- **AC1** — first-chunk timeout escapes as a `TimeoutError` with a **non-empty** message, chaining the original transport timeout (`__cause__`). ✅ `test_first_chunk_timeout_is_attributable`
- **AC2** — the message carries a **code-owned phase marker** (`_FIRST_CHUNK_PHASE`, not an English substring) and exposes the configured timeout as a **typed `.timeout` attribute** (not a brittle string scrape). ✅ same test
- **AC3** (regression guard) — the tail-silence path is **unchanged** (returns collected chunks, does not raise). ✅ `test_tail_silence_timeout_ends_drain_without_raising`
- **Deferred (live-gated)** — primary 2nd-turn timeout root cause.

## Test plan

```
cd python && uv run pytest tests/voice/test_drain_timeout_surfacing.py
```
- `test_first_chunk_timeout_is_attributable` — RED before this change, GREEN after.
- `test_tail_silence_timeout_ends_drain_without_raising` — GREEN throughout (guards no regression).
- Local validation: target 2 passed · `test_adapter_base.py` 3 passed · voice dir collects 377 · `pyright scenario/voice/adapter.py` 0 errors.

## How I can prove it

No playable artifact — the change is an exception-surfacing diagnostic. The regression net is the two unit tests above (the first exercises the real inherited `_drain_agent_response` via a dummy adapter raising a bare `asyncio.TimeoutError`).

### DIFF
--- a/python/scenario/voice/__init__.py
+++ b/python/scenario/voice/__init__.py
@@ -9,6 +9,7 @@ Public surface:
     - VoiceAgentAdapter — base class for voice-capable agents
     - AudioChunk — canonical internal audio (PCM16 @ 24kHz mono)
     - AdapterCapabilities / UnsupportedCapabilityError — capability matrix
+    - FirstChunkTimeoutError — attributable first-chunk recv timeout
     - VoiceRecording / VoiceEvent / LatencyMetrics — result-side types
     - AudioSegment — per-speaker slice of the recording
     - synthesize / STTProvider / set_stt_provider / get_stt_provider —
@@ -21,7 +22,7 @@ Public surface:
 
 from __future__ import annotations
 
-from .adapter import VoiceAgentAdapter
+from .adapter import FirstChunkTimeoutError, VoiceAgentAdapter
 from .adapters import (
     ComposableVoiceAgent,
     ElevenLabsAgentAdapter,
@@ -62,6 +63,7 @@ __all__ = [
     "ElevenLabsAgentAdapter",
     "ElevenLabsSTTProvider",
     "ElevenLabsVoiceAgent",
+    "FirstChunkTimeoutError",
     "GeminiLiveAgentAdapter",
     "InterruptionConfig",
     "LatencyMetrics",
--- a/python/scenario/voice/adapter.py
+++ b/python/scenario/voice/adapter.py
@@ -31,6 +31,32 @@ from .messages import create_audio_message, extract_audio
 from .recording import AudioSegment, VoiceEvent
 
 
+_FIRST_CHUNK_PHASE = "first-chunk"
+"""Phase marker for the first-chunk recv timeout (used in FirstChunkTimeoutError)."""
+
+
+class FirstChunkTimeoutError(asyncio.TimeoutError):
+    """Raised when the agent fails to send its first audio chunk within ``response_timeout``.
+
+    WHY this subclass exists: operators could not distinguish a first-chunk hang
+    (agent never spoke — wrong endpoint, VAD never fired, response_timeout too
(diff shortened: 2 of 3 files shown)


drewdrewthis:
## Review verdict: READY

Re-reviewed at HEAD `5c95fb0` (was NOT-READY at `d7eb1d6`). All 3 blocking findings resolved + re-validated this turn; the fix + test were always correct — these were API-completeness + clarity.

### Resolved since last review
1. **[hygiene][principles] Diagnostic error now exported** — `FirstChunkTimeoutError` added to `scenario/voice/__init__.py` import + `__all__` (mirrors the `UnsupportedCapabilityError` precedent); `from scenario.voice import FirstChunkTimeoutError` verified working. Operators can now `except` it and read `.timeout`/`.phase`.
2. **[principles] Redundant `phase` param dropped** — `__init__(*, timeout)`; `_FIRST_CHUNK_PHASE` folded in (`.phase` attr + in-message `(phase=first-chunk)` marker preserved). Call site simplified; test assertions unchanged.
3. **[hygiene] Test stubs trimmed** — both dummy adapters → bare `AdapterCapabilities()` (verified `_drain_agent_response` reads no capability field). Verbose import-comment trimmed to 2 lines; inconsistent `# pragma: no cover` removed.

### Evidence (re-verified this turn)
- `pytest test_drain_timeout_surfacing.py test_adapter_base.py` → 5 passed; `from scenario.voice import FirstChunkTimeoutError` → ok; pyright 0 errors.
- Core logic untouched: `git diff d7eb1d6..5c95fb0 -- adapter.py` removes no drain/deadline/recv line.

### New Issue (out of scope, tracked)
- `elevenlabs.py:214` + `openai_realtime.py:237` bare-timeout treatment; the test module docstring's stale `adapter.py:205` ref (now line 232).

PR #659, opened 2026-06-11
fix(voice): guard response.create on active response in recv_audio

### LINKED ISSUE #657: OpenAI Realtime adapter fires response.create unconditionally after audio commit — races an active response into a drain timeout
## For Humans

**`recv_audio` fires `response.create` unconditionally after `input_audio_buffer.commit`, racing an in-flight response into a server rejection and drain timeout.**

When user audio arrives while the previous agent response is still streaming, the adapter commits the buffer and sends `response.create` without checking whether a response is already active. The server rejects with `"Conversation already has an active response in progress: resp_<id>"`. No audio delta follows; the drain loop runs to its deadline and the run dies as `RuntimeError: TimeoutError`.

**Ask:** Add a `not self._response_active` guard to the user-audio branch of `recv_audio` (mirroring the guard already present on the agent-turn branch at line 423).

_Status: `bug` · `not-ready` · `P2-medium` · `needs-plan`. Full spec below._

### Symptoms / context

- **Error (explicit shape):** `RuntimeError` wrapping `"Conversation already has an active response in progress: resp_<id>"` — the server rejects the duplicate `response.create`.
- **Error (dominant shape in the wild):** `RuntimeError: TimeoutError` / `asyncio.TimeoutError: OpenAIRealtimeAgentAdapter: recv_audio timed out` — the server silently ignores the racing `response.create` (or the rejection event arrives before any audio delta), so the drain loop runs to its timeout.
- **Live repro observed 2026-06-11** in `langwatch/ai-interviewer-scenario-demo` voice 2×2 matrix (`test_hardened_unqualified_voice`): 10-crash streak. One crash surfaced the explicit server error `"Conversation already has an active response in progress: resp_DpVmzBHeAk4XwtaAwCjfw"`; nine presented as `RuntimeError: TimeoutError` (the dominant face of the race). The `TimeoutError`-face crashes were previously attributed entirely to issue #646 (tool-only drain timeout); after installing #646's fix (commit `a7c02e0` on `main`) the cell passed on the first attempt in 75 s — confirming the race is a contributing cause of the timeout face, not merely a separate error shape.
- **Demo-side mitigation exists:** `harness/realtime_adapter.py` `GARealtimeInterviewerAdapter.recv_audio` drains the active response before delegating to the base (`_pending_audio_bytes > 0 and self._response_active` → `_drain_until_response_idle`) and retries once on the rejection string. Every user of the base adapter without a subclass override is exposed.

### Root cause

**`recv_audio` user-audio branch fires `response.create` unconditionally; the agent-turn branch already guards; the asymmetry is the bug.**

`python/scenario/voice/adapters/openai_realtime.py`, lines 407–411 (main as of investigation):

```python
if self._pending_audio_bytes > 0:
    await self._ws.send(json.dumps({"type": "input_audio_buffer.commit"}))
    await self._ws.send(json.dumps({"type": "response.create"}))   # ← no guard
    self._pending_audio_bytes = 0
    self._agent_turn_pending = False
```

The sibling agent-turn branch at line 423 already carries the guard:

```python
elif self._agent_turn_pending and not self._response_active:   # ← guarded
    await self._ws.send(json.dumps({"type": "response.create"}))
```

`self._response_active` is set to `True` on `response.created` (line 474) and cleared on `response.done` / `response.cancelled` (line 501). It is initialized `False` in `__init__` (line 131). The flag already exists and is already maintained correctly for the agent-turn path; the user-audio path simply does not consult it.

**Race window:** The OpenAI Realtime API is full-duplex. Between the executor's last `recv_audio` call (which may still be streaming audio deltas for the agent turn) and the next `send_audio` / `recv_audio` cycle, `response.done` may not yet have arrived. When the frame boundary is tight — fast candidate reply, slow response completion, tool-call turn — the window is reliably hit.

**`_response_active` flag correctness:** confirmed by reading `__init__` (line 131: `False`), the `response.created` handler (line 474: `True`), the `response.done`/`response.cancelled` handler (line 501: `False`), and the agent-turn guard (line 423: `not self._response_active`). The flag is a sound gate; it is only absent from the user-audio branch.

### Strategies considered

| Strategy | Assessment |
|---|---|
| **Guard `response.create` on `not self._response_active` (user-audio branch)** | ✅ Preferred. Mirrors the existing agent-turn guard exactly. One-line change. No protocol side-effects: `input_audio_buffer.commit` can still be sent immediately; only `response.create` is deferred. |
| **Drain (consume events) until `_response_active` clears before committing** | Works (demo subclass does this). Heavier — consumes and discards events from the prior response mid-recv. Adds latency and complexity to the hot path. Appropriate as a fallback if the guard alone is insufficient (e.g. VAD-triggered responses that the adapter cannot track), but not the right first fix. |
| **Retry once on the rejection error** | Also works (demo subclass does this as a second safety layer). Correct as defense-in-depth but not a substitute for the guard — it relies on the error string being stable and cannot prevent the timeout face (where the server silently ignores rather than rejects). |
| **Raise `response_timeout` to absorb the drain** | Does not fix the root cause. A racing `response.create` either gets rejected (RuntimeError) or is silently ignored; a longer timeout only delays the crash. |

1. **The fix site is `recv_audio`, user-audio branch, line 408.** The `response.create` send (only — not the `commit`) should be guarded on `not self._response_active`. The `input_audio_buffer.commit` can fire unconditionally; it is idempotent and the server accepts it regardless of response state.
2. **The demo subclass (`GARealtimeInterviewerAdapter`) should be deleted or simplified** once the upstream guard lands. Its `recv_audio` override exists solely to paper over this bug (per docstring: "delete on the upstream fix"). The override must remain present until the fix is released; it is not a regression hazard but it is dead weight after.
3. **Hermetic repro path:** mirror the `_MockWS` pattern from `python/tests/voice/test_realtime_tool_calls.py`. Pre-load the mock with a `response.created` event (no `response.done` yet) to simulate an in-flight response, then call `recv_audio` with pending audio bytes. The unguarded code will emit `response.create`; the fix should suppress it and wait. No live API connection needed.
4. **The guard must be paired with a mechanism to send `response.create` once the active response clears.** Simply skipping the send without deferring it would silently swallow the turn. One approach: if `_response_active`, defer the `response.create` by setting `_agent_turn_pending = True` (the agent-turn branch will then fire it on the next `recv_audio` call after `response.done` arrives). The implementer should evaluate this shape; the plan step owns the approach.
5. **No migration surface; no rollback AC needed.** Pure in-memory state guard with no storage or external side effects.

### Caveats

- The race window is narrower on slow networks (more time for `response.done` to arrive before user audio). Tests on fast connections or with synthetic audio (zero-latency mock) hit it more reliably.
- VAD-triggered sessions (server-side VAD) may produce `response.created` events the adapter has not yet processed when `recv_audio` is called. The flag-based guard is correct for manual-VAD sessions; the implementer should verify it also covers `_response_active` transitions for VAD-mode sessions.
- The demo subclass drains events from the prior response during its pre-commit wait. The upstream fix does not need to replicate this drain — the executor's `_drain_agent_response` already handles it. The guard's job is only to block the premature `response.create`, not to consume the prior response's tail events.

## Acceptance Criteria

- [ ] **AC1 — guard present:** `recv_audio` does NOT send `response.create` when `_response_active is True` and `_pending_audio_bytes > 0`. Evidence: hermetic `_MockWS` test — pre-load mock with `response.created` (no `response.done`), call `recv_audio` with pending audio, assert `response.create` is absent from `mock_ws.sent`.
- [ ] **AC2 — response.create deferred, not dropped (ordering):** After `_response_active` clears (mock yields `response.done`), `response.create` IS sent on the same `recv_audio` call that processes `response.done` (or the immediately following call — implementer pins one). Evidence: mock test captures the full send sequence and asserts that no `response.create` entry appears in `mock_ws.sent` until AFTER `response.done` has been yielded by the mock — ordering is asserted on `mock_ws.sent` index positions, not mere presence.
- [ ] **AC3 — single commit, single create:** Across the full event sequence (guard fires, `response.done` received, deferred send fires), `mock_ws.sent` contains `input_audio_buffer.commit` exactly once and `response.create` exactly once. Evidence: `mock_ws.sent.count("input_audio_buffer.commit") == 1` and `mock_ws.sent.count("response.create") == 1` asserted in the same test.
- [ ] **AC4 — agent-turn branch unaffected:** `recv_audio` with `_agent_turn_pending=True` and `_response_active=False` still sends `response.create` exactly once. Evidence: targeted `_MockWS` test asserts `mock_ws.sent.count("response.create") == 1`.
- [ ] **AC5 — no regression on normal path:** `recv_audio` with `_pending_audio_bytes > 0` and `_response_active=False` (the common, uncontested path) still sends `commit` then `response.create` as before. Evidence: existing unit tests pass; or a mock test asserting both sends fire.

### Scope

- [ ] **AC-scope — manual VAD only:** This fix targets manual-VAD sessions (`turn_detection: None`, configured at line 316). Server-VAD mode is explicitly OUT OF SCOPE for this fix. No AC or test exercises server-VAD behavior; the PR must not claim correctness for server-VAD sessions.

- [ ] **AC6 — explicit rejection face (falsifiable):** A `_MockWS` test that bypasses the guard (or runs against the pre-fix code) and injects the server error event `"Conversation already has an active response in progress"` asserts that the adapter raises `RuntimeError`. Evidence: `pytest.raises(RuntimeError)` assertion — not an observation of live-repro behavior.
- [ ] **AC7 — timeout face eliminated:** A `_MockWS` event sequence that pre-fix triggers `asyncio.TimeoutError` (in-flight response, user audio committed, `response.done` never arriving before deadline) post-fix resolves by deferring `response.create` until `response.done` is yielded within the sequence, then returns a non-empty `AudioChunk`. Evidence: mock test that would previously raise `asyncio.TimeoutError` now returns a valid `AudioChunk` without timing out.

> **PR checklist note (not an AC):** The `GARealtimeInterviewerAdapter.recv_audio` override in `langwatch/ai-interviewer-scenario-demo` is a demo-side mitigation pending this upstream fix. Once this fix ships, that override is dead weight and can be removed by downstream consumers on upgrade. The PR description should note this.

### Approach

Guard the `response.create` send in the user-audio branch on `not self._response_active`. The `input_audio_buffer.commit` continues to fire unconditionally (idempotent; server accepts it in any state). When the guard fires, defer the `response.create` by setting `_agent_turn_pending = True`; the existing agent-turn branch (`elif self._agent_turn_pending and not self._response_active`) will then fire exactly one `response.create` on the next `recv_audio` call after `response.done` clears `_response_active`. All invariants (single commit, single create, ordering) are pinned by hermetic `_MockWS` tests — no live API needed.

### Sequencing

1. Write failing `_MockWS` tests: guard suppresses `response.create` when active; deferred create fires after `response.done`; `input_audio_buffer.commit` appears exactly once; `response.create` appears exactly once; ordering asserted on index positions in `mock_ws.sent`.
2. Implement the guard + deferral in `recv_audio` (user-audio branch only).
3. Verify agent-turn branch and normal-path regressions: existing voice tests (`test_realtime_tool_calls.py`) pass unchanged.
4. PR description notes downstream demo-subclass removal.

### Key decisions

- **Defer, do not drop** — silently skipping `response.create` would swallow the user's turn.
- **Guard only `response.create`** — `commit` is idempotent and always safe.
- **Manual-VAD only** — server-VAD is out of scope; no changes to VAD-mode paths.
- **No drain in adapter** — the executor's `_drain_agent_response` owns draining; the adapter's guard job is only to block the premature send.

### Risks / open questions

The primary risk is interaction of the deferral flag (`_agent_turn_pending = True`) with the `_pending_audio_bytes` reset (line 410: `self._pending_audio_bytes = 0`) and the `_response_ever_active` short-circuit near line 428. The implementer must confirm which branch fires the deferred send and that resetting `_pending_audio_bytes` to zero before `response.done` arrives does not cause the agent-turn branch to skip or double-fire `response.create`.

### Out of scope

Server-VAD sessions; demo-repo subclass removal; retry-on-rejection defense layer.

## Evidence

- Live repro 2026-06-11: explicit server error `"Conversation already has an active response in progress: resp_DpVmzBHeAk4XwtaAwCjfw"` observed in `langwatch/ai-interviewer-scenario-demo` voice matrix, 1 of 10 crashes. Remaining 9 presented as `RuntimeError: TimeoutError` (same race, timeout face).
- Root cause file:line: `python/scenario/voice/adapters/openai_realtime.py:407-411` (unconditional `response.create`); guard comparison at line 423 (`not self._response_active`).
- Demo mitigation: `harness/realtime_adapter.py:472-529` (`GARealtimeInterviewerAdapter.recv_audio`).
- Hermetic repro shape: `python/tests/voice/test_realtime_tool_calls.py` `_MockWS` pattern.

</details>

drewdrewthis:
## Why

The OpenAI Realtime API silently drops a `response.create` sent while a response is already in flight. In multi-turn voice scenarios this caused `recv_audio` to hang until timeout — the committed user audio had no response kicked off, so the recv loop drained forever instead of producing a chunk. Closes #657.

## What changed

- **`_deferred_response_create` flag (separate from `_agent_turn_pending`)** — when user audio arrives while `_response_active` is true, defer the `response.create` via this flag instead of sending it immediately (which the server ignores). Keeping the flags separate preserves each flag's single meaning.
- **`response.done` / `response.cancelled` handler fires the deferred create** — once `_response_active` clears, the handler sends `response.create` and re-arms `_response_active`, so the next recv iteration picks up the response for the committed audio.
- **`_agent_turn_pending` cleared in deferred path** — the deferred path now clears `_agent_turn_pending` so a stale flag cannot fire a spurious second `response.create` after the deferred one completes.
- **`_deferred_response_create` reset in `call()` turn-start block** — prevents a turn interrupted mid-deferral from leaking the flag into the next turn's `response.done` handler.

## Test plan

Regression tests in `python/tests/voice/test_realtime_response_create_guard.py`:

```bash
cd python && uv run pytest tests/voice/test_realtime_response_create_guard.py -v
```

| Test | What it asserts |
|---|---|
| `test_ac1_response_create_suppressed_while_response_active` | guard suppresses `response.create` when `_response_active=True`; only commit sent |
| `test_ac2_deferred_response_create_fires_after_response_done` | deferred create appears in the send log *after* `response.done` is received |
| `test_ac3_exactly_one_commit_and_one_create` | exactly 1 commit + 1 create across the full race sequence |
| `test_ac4_agent_turn_branch_still_fires_response_create` | agent-turn branch unaffected by the guard (control) |
| `test_ac5_normal_path_commit_then_create` | normal (non-race) path still fires `response.create` immediately |
| `test_ac6_server_rejection_raises_runtime_error` | explicit server rejection of `response.create` raises `RuntimeError` |
| `test_ac7_race_sequence_returns_audio_chunk_not_timeout` | race sequence returns `AudioChunk` instead of timing out |
| `test_deferred_path_clears_agent_turn_pending` | deferred path clears `_agent_turn_pending` (line 428); mutation-test-confirmed |

All 8 pass locally. CI green on `4e2e858d`.

Feature file: `specs/realtime-response-create-guard.feature`

## Prove-it 2.7 — live demo output

Guard demonstrated on the **real** OpenAI Realtime API (`gpt-realtime-mini`), with no mock WebSocket. Key excerpt from the live wire log:

```
[   1.633s]     STATE  _deferred_response_create: False -> True  <<< GUARD FIRES
             (user audio committed while response in flight)

[105] RECV <- response.done
[106] SEND -> response.create   *** deferred response.create ON THE WIRE ***
[108] RECV <- response.created  (new response started — deferred turn picked up)

GUARD PROOF: PASS
[PASS] DEFERRED response.create sent AFTER response.done received
       first RECV response.done at log[105], last SEND response.create at log[106]
[PASS] Exactly 2 response.create on the wire (1 normal + 1 deferred)
[PASS] Two input_audio_buffer.commit on the wire (one per user turn)
[PASS] At least 2 response.done received (original + deferred response)
```

Full 311-line output: `/tmp/demo_659_guard_proof_output.txt` (run at commit `82d6c52c`; guard code unchanged in subsequent commits).

### DIFF
--- a/python/scenario/voice/adapters/openai_realtime.py
+++ b/python/scenario/voice/adapters/openai_realtime.py
@@ -425,6 +425,7 @@ class OpenAIRealtimeAgentAdapter(VoiceAgentAdapter):
                 # the response.done handler fires response.create once the in-flight
                 # response clears.
                 self._deferred_response_create = True
+                self._agent_turn_pending = False  # user spoke → consume turn signal even when deferred
             else:
                 await self._ws.send(json.dumps({"type": "response.create"}))
                 self._agent_turn_pending = False  # user spoke → per-turn signal consumed
--- a/python/scenario/voice/adapters/openai_realtime.py
+++ b/python/scenario/voice/adapters/openai_realtime.py
@@ -623,6 +623,7 @@ class OpenAIRealtimeAgentAdapter(VoiceAgentAdapter):
         """
         # Reset per-turn tool-call state so a prior turn's calls don't bleed
         # through (mirrors the transcript reset in _drain_agent_response).
+        self._deferred_response_create = False
         self._completed_tool_calls = []
         self._tool_call_accumulators = {}
 
--- a/python/scenario/voice/adapters/openai_realtime.py
+++ b/python/scenario/voice/adapters/openai_realtime.py
@@ -406,9 +406,17 @@ class OpenAIRealtimeAgentAdapter(VoiceAgentAdapter):
         # If send_audio was called since last recv, commit and request response.
         if self._pending_audio_bytes > 0:
             await self._ws.send(json.dumps({"type": "input_audio_buffer.commit"}))
-            await self._ws.send(json.dumps({"type": "response.create"}))
             self._pending_audio_bytes = 0
-            self._agent_turn_pending = False  # user spoke → per-turn signal consumed
+            if self._response_active:
+                # Race guard (#657): server rejects/ignores a duplicate
+                # response.create while a response is in flight, which causes the
+                # recv loop to drain to timeout instead of completing.  Defer by
+                # re-arming _agent_turn_pending so the response.done handler below
+                # fires response.create as soon as the in-flight response clears.
+                self._agent_turn_pending = True
+            else:
+                await self._ws.send(json.dumps({"type": "response.create"}))
+                self._agent_turn_pending = False  # user spoke → per-turn signal consumed
 
         # Gap 1: agent-speaks-first / multi-turn agent initiation.
         # When the executor signals an agent turn via notify_agent_turn() and
@@ -499,6 +507,13 @@ class OpenAIRealtimeAgentAdapter(VoiceAgentAdapter):
                 # Response finished or was cancelled — mark it so the next
                 # drain re-entry returns an empty chunk (clean exit).
                 self._response_active = False
+                # Issue #657: if user audio was committed while a response was
+                # in flight, _agent_turn_pending was set as a deferral flag.
+                # Now that _response_active is cleared, fire the deferred create.
+                if self._agent_turn_pending:
+                    await self._ws.send(json.dumps({"type": "response.create"}))
+                    self._agent_turn_pending = False
+                    self._response_active = True
                 # Issue #646: a tool-only turn (function call, NO audio delta) would
                 # otherwise loop here to the deadline and raise — the accumulated tool
                 # call is parsed but never returned. When the response is done and at
--- a/python/scenario/voice/adapters/openai_realtime.py
+++ b/python/scenario/voice/adapters/openai_realtime.py
@@ -142,6 +142,16 @@ class OpenAIRealtimeAgentAdapter(VoiceAgentAdapter):
         # proceed()/resume). Consumed (cleared) when the kick fires.
         self._agent_turn_pending: bool = speaks_first  # first turn armed if speaks_first
 
+        # Issue #657: distinct deferral flag for the user-audio race guard.
+        # Set when user audio arrives while a response is in flight; cleared and
+        # consumed in the response.done/cancelled handler to fire the deferred
+        # response.create.  Kept separate from _agent_turn_pending so each flag
+        # retains its single original meaning.
+        # Known limitation: a second user-audio commit while one create is already
+        # deferred collapses into this single boolean.  Manual-VAD single-turn flow
+        # makes that path unreachable in practice; FSM refactor tracked as follow-up.
+        self._deferred_response_create: bool = False
+
         # --- Issue #630: realtime function-call (tool-call) surfacing ---
         # In-progress function calls keyed by call_id. Each value is a dict
         # {"name": str|None, "arguments": str} assembled from the streaming
@@ -410,10 +420,11 @@ class OpenAIRealtimeAgentAdapter(VoiceAgentAdapter):
             if self._response_active:
                 # Race guard (#657): server rejects/ignores a duplicate
                 # response.create while a response is in flight, which causes the
-                # recv loop to drain to timeout instead of completing.  Defer by
-                # re-arming _agent_turn_pending so the response.done handler below
-                # fires response.create as soon as the in-flight response clears.
-                self._agent_turn_pending = True
+                # recv loop to drain to timeout instead of completing.  Defer via
+                # _deferred_response_create (distinct from _agent_turn_pending) so
+                # the response.done handler fires response.create once the in-flight
+                # response clears.
+                self._deferred_response_create = True
             else:
                 await self._ws.send(json.dumps({"type": "response.create"}))
                 self._agent_turn_pending = False  # user spoke → per-turn signal consumed
@@ -508,11 +519,11 @@ class OpenAIRealtimeAgentAdapter(VoiceAgentAdapter):
                 # drain re-entry returns an empty chunk (clean exit).
                 self._response_active = False
                 # Issue #657: if user audio was committed while a response was
-                # in flight, _agent_turn_pending was set as a deferral flag.
+                # in flight, _deferred_response_create was set as a deferral flag.
                 # Now that _response_active is cleared, fire the deferred create.
-                if self._agent_turn_pending:
+                if self._deferred_response_create:
                     await self._ws.send(json.dumps({"type": "response.create"}))
-                    self._agent_turn_pending = False
+                    self._deferred_response_create = False
                     self._response_active = True
                 # Issue #646: a tool-only turn (function call, NO audio delta) would
                 # otherwise loop here to the deadline and raise — the accumulated tool
--- a/python/tests/voice/test_realtime_response_create_guard.py
+++ b/python/tests/voice/test_realtime_response_create_guard.py
@@ -424,3 +424,47 @@ async def test_ac7_race_sequence_returns_audio_chunk_not_timeout():
         "Post-fix: guard defers response.create until response.done is processed, "
         "then the audio-delta sequence completes with a non-empty chunk."
     )
+
+
+# ---------------------------------------------------------------------------
+# Deferred path also consumes the agent-turn signal: when user audio is
+# committed while a response is in flight, recv_audio takes the deferral
+# branch (_deferred_response_create=True) and MUST also clear
+# _agent_turn_pending (openai_realtime.py line 428) — the user spoke, so the
+# pending agent-turn signal is consumed even though the create is deferred.
+# Without that clear, a later drain re-entry would fire a spurious
+# response.create off the still-pending agent-turn flag.
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.asyncio
+async def test_deferred_path_clears_agent_turn_pending():
+    """Deferred branch clears _agent_turn_pending (line 428) when user audio is committed mid-response."""
+    # response.created fires, but no response.done — response stays active and
+    # no audio delta arrives, so the drain terminates via TimeoutError from the
+    # MockWS (same shape as AC1). The flag clear happens in the preamble before
+    # the recv loop, so the timeout does not affect what we assert.
+    events = [
+        json.dumps({"type": "response.created"}),
+    ]
+    adapter, _mock_ws = _make_adapter(events, short_timeout=True)
+
+    # Both flags set: an agent turn is pending AND a response is already in
+    # flight. User audio committed below → preamble takes the deferral branch.
+    adapter._agent_turn_pending = True
+    adapter._response_active = True
+    adapter._pending_audio_bytes = 960  # 480 samples × 2 bytes
+
+    # recv_audio times out (no audio delta, no response.done) — expected.
+    with pytest.raises((asyncio.TimeoutError, RuntimeError)):
+        await adapter.recv_audio(timeout=0.2)
+
+    # The deferral branch must have consumed the agent-turn signal (line 428).
+    assert adapter._agent_turn_pending is False, (
+        "deferred path did not clear _agent_turn_pending; a later drain "
+        "re-entry would fire a spurious response.create off the stale flag"
+    )
+    # Sanity: it really was the deferral branch (flag set), not the else branch.
+    assert adapter._deferred_response_create is True, (
+        "expected the deferral branch (_deferred_response_create=True) to be taken"
+    )
--- a/python/tests/voice/test_realtime_response_create_guard.py
+++ b/python/tests/voice/test_realtime_response_create_guard.py
@@ -1,0 +1,507 @@
+"""
+Issue #657 — regression tests for recv_audio response.create race condition.
+
+The user-audio branch at lines ~407-411 of openai_realtime.py sends
+``response.create`` unconditionally after ``input_audio_buffer.commit``, even
+while ``self._response_active`` is True (set on response.created, cleared on
+response.done/response.cancelled).  The agent-turn elif at ~line 423 already
+guards on ``not self._response_active``.
+
+Fix (NOT in this file): add the same guard to the user-audio branch and defer
+the create by setting ``_agent_turn_pending=True`` so it fires after
+response.done clears ``_response_active``.
+
+Test layout
+-----------
+AC1 — guard present: response.create suppressed while response is active.
+AC2 — deferred response.create fires AFTER response.done (ordering asserted).
+AC3 — single commit + single create across the full guarded sequence.
+AC4 — control: agent-turn branch unaffected (should PASS now).
+AC5 — control: normal path (no active response) sends commit then create (should PASS now).
+AC6 — explicit server rejection raises RuntimeError.
+AC7 — pre-fix timeout face resolves to a valid AudioChunk post-fix.
+
+AC1/AC2/AC3/AC7 MUST FAIL on current (pre-fix) code.
+AC4/AC5 MUST PASS on current code.
+AC6 MUST PASS if the adapter already surfaces error events as RuntimeError.
+
+Mock strategy: hermetic _MockWS from test_realtime_tool_calls.py — queue-based,
+no live API key required.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import base64
+import json
+from typing import Any, List
+
+import pytest
+
+from scenario.voice.adapters.openai_realtime import OpenAIRealtimeAgentAdapter
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _make_pcm(n_samples: int = 480) -> bytes:
+    """Minimal silent PCM16 mono @ 24 kHz."""
+    return b"\x00\x00" * n_samples
+
+
+def _b64_pcm(n_samples: int = 480) -> str:
+    return base64.b64encode(_make_pcm(n_samples)).decode()
+
+
+class _MockWS:
+    """Queue-backed WebSocket mock.
+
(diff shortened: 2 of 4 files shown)


drewdrewthis:
### Sweep: unguarded `response.create` sends (race with active response)

Checked all `response.create` send sites in python/ + javascript/. Source: this PR's fix (#657).

**Must-fix (out of scope here, follow-up issue to be filed):**
| File | Line | Note |
|------|------|------|
| `javascript/src/voice/adapters/openai-realtime.ts` | 376 | `receiveAudio` unguarded send; JS adapter has no `_responseActive` tracking at all |
| `javascript/src/voice/adapters/openai-realtime.ts` | 680 | `sendText` bare unconditional send |
| `javascript/src/agents/realtime/realtime-agent.adapter.ts` | 189 | `handleInitialResponse` unguarded |
| `javascript/src/agents/realtime/realtime-agent.adapter.ts` | 231 | `handleAudioInput` commit+create unguarded |

**Review:** `python/scenario/voice/adapters/openai_realtime.py:691` `send_text` — unconditional send, same race shape.

**Clean:** all other occurrences are tests/docs.

drewdrewthis:
## Review verdict: READY

**Commit:** `ed4a8359b5ca5a3419f8f91f9d5356f62fe64076`
**Delta reviewed since:** `82d6c52c010e11628245900f25e073efb7ca5c82`

All prior blocking threads resolved. Delta review (two fix commits + rebase + uv.lock sync) surfaces no new blocking concerns. All three personas (Uncle Bob, Metz/Beck, Fowler) agree: proceed.

---

### Non-blocking (prose — do not block merge)

**From full review (prior):**

- **[hygiene]** `openai_realtime.py:428+431` — `_agent_turn_pending = False` appears identically in both branches; could hoist above the `if`. Clean, no test impact.
- **[principles]** `openai_realtime.py:525–528` — deferred `response.create` fires before the `_completed_tool_calls` tool-only-turn check at line 539; coupling is undocumented. Decide: add ordering comment or restructure.
- **[test-reviewer]** AC2 is a strict subset of AC3. Decide: merge or keep for clarity.
- **[hygiene]** `test_ac1_`…`test_ac7_` naming encodes spec numbers, not behavior. Decide.
- **[test-reviewer]** `tuple[str, str]` annotation inconsistency on `_MockWS.log`. Minor.

**From delta review (`82d6c52c`…`4e2e858d`):**

- **[principles]** `openai_realtime.py:624` — comment reads "per-turn tool-call state" but `_deferred_response_create` is a race-guard flag, not tool-call state. Broadening the comment to "per-turn state" or splitting with its own note would stop the next engineer misreading the grouping. **Decide.**
- **[hygiene]** `test_realtime_response_create_guard.py:441` — `test_deferred_path_clears_agent_turn_pending` breaks the `test_acN_` naming convention the rest of the file uses. **Decide** (rename to `test_ac8_deferred_path_consumes_agent_turn_signal` or similar).
- **[hygiene][test]** `test_realtime_response_create_guard.py:468` — second assertion `_deferred_response_create is True` is a sanity check on branch-taken, not a product invariant. Consider removing; AC1 already confirms the deferred branch ran. **Decide.**
- **[test][metz-beck]** `pytest.raises((asyncio.TimeoutError, RuntimeError))` broad catch masks which exit path was taken. If only `TimeoutError` is expected, tighten. **Decide.**
- **[metz-beck]** Docstring and section comment hard-code `"line 428"` — line numbers in prose rot on the next edit above them. Reference the behavior ("deferral branch"), not the address. **Fix** (trivial, but worth noting).

**From scoped re-review (`39c7810a`…`ed4a8359`, uv.lock sync only):**

- **[proof-reviewer]** AC-gating files (`openai_realtime.py`, `test_realtime_response_create_guard.py`, `specs/realtime-response-create-guard.feature`) are byte-identical between pre-rebase tip and HEAD. Uv.lock version bump (0.7.30→0.7.31) is inert to all ACs. No new concerns.

**New Issues (file as follow-ups):**

- **[hygiene]** `_MockWS` defined identically in 3 test files. Extract to `tests/voice/conftest.py`.
- **[test]** `_deferred_response_create = False` in `call()` (line 626) has no dedicated test. An interrupted turn leaving the flag set would silently drop the deferred create; a test for that edge belongs here.
- **[fowler][principles]** Two turn-start reset sites: `call()` (lines 626–628) and `_drain_agent_response()` (transcripts). Asymmetric membership; extract `_begin_turn()` to prevent next flag landing in only one.
- **[uncle-bob][fowler][metz-beck]** 4 booleans (`_response_active`, `_response_ever_active`, `_agent_turn_pending`, `_deferred_response_create`) model an implicit FSM with ~11 illegal combinations. Known limitation at line 150 (second deferred commit collapses silently) is a symptom. File FSM-refactor issue and assign before the fifth flag arrives.

---

### Prove-it summary

Live demo on real OpenAI Realtime API (`gpt-realtime-mini`) at commit `82d6c52c` (guard code unchanged since):
- Guard fires: `_deferred_response_create: False → True` during active response
- Deferred `response.create` sent AFTER `response.done` (log positions 106 > 105)
- Exactly 2 `response.create` on wire; no hang; response completed — `GUARD PROOF: PASS`

AC-gating files verified byte-identical `4e2e858d`→`ed4a8359` (0-line diff); `test (3.12)` CI green at HEAD.

python/tests/voice/test_realtime_response_create_guard.py
```diff
@@ -0,0 +1,426 @@
...
+        json.dumps({"type": "response.done"}),
+    ]
+    adapter, mock_ws = _make_adapter(events, short_timeout=True)
+
+    # In-flight response: _response_active=True, pending audio.
+    adapter._response_active = True
+    adapter._response_ever_active = True
+    adapter._pending_audio_bytes = 960
+
+    result = await adapter.recv_audio(timeout=2.0)
```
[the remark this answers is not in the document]
drewdrewthis: Fixed in b6ed2f3 — removed the unused binding.

python/scenario/voice/adapters/openai_realtime.py
```diff
@@ -406,9 +416,19 @@ async def recv_audio(self, timeout: float) -> AudioChunk:
...
-            self._agent_turn_pending = False  # user spoke → per-turn signal consumed
+            if self._response_active:
+                # Race guard (#657): server rejects/ignores a duplicate
+                # response.create while a response is in flight, which causes the
+                # recv loop to drain to timeout instead of completing.  Defer via
+                # _deferred_response_create (distinct from _agent_turn_pending) so
+                # the response.done handler fires response.create once the in-flight
+                # response clears.
+                self._deferred_response_create = True
+                self._agent_turn_pending = False  # user spoke → consume turn signal even when deferred
```
drewdrewthis, line 428: **[proof-reviewer] `_agent_turn_pending = False` in deferred branch has zero test coverage**

This line is the entire payload of commit `82d6c52c` ("fix(voice/#657): clear _agent_turn_pending in deferred response.create path") — yet no test asserts this transition. Mutation test: deleting this line leaves all 7 AC tests green. No test sets `_agent_turn_pending = True`, calls `recv_audio` while `_response_active = True` (the deferred path), and then checks that `_agent_turn_pending` is `False`. The behavioral fix is unwitnessed.

**Action: Fix** — add a test that enters the deferred branch (`_response_active=True` before `recv_audio`) and asserts `adapter._agent_turn_pending is False` afterward.

python/scenario/voice/adapters/openai_realtime.py
```diff
@@ -142,6 +142,16 @@ def __init__(
...
 
+        # Issue #657: distinct deferral flag for the user-audio race guard.
+        # Set when user audio arrives while a response is in flight; cleared and
+        # consumed in the response.done/cancelled handler to fire the deferred
+        # response.create.  Kept separate from _agent_turn_pending so each flag
+        # retains its single original meaning.
+        # Known limitation: a second user-audio commit while one create is already
+        # deferred collapses into this single boolean.  Manual-VAD single-turn flow
+        # makes that path unreachable in practice; FSM refactor tracked as follow-up.
+        self._deferred_response_create: bool = False
```
drewdrewthis, line 153: **[principles] `_deferred_response_create` not reset in `call()` turn-start block (lines 624–627)**

`call()` resets `_completed_tool_calls` and `_tool_call_accumulators` at turn start so prior-turn state does not bleed through. `_deferred_response_create` is new per-turn state with the same leak risk: if a turn exits abnormally (interrupt / timeout) while the flag is set, the next turn's `response.done` handler will fire a spurious `response.create`. Missing reset:

```python
# line 627 area — needs:
self._deferred_response_create = False
```

**Action: Fix** — add the reset alongside `_completed_tool_calls = []` and `_tool_call_accumulators = {}`.

rogeriochaves (approved):

PR #665, opened 2026-06-12
fix(#664): transcribe agent turns at runtime so the voice user simulator can read them

### LINKED ISSUE #664: Voice user simulator is blind on audio-only transports: every agent turn reaches it as "[audio message]"
## Symptom

Voice scenarios over `PipecatAgentAdapter` (and any adapter relying on the default `VoiceAgentAdapter.call()` with an audio-only transport, e.g. Twilio Media Streams) produce conversations where the user simulator and the agent under test talk completely past each other:

- the agent asks "What's your membership number?", the simulator replies with an unrelated confirmation of its own previous sentence
- the agent answers a question fully, the simulator asks the exact same question again on the next turn, sometimes three times in a row
- the final transcript in the report and in LangWatch looks superficially coherent, and the judge frequently passes the run anyway

Observed on 0.7.31 with a Pipecat bot behind `ws://localhost:8765/stream` (Twilio Media Streams protocol, mulaw 8k).

## Root cause

The simulator never sees a single word the agent says:

1. `PipecatAgentAdapter._recv_loop` enqueues `AudioChunk(data=pcm)` with no transcript. The Twilio Media Streams wire protocol carries audio frames only, so there is nothing to attach.
2. The default `VoiceAgentAdapter.call()` merges the drained chunks and returns `create_audio_message(merged, role="assistant")`. Since `merged.transcript` is `None`, the assistant message contains only an `input_audio` part, no `text` part.
3. `UserSimulatorAgent._generate_text` runs `_strip_audio_content` over the history before calling its text-only LLM. Assistant messages without a text part are replaced with the literal placeholder `"[audio message]"`.
4. The simulator therefore generates every turn knowing only the scenario description and its own previous messages. It improvises plausible caller lines, which is why transcripts look almost reasonable while no actual information exchange is happening.

The defect is invisible in reports because the judge path is different: `JudgeAgent` falls back to `transcribe_segments()` over the recording, so the judge and the saved transcript get real agent text after the fact. The transcripts users read are not the transcripts the simulator acted on.

Side finding while debugging: `PipecatAgentAdapter.capabilities` advertises `streaming_transcripts=True`, which is false for this transport. It also makes `interrupt(after_words=N)` pass its capability gate and then poll an `adapter.streaming_transcript` attribute that never exists, hanging the script step forever.

## Fix

`VoiceAgentAdapter.call()` now runs best-effort runtime STT (`scenario.voice.stt.transcribe`, the already-configurable provider) over the merged agent turn when the transport supplied no transcript, before the assistant message is built. The simulator then receives the agent's words through the existing `[the agent said: ...]` reframe in `_strip_audio_content`. Adapters that already ship transcripts (realtime APIs) are unaffected, STT failures degrade to the previous behavior with a warning, and the recording segments get their transcript filled at the same time so the judge fallback has nothing left to redo.

`PipecatAgentAdapter` now advertises `streaming_transcripts=False`.

Related but distinct from #623 (that issue tracks agent-speaks-first and echo-safety porting for realtime adapters; this one is about the default drain path used by the websocket transports).

## What the blind simulator actually produced (real judged transcripts)

Both transcripts below are verbatim judge-span transcripts from runs that the framework treated as normal (audio payloads elided). The agent under test was responsive and in sync the whole time per its own logs; every derailment here is the simulator replying without having read a single agent turn.

Run 1. The simulator gives the membership number, ignores the follow-up question, asks the same question four times after it was answered, and the capture even shows a stacked stale-reply turn:

```text
assistant: Wegwise breakdown line, this is Emma. How can I help you today?
user: hi, i'm calling because my car won't start and i'm stuck in the jumbo car park in overvecht, utrecht
assistant: I'm sorry, that's really frustrating. What's your membership number?
user: okay, my membership number is 4-4-2-9-1-7
assistant: If you don't have it handy, that's okay. What's your full name?     <- number was just given
user: I'm at the Jumbo car park in Overvecht, Utrecht.                         <- "full name" answered with location
assistant: Thanks, what car is it?
user: it's a grey volkswagen polo plate xk-481-p and when i turn the key it just makes a single click and then nothing
assistant: Thanks. What happened with the car?
user: okay, thanks — i'm a bit stuck here, can you tell me roughly how long help will take?
assistant: Got it. What's your membership number? Thanks. What's your membership number? You're welcome. Take care.
user: okay, could you give me a rough estimate of how long help usually takes? i'm a bit stuck here.
assistant: I'm sorry, that's stressful. After dispatch the average wait is around 45 minutes and you'll get a live tracking link by SMS.
user: please tell me roughly how long help usually takes, i know you can't promise an exact time      <- answered one turn earlier
assistant: I'm sorry, that's stressful. The average is around 45 minutes, and after dispatch, you'll get a live tracking link by SMS.
user: that's fine, a rough average is fine — about thirty to forty-five minutes would be helpful
assistant: Usually around 45 minutes on average. After dispatch, you'll get a live tracking link by SMS.
user: okay, that's fine, please pass me to the dispatcher now
assistant: A rough average is around 45 minutes. Once it's dispatched, you'll get a live tracking link by SMS.
user: please connect me to the dispatcher now
assistant: I'll connect you with a human dispatcher now. They'll help you right away.
```

Run 2. Subtler and arguably worse because it PASSED: asked for the membership number, the simulator replies by confirming its own previous sentence and asking whether the agent needs the number it was just asked for:

```text
assistant: Wegweiss Breakdown Line. This is Emma. How can I help you today?
user: Hi, I'm at the Jumbo car park in Overvecht in Utrecht and my car won't start — it just makes a single click when I turn the key.
assistant: Got it. What's your membership number?
user: Yes, that's right — it just makes a single click. Do you need my membership number and the car details?   <- replying blind
assistant: Yes, I do. What's your membership number?
user: Yes — my membership number is 4-4-2-9-1-7. I'm parked at the Jumbo car park in Overvecht, Utrecht; it's a grey Volkswagen Polo, plate XK-481-P, and when I turn the key it just makes a single click and nothing else.
assistant: I'm connecting you with my colleague who will dispatch the patrol. I've noted member 4421917, Jumbo car park Overvecht in Utrecht, grey Volkswagen Polo XK481P. Single click when you turn the key.
```

In the simulator's own prompt, every one of those assistant turns was the literal string `[audio message]`.

rogeriochaves:
## Why

Closes #664

On audio-only transports (Pipecat, Twilio Media Streams) the voice user simulator — a text-only LLM — saw every agent turn as \`[audio message]\`. It could not read what the agent said, so both sides talked past each other: the simulator echoed old lines, re-asked answered questions, and the judge still passed because it transcribes the recording post-hoc and sees coherent audio regardless. This adds a **best-effort runtime STT step** after each agent drain so the simulator receives actual words.

## What changed

- **\`_ensure_transcript(merged)\` in \`VoiceAgentAdapter\`** — runs STT on the merged chunk if no transcript is already present (realtime APIs that stream text are a no-op); STT failure is logged and the audio-only chunk is returned unchanged, same contract as \`transcribe_segments\`.
- **\`mark_agent_end()\` called before STT, not after** — agent.end and the \`agent_stop_speaking\` event reflect when audio stopped flowing, not when the potentially-slow transcription returned. Avoids inflated latency numbers.
- **\`dataclasses.replace(merged, transcript=text)\`** — preserves \`start_time\`/\`end_time\` (and any future fields) instead of reconstructing \`AudioChunk\` from scratch.

## Test plan

\`\`\`
cd python && python -m pytest tests/voice/ -x -q
\`\`\`

Four tests in \`tests/voice/test_adapter_base.py\` cover the STT path:

- **\`test_default_call_transcribes_transcriptless_agent_audio\`** — adapter returns audio-only chunk → STT fires → result message carries real text.
- **\`test_default_call_skips_stt_when_adapter_ships_transcript\`** — adapter returns chunk with transcript already set → STT skipped (call count = 0).
- **\`test_default_call_returns_audio_only_when_stt_fails\`** — STT raises → warning logged → audio-only chunk returned unchanged.
- **\`test_default_call_returns_audio_only_when_stt_returns_empty\`** — STT returns empty string → \`if not text\` guard fires → audio-only chunk returned unchanged.

The \`mark_agent_end()\` timing claim is structural: drain returns and marks \`_agent_end\` before \`_ensure_transcript\` is called (lines 219–222 of \`adapter.py\`); a regression that swapped those lines would inflate latency measurements. The ordering is enforced at the call site rather than by a dedicated timing test.

Live run at HEAD \`fb737d9c\` against a real PCM16/24 kHz recording from \`outputs/recordings/pipecat_scenario/full.wav\` using the real OpenAI STT provider:

\`\`\`
=================================================================
DEMO: _ensure_transcript() — real OpenAI STT on agent audio
=================================================================

Loading audio from: full.wav
  user audio:  72000 bytes (1.5s)
  agent audio: 240000 bytes (5.0s)

Calling adapter.call()  (this triggers _ensure_transcript + OpenAI STT)
-----------------------------------------------------------------
  [adapter] emitting agent audio: 240000 bytes (5.0s), transcript=None  <-- audio-only transport
-----------------------------------------------------------------

RESULT (what the voice user simulator receives as 'assistant' message):

  role: assistant
  content parts: 2 (1 text, 1 audio)

  >>> TEXT CONTENT (user simulator reads this):
      "Thank you for calling. How can I help you today?"

SUCCESS: agent turn transcript populated via runtime STT
The voice user simulator receives actual words,
NOT '[audio message]' — Issue #664 is fixed.
\`\`\`

The adapter had \`streaming_transcripts=False\` and returned \`AudioChunk(data=..., transcript=None)\`. After \`call()\`, the assistant message contains a \`text\` content part with the real STT transcript — exactly what the voice user simulator reads from conversation history.

## Anything surprising?

Adapters that already populate \`chunk.transcript\` (OpenAI Realtime, ElevenLabs Conversational AI) skip the STT call entirely — zero behavioral change for them. The STT path is guarded by \`if not merged.data or merged.transcript: return merged\`.


python/scenario/voice/adapter.py
```diff
@@ -215,9 +215,48 @@ async def call(self, input: AgentInput) -> AgentReturnTypes:
...
+        except Exception:
+            logger.warning(
+                "voice: agent-turn STT failed; the user simulator will see "
+                "'[audio message]' instead of this turn's words",
+                exc_info=True,
+            )
+            return merged
+        if not text:
+            return merged
+        return AudioChunk(data=merged.data, transcript=text)
```
drewdrewthis: Reconstructing the chunk drops `start_time`/`end_time` when the drain returned a single chunk that carried them (`_merge_chunks` returns `chunks[0]` as-is, so those fields can survive to this point). `dataclasses.replace(merged, transcript=text)` preserves all fields and stays correct if `AudioChunk` grows new ones.

drewdrewthis (commented):
Reviewed both passes (line-by-line + structural). Fix is at the right layer: default `call()` path, no-op for adapters that ship transcripts, graceful STT-failure degradation, and the `mark_agent_end`-before-STT ordering keeps segment timing honest. Pipecat capability flip matches the wire reality (Twilio Media Streams = audio frames only) and `TwilioAgentAdapter` precedent. One small robustness nit inline.

python/scenario/voice/adapter.py
drewdrewthis: Good catch — fixed with `dataclasses.replace(merged, transcript=text)` to preserve `start_time`/`end_time` and any future fields.

python/scenario/voice/adapter.py
```diff
@@ -215,9 +216,48 @@ async def call(self, input: AgentInput) -> AgentReturnTypes:
...
+            return merged
+        try:
+            from .stt import transcribe
+
+            text = await transcribe(merged)
+        except Exception:
+            logger.warning(
+                "voice: agent-turn STT failed; the user simulator will see "
+                "'[audio message]' instead of this turn's words",
+                exc_info=True,
```
drewdrewthis: **[security] `exc_info=True` at WARNING level may expose voice-conversation PII in exception chains.**

If the STT provider fails after receiving a response (e.g., a JSON decode error that echoes partial transcription text), `exc_info=True` embeds the full exception message and cause chain in the WARNING log record. Voice conversations may contain membership numbers, names, and other caller PII (as shown in the issue #664 example transcripts). Log sinks that are not explicitly scrubbed (CI, observability pipelines) would receive that content at WARNING level.

**Fix:** Drop `exc_info=True` from this `warning(...)` call. If trace detail is needed for debugging, add a separate `logger.debug("...", exc_info=True)` follow-up — DEBUG is not consumed by most production log aggregators and is opt-in.

drewdrewthis (commented):
Reviewing diff at f5f68b66

python/tests/voice/test_adapter_base.py
```diff
@@ -71,3 +71,125 @@ def test_capabilities_matrix_is_published():
...
+# _ensure_transcript step, the assistant message went into history without
+# a text part, the text-only user simulator rendered it as the
+# "[audio message]" placeholder, and the simulator replied blind to every
+# agent turn — while the judge, transcribing the recording post-hoc, still
+# showed a coherent-looking transcript.
+
+import asyncio
+from typing import Any, Optional, cast
+
+from scenario.voice import STTProvider, get_stt_provider, set_stt_provider
```
drewdrewthis: **[hygiene][test] Mid-file imports break the file's established layout.**

`import asyncio`, `from typing import Any, Optional, cast`, and `from scenario.voice import STTProvider, get_stt_provider, set_stt_provider` appear at lines 88–91, after three existing test functions. This violates PEP 8 E402 and the layout of this file and all other test files in the suite. A reader scanning the file hits the first batch of tests, assumes the file is done, and never sees the 120 lines below — a documented risk in this codebase.

**Fix:** Move all three import statements to the top of the file with the existing imports.

drewdrewthis:
## Review verdict: READY

All three blocking items from the NOT-READY pass are fixed. Four additional issues surfaced and addressed in this pass. Two decide items remain non-blocking.

---

### Fixes since NOT-READY (commits `083c54cf`→`fb737d9c`)

| Fix | Commit |
|-----|--------|
| **[security]** Dropped `exc_info=True` from `logger.warning` in `_ensure_transcript`; trace detail moved to `logger.debug` | `083c54cf` |
| **[hygiene][test]** Moved mid-file imports (`asyncio`, `from typing import ...`, `from scenario.voice import STTProvider...`) to top of `test_adapter_base.py` | `083c54cf` |
| **[proof]** Removed citation of non-existent `test_adapter.py::test_ensure_transcript_called_before_record_agent`; replaced with four real tests | PR body update |
| **[hygiene]** Moved lazy `from .stt import transcribe` inside try-block to module-level imports in `adapter.py` | `fb737d9c` |
| **[test]** Added `_EmptySTT` + `test_default_call_returns_audio_only_when_stt_returns_empty` covering the `if not text: return merged` branch | `fb737d9c` |
| **[proof]** Re-verified live demo at HEAD `fb737d9c`; PR body updated with fresh run output | PR body update |

---

### Non-blocking (Decide)

**[principles][fowler]** `merged.transcript` guard in `_ensure_transcript` is redundant with `stt.transcribe()`'s own early-exit — two layers own the same invariant. Options: (a) drop from `_ensure_transcript` and rely on `transcribe()`; (b) drop from `stt.transcribe()` and let callers own it. Either is fine. Current state (both guard) is defensively correct if verbose.

**[uncle-bob][metz-beck]** `except Exception` swallows `ImportError` alongside transient network failures, making a misconfigured install look like a flaky STT call (silent per-turn degradation at WARNING, not startup failure). Consider narrowing to the exception families STT actually raises, or at minimum letting `ImportError` propagate.

**[test]** Three tests mutate the global `_provider` singleton via `set_stt_provider` without a shared fixture, creating a race if tests ever run under `pytest-xdist -n auto`. Not a problem for the current sequential suite; extract a scoped fixture if parallelism is added.

**[hygiene]** The `record_user` fallback change (`max(0.0, end - chunk.duration_seconds)` → `end`) is correct but bundled into a PR whose stated purpose is agent-turn STT — a different reason to change. The deleted comment explained who the fallback protected; that rationale is now undocumented. A follow-up docstring note on `record_user` would restore the intent.

---

### New Issues (out of scope for this PR)

- `_merge_chunks` constructs `AudioChunk(data=..., transcript=...)` without `start_time`/`end_time`, so any turn with >1 audio chunk loses those fields before `_ensure_transcript` runs. Pre-existing, not introduced here.
- `streaming_transcripts` in `AdapterCapabilities` means "emits incremental deltas mid-utterance," not "has a final turn transcript" — the docstring should clarify this to prevent future attempts to gate `_ensure_transcript` on that capability (which would re-introduce the adapter-hang bug pipecat.py:58-64 documents).

---

### Live demo at HEAD `fb737d9c9c6f82db29444e62992133c59f4a2df5`

```
[adapter] emitting agent audio: 240000 bytes (5.0s), transcript=None  <-- audio-only transport

RESULT (what the voice user simulator receives as 'assistant' message):
  role: assistant, content parts: 2 (1 text, 1 audio)

  >>> TEXT CONTENT (user simulator reads this):
      "Thank you for calling. How can I help you today?"

SUCCESS: agent turn transcript populated via runtime STT
```

python/tests/voice/test_adapter_base.py
drewdrewthis: Fixed in commit `083c54cf`: all three imports (`import asyncio`, `from typing import Any, Optional, cast`, `from scenario.voice import STTProvider, get_stt_provider, set_stt_provider`) moved to the top of the file and consolidated into the existing `from scenario.voice import (...)` block.

python/scenario/voice/adapter.py
drewdrewthis: Fixed in commit `083c54cf`: dropped `exc_info=True` from `logger.warning`; moved to `logger.debug("voice: agent-turn STT failure detail", exc_info=True)` so exception chains stay out of WARNING-level sinks.

drewdrewthis (commented):
lgtm

drewdrewthis (approved):
lgtm

PR #649, opened 2026-06-10
fix(voice/#493): keepalive-aware recv_audio — tolerate silent-but-pinging stretches

drewdrewthis:
## Why
Closes #493. `ElevenLabsAgentAdapter.recv_audio()` aborted with a bare `TimeoutError` when the hosted agent went briefly silent (~30s of keep-alive pings, no audio frames), **even though the WebSocket was healthy** — a flake-class, retry-recoverable failure that surfaces when the agent pauses for tool calls / RAG / MCP roundtrips.

## What changed
`recv_audio` is now **keepalive-aware**: the timeout deadline resets on **every received message** (ping or audio), so it measures *time since last liveness signal*, not *time since start*. Silent-but-pinging stretches stay alive; only a **truly dead socket** (no messages for `timeout`) still raises. One-line change in `elevenlabs.py`.

## Test plan
`python/tests/voice/test_elevenlabs_recv_keepalive.py`: (1) silent-but-pinging-past-timeout-then-audio → returns the `AudioChunk` (load-bearing red→green, RED on `main`); (2) truly-dead socket → still raises (guard). Validated locally: both green; `tests/voice/test_adapters.py` **62 passed** (no EL regression); `uv run pyright .` **0 errors**.

## Anything surprising?
Tolerates silence-with-pings but does **not** bound a pathological *forever-pinging, never-audio* agent (`recv_audio` would wait indefinitely, bounded only by the higher-level scenario/turn timeout). Longer-term asks from the issue — a **configurable** recv timeout and a **typed error** distinguishing "agent silent" from "socket dead" — are deferred follow-ups.

## How I can prove I was successful
No playable artifact — async timeout-logic fix; proof is the red→green keepalive test + the dead-socket guard.

PR #668, opened 2026-06-15
fix(voice/#661): port sliding-idle-deadline to TypeScript receiveAudio

### LINKED ISSUE #661: Hosted elevenLabsAgent documented single-exchange happy path is flaky — intermittent receiveAudio timeout on the reply leg
## For Humans

**TypeScript `receiveAudio` uses a fixed timer that ignores ElevenLabs keep-alive pings — the same bug the Python adapter had before PR #649.** ElevenLabs ConvAI sends pings while processing user audio; in Python these reset the 30s idle deadline; in TypeScript they do not, so when the server takes >30s wall-clock to reply the TS adapter times out even though the socket is alive and pinging.

**Ask:** Port the Python sliding-idle-deadline pattern to the TypeScript `ElevenLabsAgentAdapter.receiveAudio`.

- PR #649 (commit `1e6e1f3`) fixed Python only — TypeScript excluded
- Fixed in Python: `deadline` resets after every WS frame; in TypeScript: `setTimeout` fires at fixed 30s, never reset
- Keyless unit test already exists as proof: `test_elevenlabs_recv_keepalive.py` — port to TS vitest

_Status: investigation complete, plan pending. Full spec below._

### Symptoms / context

On published `@langwatch/scenario@0.4.13`, the greeting-led single-exchange hosted ElevenLabs demo (`agent()→user()→agent()→judge()`) intermittently fails with:

```
ElevenLabsAgentAdapter: receiveAudio timed out. Hosted ElevenLabs ConvAI is server-VAD-driven ...
```

at the **second `agent()` call** (the reply leg). Run 1 fails (~40s); immediate retry (Run 2) passes (~30s). The adapter's `responseTimeout` default is 30s.

### Root cause / key findings

**PR #649 (commit `1e6e1f3`) fixed the Python `recv_audio` with a sliding idle deadline but did not update the TypeScript port.**

- **Python `recv_audio`** (`python/scenario/voice/adapters/elevenlabs.py:225–235`): after every received WS frame — audio, ping, transcript, any event — `deadline` is reset: `deadline = asyncio.get_running_loop().time() + timeout`. Timeout only fires if the socket goes **silent for `timeout` seconds with no messages at all**.

- **TypeScript `receiveAudio`** (`javascript/src/voice/adapters/elevenlabs.ts:264–288`): a `setTimeout(..., timeout * 1000)` is set once when the method is called. Ping frames arriving in `onMessage` send pongs but never call `clearTimeout`/`setTimeout` to reset the timer. Timeout fires at `now + 30s wall-clock` regardless of received pings.

ElevenLabs ConvAI sends keep-alive `ping` frames every few seconds while processing. When the server takes 31–40s to reply (legitimate — tool calls, slow VAD, load), the Python adapter survives (pings kept resetting the 30s window); the TypeScript adapter times out (fixed 30s fired before audio arrived).

### Strategies considered

| Strategy | Verdict |
|---|---|
| **Port sliding idle deadline to TS** (timerResetters array in adapter, reset on every `onMessage`) | ✅ Recommended — exact Python parity, testable keyless |
| Increase default `responseTimeout` to 60s | ⚠️ Mitigates, does not fix; a slow server still breaks at 61s |
| Restructure `receiveAudio` into direct WS read loop | ⚠️ Larger refactor; waiter-queue model can stay with timerResetters |
| Retry logic at scenario executor level | ❌ Wrong layer |

1. **`timerResetters: Array<() => void>`** — add to the adapter class (alongside existing `waiters` and `audioQueue`).
2. **`onMessage` reset fan-out** — at the top of `onMessage`, before the switch on `etype`, iterate `this.timerResetters` and call each resetter.
3. **`receiveAudio` registers a resetter** — the `setTimeout` callback is extracted into `startTimer()`; a `resetTimer` closure calls `clearTimeout(timer); timer = startTimer();`. Register `resetTimer` into `timerResetters` when the waiter is pushed; remove it when the waiter fires or times out.
4. **`drainPendingWaiters`** — also splice all `timerResetters` (prevents timer leaks on socket close/error).
5. **Keyless unit test** — port `python/tests/voice/test_elevenlabs_recv_keepalive.py` to `javascript/src/voice/__tests__/elevenlabs-recv-keepalive.test.ts`. Use the existing `webSocketFactory` injection seam on `ElevenLabsAgentAdapterOptions`.

### Caveats

- The fix must NOT prevent the dead-socket timeout (AC-KA2): resetting only fires on received messages, so a truly silent socket still trips the timer.
- Confirm `drainPendingWaiters` cleans up `timerResetters` so no leaked `setTimeout` fires after disconnect.
- The `tail-silence` logic (SILENCE_TAIL_BYTES = 16000) is a separate mechanism; do not change it.

### `repro-attempt:`

**BLOCKER inputs:** No `ELEVENLABS_API_KEY` or `ELEVENLABS_AGENT_ID` in any `.env` file or `~/.claude/.secrets/` — live E2E cannot be run by this agent. The mechanistic proof is available keyless via the unit test (AC-KA1): a mock WS injected via `webSocketFactory` that sends 8 pings at 80ms intervals (640ms > 300ms timeout) followed by one audio frame should cause `receiveAudio(0.30)` to throw on current code and pass on fixed code.

The exact repro command for whoever holds the keys:
```
ELEVENLABS_API_KEY=... ELEVENLABS_AGENT_ID=... OPENAI_API_KEY=... \
  npx vitest run tests/voice/elevenlabs-hosted.test.ts --reporter=verbose
```
Run twice in immediate succession and compare — first run intermittently fails at ~40s.

## Acceptance Criteria

- [ ] **AC-KA1 (keepalive fix — RED on current code)**: `receiveAudio` tolerates a silent-but-pinging stretch longer than `timeout`. Evidence: keyless unit test — `receiveAudio(0.30)` is awaited while, on a connected `FakeWebSocket` (injected via `webSocketFactory`), 8 `ping` frames are emitted as `message` events at ~80ms real gaps (total ~640ms > 300ms timeout), then one `audio` frame is emitted; the promise resolves to an `AudioChunk` with the expected PCM data. Test includes: (a) a static timing-invariant assertion `expect(NUM_PINGS * PING_GAP).toBeGreaterThan(TIMEOUT)` so future constant changes cannot make it pass trivially; (b) elapsed-time ≥ TIMEOUT assert (resolution must occur AFTER the fixed timer would have fired). RED on current code (single `setTimeout` fires at 300ms before audio arrives); GREEN after `timerResetters` pattern. Shape: passing `vitest run` output.

- [ ] **AC-KA2 (dead-socket regression guard — partly green-before-work, guards fix doesn't break the wall)**: After the keepalive change, a socket that emits **zero messages** still times out. Evidence: keyless unit test — `receiveAudio(0.30)` on a connected `FakeWebSocket` with no `message` events emitted rejects with an `Error` matching `/timed out/`, resolving in ≥300ms and ≤450ms. This verifies the resetter fires only on received frames (not unconditionally), so a truly silent socket still trips the deadline. Note: the pre-existing test at `src/voice/adapters/__tests__/elevenlabs.test.ts:675` ("receiveAudio rejects with timeout when no audio arrives in time") already covers this case — AC-KA2 is explicitly a no-regression guard after the fix, not new verification. Shape: passing `vitest run` output.

- [ ] **AC-KA3 (no regression — correct scope)**: All existing adapter tests pass after the change. Evidence: `npm test` (= `vitest run`, `include: ['src/**/*.test.ts']`) exits 0 with 0 failures — this scope MUST include `src/voice/adapters/__tests__/elevenlabs.test.ts` (35 it-blocks covering `onMessage` branches, socket-error/close `drainPendingWaiters` drain, existing silent-timeout; these are the regression neighbors of the changed code). Shape: vitest summary line showing total test count and 0 failures.

- [ ] **AC-KA4 (informative, gated, non-blocking — waiver path is the real one)**: Not required for merge. If `ELEVENLABS_API_KEY` + `ELEVENLABS_AGENT_ID` are present, 2 consecutive passes of `elevenlabs-hosted.test.ts`. Otherwise this AC is explicitly waived (keys unavailable per `repro-attempt:` block) and AC-KA1 is the merge-gating mechanistic proof — record the waiver in the prove-it Evidence block. Shape: vitest summary or explicit waiver note.

- [ ] **AC-KA5 (cleanup — no timer leaks on socket close/error)**: A pending `receiveAudio` whose socket emits `close` (or `error`) leaves no surviving timer and the `timerResetters` list is cleared. Evidence: keyless unit test — start `receiveAudio(5)`, emit `close` on the `FakeWebSocket`; the promise resolves to the empty drain `AudioChunk` (no throw), AND no timer fires afterward (assert `vi.getTimerCount() === 0` under fake timers OR assert no late rejection fires within a real 100ms window after the drain). Neighbors: existing socket-error test at `src/voice/adapters/__tests__/elevenlabs.test.ts` (the same `drainPendingWaiters` call path). Shape: passing `vitest run` output.

- **AC-KA2**: Dead-socket guard — the sliding deadline does not produce infinite wait on a truly silent socket.
- **AC-KA3**: All 35 adapter unit tests pass — no regression to `onMessage` branch handling, existing silent-timeout behavior, or `sendAudio` path.
- **AC-KA5**: Timer-leak failure mode on socket close/error — `drainPendingWaiters` cleans `timerResetters`; no open handle survives disconnect.
- No migration/rollback surface — adapter is in-process, pure JavaScript, no schema or data changes.
- **Intentional non-coverage (tracked separately)**: A hosted agent that pings forever but never sends audio makes `receiveAudio` wait indefinitely (correct Python parity — "no total-duration ceiling" per Python adapter docstring); a caller-side absolute backstop is a separate follow-up tracked for both Python and TypeScript.

## Plan

**Approach:** Add a `timerResetters` array to `ElevenLabsAgentAdapter` and fan-out through `onMessage` — every inbound WS message resets all active `receiveAudio` timers, matching the Python sliding-idle-deadline contract. Write a failing keepalive unit test first; make it pass with the `timerResetters` change; then add the `drainPendingWaiters` cleanup AC-KA5 guard.

**Sequencing:**

1. **Write failing keepalive test** (`javascript/src/voice/adapters/__tests__/elevenlabs.test.ts` or a new `elevenlabs-recv-keepalive.test.ts` colocated in `src/voice/adapters/__tests__/`): mock WS via `webSocketFactory` sends 8 pings at ~80ms intervals then one audio frame; asserts `receiveAudio(0.30)` resolves without throwing. RED on current code. Completion: `vitest run` shows this test failing with timeout error.

2. **Port sliding-idle-deadline to TypeScript** (`javascript/src/voice/adapters/elevenlabs.ts`):
   - Add `private readonly timerResetters: Array<() => void> = []` field
   - In `onMessage` top: `for (const r of this.timerResetters) r()` (before `etype` switch)
   - In `receiveAudio`: extract `startTimer()` helper; a `resetTimer` closure `clearTimeout`s and re-arms; register `resetTimer` into `timerResetters` alongside the waiter; remove from both arrays on resolve/timeout
   - In `drainPendingWaiters`: also splice `this.timerResetters`
   Completion: AC-KA1 keepalive test turns GREEN; AC-KA2 (silent-socket) still passes; AC-KA5 (cleanup) can be written.

3. **Write cleanup test (AC-KA5)**: emit `close` on a FakeWebSocket with a pending `receiveAudio(5)` — assert empty-chunk resolution and no surviving timer. Completion: AC-KA5 test passes.

4. **Run full suite (AC-KA3)**: `npm test` in `javascript/` exits 0. Completion: vitest summary 0 failures.

**Key decisions:**

- **`timerResetters` alongside `waiters`** rather than restructuring into a direct WS read loop: preserves the push-waiter-queue architecture the existing 35 adapter tests exercise; the Python-loop alternative would require rewriting the entire adapter.
- **New test colocated in `src/voice/adapters/__tests__/`**, not `src/voice/__tests__/`: matches where the adapter unit suite lives; AC-KA3's corrected scope covers both directories.
- **`drainPendingWaiters` cleans `timerResetters`**: same cleanup pattern as `waiters`; prevents open timer handles after socket close/error.

**Risks / open questions:**

- Fake timer (`vi.useFakeTimers`) vs real timers for AC-KA1: the Python keepalive test uses real `asyncio.sleep` gaps; the TS unit test should use real time as well (or fake timers with careful manual advancing), since the goal is to prove the actual per-message reset fires during genuine elapsed time. Implementer chooses; either is acceptable if the timing-invariant assert is present.
- The `timerResetters` array is iterated inside `onMessage` which fires on every inbound frame including audio. This means an audio frame also triggers all resetters before the waiter resolves — harmless (resetter just clears + re-arms a timer that's about to be cancelled anyway) but worth noting.

**Out of scope:**

- Caller-side absolute backstop for a pinging-but-never-audio wedged agent (tracked as follow-up for both Python and TypeScript)
- Any change to `SILENCE_TAIL_BYTES` or the user audio send path
- Any change to the Python adapter (already fixed in PR #649)
- `AC-KA4` live E2E (keys unavailable in this environment; waived per repro-attempt block)

## Evidence

_Filled at prove-it time._

</details>

drewdrewthis:
## Why

**ElevenLabs ConvAI sends keep-alive pings while processing audio.** The Python `recv_audio` already handles this — its `deadline` resets on every inbound WS frame (fixed in PR #649). The TypeScript `receiveAudio` was never ported: it fires a fixed `setTimeout` 30 s after the call begins, ignoring pings entirely. Slow-but-healthy servers trip the timeout when processing exceeds 30 s wall-clock on the reply leg.

Closes #661

## What changed

- **`timerResetters` fan-out in `onMessage`** — every inbound WS frame (ping, transcript, audio, any event) calls `resetTimer()` for each active `receiveAudio`, sliding the idle deadline forward instead of firing at a fixed offset.
- **`resetTimer` callback injected into `timerResetters` array** — registered when `receiveAudio` begins, removed when the chunk arrives or the timeout fires, keeping the cleanup symmetric with `waiters`.
- **Regression test (AC-KA1)** — 8 pings × 80 ms (640 ms) then audio; fixed timer would fire at 300 ms timeout, port of `test_elevenlabs_recv_keepalive.py`. Red before fix, green after.
- **Drain test (AC-KA5)** — socket close cancels the deadline timer; asserts `vi.getTimerCount()` goes 1 → 0 on disconnect, proving no open-handle leak.

## How it works

`receiveAudio` now registers a `resetTimer` closure alongside its `waiter` in two parallel arrays on the adapter:

```
ElevenLabsAgentAdapter
  ├── waiters[]       ← resolves the promise when audio arrives
  └── timerResetters[] ← clears + re-arms the idle timer on each WS frame
```

`onMessage` fans out to all `timerResetters` on every frame before routing the event. When audio arrives the waiter fires, clearing the timer and splicing itself out of both arrays. If the socket closes, `drainPendingWaiters` rejects all waiters and clears their timers — same path that makes the AC-KA5 drain test pass.

## Test plan

```bash
cd javascript
pnpm vitest run src/voice/adapters/__tests__/elevenlabs.test.ts \
  --reporter=verbose 2>&1 | grep -E "KA[15]|keepalive|PASS|FAIL"
```

Key tests:
- `AC-KA1: receiveAudio tolerates a silent-but-pinging stretch longer than timeout` — 640 ms of pings, audio at 720 ms, timeout = 300 ms; must resolve, not reject
- `timing invariant: ping stretch exceeds timeout budget` — guard that the test scenario is actually RED on pre-fix code
- `AC-KA5: socket close drains cleanly, no surviving timer` — `vi.getTimerCount()` asserts 1 → 0 after the WS close event

Regression: the AC-KA1 test was written as a failing test first (commit `bb922ea`) and confirmed red against the pre-fix code.

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| KA1 | `receiveAudio` resets deadline on each ping; 8 × 80 ms pings + audio at 720 ms resolves past the 300 ms timeout | ✅ verified-in-tree |
| KA2 | Truly silent socket still times out (regression guard) | ✅ verified-in-tree |
| KA3 | Full vitest run exits 0, 39 tests pass | ✅ verified-in-tree |
| KA4 | Live single-exchange passes twice (or explicitly waived) | ✅ waived — ELEVENLABS_API_KEY/AGENT_ID absent; AC-KA1 is merge gate per feature spec |
| KA5 | Socket close drains cleanly; vi.getTimerCount() 1 → 0 | ✅ verified-in-tree |

### Behavior demo — standalone script, no test runner

`javascript/demo-sliding-deadline.ts` imports the real `ElevenLabsAgentAdapter`, wires a fake WebSocket, and drives it with timed ping + audio events that mimic a live exchange. The script proves the observable behavior: 8 pings at 250 ms intervals (2000 ms total silent stretch, 4× the 500 ms timeout) keep the idle deadline alive, and audio arriving at ~2100 ms resolves cleanly with no premature timeout.

```
$ cd javascript && node_modules/.bin/tsx demo-sliding-deadline.ts

=== ElevenLabsAgentAdapter: sliding-idle-deadline demo ===

  receiveAudio timeout : 500ms
  ping interval        : 250ms
  total silent stretch : 2000ms  (8 pings)
  audio arrives at     : ~2100ms

  WITHOUT fix: timer fires at 500ms → TimeoutError
  WITH fix   : each ping resets deadline → resolves cleanly

+   49ms [adapter] connected

+   49ms [adapter] receiveAudio(0.5s) ← idle deadline armed
+  300ms [ws←server] ping #1  → deadline reset (+500ms from now)
+  300ms [ws→server] pong event_id=1
+  551ms [ws←server] ping #2  → deadline reset (+500ms from now)
+  551ms [ws→server] pong event_id=2
+  802ms [ws←server] ping #3  → deadline reset (+500ms from now)
+  802ms [ws→server] pong event_id=3
+ 1051ms [ws←server] ping #4  → deadline reset (+500ms from now)
+ 1051ms [ws→server] pong event_id=4
+ 1302ms [ws←server] ping #5  → deadline reset (+500ms from now)
+ 1302ms [ws→server] pong event_id=5
+ 1553ms [ws←server] ping #6  → deadline reset (+500ms from now)
+ 1553ms [ws→server] pong event_id=6
+ 1803ms [ws←server] ping #7  → deadline reset (+500ms from now)
+ 1803ms [ws→server] pong event_id=7
+ 2054ms [ws←server] ping #8  → deadline reset (+500ms from now)
+ 2054ms [ws→server] pong event_id=8
+ 2154ms [ws←server] audio event (100 bytes PCM16 silence)

+ 2155ms [adapter] receiveAudio RESOLVED — chunk.data.length=100 bytes

PASS: no premature timeout after 2000ms of server silence kept alive by pings.
```

**What this proves:** `receiveAudio` survived 2000 ms of silence (4× the 500 ms timeout) because each ping reset the idle deadline. The `[ws→server] pong` lines confirm `onMessage` fires on pings and the adapter replies correctly. Audio at 2154 ms resolved the promise — no `TimeoutError`.

**AC-KA4 live E2E waiver:** `ELEVENLABS_API_KEY` and `ELEVENLABS_AGENT_ID` confirmed absent this turn from `javascript/.env`, `scenario/javascript/.env`, `scenario/python/.env`, `~/.claude/.secrets/`, and shell env. Per `specs/elevenlabs-ts-keepalive.feature` lines 86–90, this AC is explicitly waivable when keys are unavailable; waiver recorded; AC-KA1 is the merge gate.

### DIFF
--- a/javascript/src/voice/adapters/__tests__/elevenlabs.test.ts
+++ b/javascript/src/voice/adapters/__tests__/elevenlabs.test.ts
@@ -772,4 +772,32 @@ describe("receiveAudio — keepalive-aware sliding idle deadline (#661)", () => 
       await adapter.disconnect();
     },
   );
+
+  it("AC-KA5: socket close drains cleanly with no surviving timer", { timeout: 5000 }, async () => {
+    // Use fake timers so we can inspect surviving timer count after drain
+    vi.useFakeTimers();
+    try {
+      const { adapter, socket } = await makeConnected();
+
+      // Start a receiveAudio with a long timeout (5s) — we'll close before it fires
+      const receivePromise = adapter.receiveAudio(5);
+
+      // One timer should be active (the receiveAudio deadline)
+      expect(vi.getTimerCount()).toBe(1);
+
+      // Emit close — triggers drainPendingWaiters
+      socket.emit("close");
+
+      // The promise resolves to the empty drain chunk
+      const result = await receivePromise;
+      expect(result.data.length).toBe(0); // empty chunk = drained, not a real audio payload
+
+      // No surviving timers — the waiter cancelled the timer on drain
+      expect(vi.getTimerCount()).toBe(0);
+
+      await adapter.disconnect();
+    } finally {
+      vi.useRealTimers();
+    }
+  });
 });
--- a/javascript/src/voice/adapters/__tests__/elevenlabs.test.ts
+++ b/javascript/src/voice/adapters/__tests__/elevenlabs.test.ts
@@ -678,3 +678,98 @@ describe("ElevenLabsAgentAdapter wire-protocol (onMessage branches)", () => {
     await adapter.disconnect();
   });
 });
+
+// ---------------------------------------------------------------------------
+// Issue #661 — keepalive-aware sliding idle deadline
+// ---------------------------------------------------------------------------
+describe("receiveAudio — keepalive-aware sliding idle deadline (#661)", () => {
+  // Timing constants
+  const TIMEOUT_S = 0.30;
+  const NUM_PINGS = 8;
+  const PING_GAP_MS = 80;
+  const AUDIO_GAP_MS = 80;
+
+  // PCM payload: 8 bytes of valid PCM16 (even-byte count)
+  const pcmB64 = Buffer.from("\x12\x34".repeat(4)).toString("base64");
+
+  // Helper: build adapter + connected FakeWebSocket
+  async function makeConnected(): Promise<{
+    adapter: ElevenLabsAgentAdapter;
+    socket: FakeWebSocket;
+  }> {
+    const fake = makeFakeSocketFactory();
+    const adapter = new ElevenLabsAgentAdapter({
+      agentId: "agt-keepalive",
+      apiKey: "sk-keepalive",
+      webSocketFactory: fake.factory,
+    });
+    await adapter.connect();
+    return { adapter, socket: fake.socket.current! };
+  }
+
+  it("timing invariant: ping stretch exceeds timeout budget", () => {
+    // If this fails, the test scenario is misconfigured — it must be RED on pre-fix code.
+    expect(NUM_PINGS * PING_GAP_MS).toBeGreaterThan(TIMEOUT_S * 1000);
+  });
+
+  it(
+    "AC-KA1: receiveAudio tolerates a silent-but-pinging stretch longer than timeout",
+    { timeout: 10000 },
+    async () => {
+      const { adapter, socket } = await makeConnected();
+
+      // Schedule 8 ping frames at 80 ms real intervals, then one audio frame.
+      // Total ping stretch = 640 ms; audio arrives at 720 ms.
+      // Fixed-timer code fires at 300 ms → RED.
+      // Sliding-deadline code resets on each ping → GREEN.
+      for (let i = 0; i < NUM_PINGS; i++) {
+        setTimeout(() => {
+          socket.emit(
+            "message",
+            Buffer.from(
+              JSON.stringify({
+                type: "ping",
+                ping_event: { event_id: i, ping_ms: 5 },
+              }),
+            ),
+          );
+        }, (i + 1) * PING_GAP_MS);
+      }
+      setTimeout(() => {
+        socket.emit(
+          "message",
+          Buffer.from(
+            JSON.stringify({
+              type: "audio",
+              audio_event: { audio_base_64: pcmB64 },
+            }),
+          ),
+        );
+      }, NUM_PINGS * PING_GAP_MS + AUDIO_GAP_MS);
+
+      const start = performance.now();
+      const result = await adapter.receiveAudio(TIMEOUT_S);
+      const elapsed = performance.now() - start;
+
+      // Resolved with a valid AudioChunk containing the PCM bytes
+      expect(result).toBeInstanceOf(AudioChunk);
+      expect(result.data.length).toBeGreaterThan(0);
+
+      // Proved we waited past what a fixed timer would have allowed
+      expect(elapsed).toBeGreaterThanOrEqual(TIMEOUT_S * 1000);
+
+      await adapter.disconnect();
+    },
+  );
+
+  it(
+    "AC-KA2 regression guard: truly silent socket still times out",
+    { timeout: 10000 },
+    async () => {
+      // No messages emitted — the idle deadline must still fire.
+      const { adapter } = await makeConnected();
+      await expect(adapter.receiveAudio(TIMEOUT_S)).rejects.toThrow(/timed out/);
+      await adapter.disconnect();
+    },
+  );
+});
--- a/javascript/src/voice/adapters/__tests__/elevenlabs.test.ts
+++ b/javascript/src/voice/adapters/__tests__/elevenlabs.test.ts
@@ -776,8 +776,11 @@ describe("receiveAudio — keepalive-aware sliding idle deadline (#661)", () => 
   it("AC-KA5: socket close drains cleanly with no surviving timer", { timeout: 5000 }, async () => {
     // Use fake timers so we can inspect surviving timer count after drain
     vi.useFakeTimers();
+    let adapter: ElevenLabsAgentAdapter | undefined;
     try {
-      const { adapter, socket } = await makeConnected();
+      const connected = await makeConnected();
+      adapter = connected.adapter;
+      const socket = connected.socket;
 
       // Start a receiveAudio with a long timeout (5s) — we'll close before it fires
       const receivePromise = adapter.receiveAudio(5);
@@ -794,9 +797,8 @@ describe("receiveAudio — keepalive-aware sliding idle deadline (#661)", () => 
 
       // No surviving timers — the waiter cancelled the timer on drain
       expect(vi.getTimerCount()).toBe(0);
-
-      await adapter.disconnect();
     } finally {
+      await adapter?.disconnect();
       vi.useRealTimers();
     }
   });
--- a/javascript/demo-sliding-deadline.ts
+++ b/javascript/demo-sliding-deadline.ts
@@ -1,0 +1,145 @@
+/**
+ * demo-sliding-deadline.ts
+ *
+ * Standalone behavior demo for the sliding-idle-deadline fix (issue #661 / #668).
+ *
+ * Scenario:
+ *   receiveAudio timeout = 500ms (0.5 s)
+ *   Pings arrive every 250ms — 8 pings = 2000ms total silent stretch
+ *   Without fix:  timer fires at 500ms → TimeoutError
+ *   With fix:     each ping resets the 500ms idle deadline → audio at ~2100ms resolves OK
+ *
+ * Run: node_modules/.bin/tsx demo-sliding-deadline.ts
+ */
+import { ElevenLabsAgentAdapter } from "./src/voice/adapters/elevenlabs.js";
+import { Buffer } from "node:buffer";
+import type { RawData } from "ws";
+
+// ── timing constants ──────────────────────────────────────────────────────────
+const TIMEOUT_S = 0.5;         // 500ms raw idle deadline
+const PING_INTERVAL_MS = 250;  // each ping resets the timer — buys another 500ms
+const N_PINGS = 8;             // 8 × 250ms = 2000ms total >> 500ms without fix
+const AUDIO_DELAY_MS = 100;    // audio arrives 100ms after last ping
+
+// ── helpers ───────────────────────────────────────────────────────────────────
+const t0 = process.hrtime.bigint();
+function elapsed(): string {
+  const ms = Number((process.hrtime.bigint() - t0) / 1_000_000n);
+  return `+${ms.toString().padStart(5)}ms`;
+}
+
+function pingEvent(id: number): Buffer {
+  return Buffer.from(JSON.stringify({ type: "ping", ping_event: { event_id: id } }));
+}
+
+function audioEvent(): Buffer {
+  const silence = new Uint8Array(100); // 100 bytes PCM16 silence (even, valid)
+  const b64 = Buffer.from(silence).toString("base64");
+  return Buffer.from(
+    JSON.stringify({ type: "audio", audio_event: { audio_base_64: b64 } }),
+  );
+}
+
+// ── fake WebSocket ─────────────────────────────────────────────────────────────
+function createFakeWS() {
+  const onceHandlers = new Map<string, (...args: unknown[]) => void>();
+  const onHandlers   = new Map<string, (...args: unknown[]) => void>();
+  let openFired = false;
+
+  const ws = {
+    readyState: 1 as number | undefined,
+    send: (data: string) => {
+      try {
+        const msg = JSON.parse(data) as Record<string, unknown>;
+        if (msg["type"] === "pong") {
+          console.log(`${elapsed()} [ws→server] pong event_id=${String(msg["event_id"])}`);
+        }
+      } catch { /* ignore */ }
+    },
+    close: () => {},
+    once: (event: string, listener: (...args: unknown[]) => void) => {
+      onceHandlers.set(event, listener);
+      if (event === "open" && !openFired) {
+        openFired = true;
+        setImmediate(() => {
+          const h = onceHandlers.get("open");
+          if (h) { onceHandlers.delete("open"); h(); }
+        });
+      }
+      return ws;
+    },
+    on: (event: string, listener: (...args: unknown[]) => void) => {
+      onHandlers.set(event, listener);
+      return ws;
+    },
+    removeAllListeners: () => { onceHandlers.clear(); onHandlers.clear(); },
+    inject: (data: Buffer) => { onHandlers.get("message")?.(data as unknown as RawData); },
+  };
+  return ws;
+}
+
+// ── main ───────────────────────────────────────────────────────────────────────
+async function main() {
+  console.log("=== ElevenLabsAgentAdapter: sliding-idle-deadline demo ===");
+  console.log();
+  console.log(`  receiveAudio timeout : ${TIMEOUT_S * 1000}ms`);
+  console.log(`  ping interval        : ${PING_INTERVAL_MS}ms`);
+  console.log(`  total silent stretch : ${N_PINGS * PING_INTERVAL_MS}ms  (${N_PINGS} pings)`);
+  console.log(`  audio arrives at     : ~${N_PINGS * PING_INTERVAL_MS + AUDIO_DELAY_MS}ms`);
+  console.log();
+  console.log(`  WITHOUT fix: timer fires at ${TIMEOUT_S * 1000}ms → TimeoutError`);
+  console.log(`  WITH fix   : each ping resets deadline → resolves cleanly`);
+  console.log();
+
+  let fakeWS!: ReturnType<typeof createFakeWS>;
+
+  const adapter = new ElevenLabsAgentAdapter({
+    agentId: "demo-agent-id",
+    apiKey: "demo-api-key",
+    webSocketFactory: (_url, _headers) => {
+      fakeWS = createFakeWS();
+      return fakeWS;
+    },
+  });
+
+  // 1. Connect
+  await adapter.connect();
+  console.log(`${elapsed()} [adapter] connected`);
+  console.log();
+
+  // 2. Arm receiveAudio — idle deadline starts NOW
+  console.log(`${elapsed()} [adapter] receiveAudio(${TIMEOUT_S}s) ← idle deadline armed`);
+  const receivePromise = adapter.receiveAudio(TIMEOUT_S);
+
+  // 3. Fire pings every PING_INTERVAL_MS — each resets the idle deadline
+  for (let i = 1; i <= N_PINGS; i++) {
+    await new Promise<void>((r) => setTimeout(r, PING_INTERVAL_MS));
+    console.log(
+      `${elapsed()} [ws←server] ping #${i}  → deadline reset (+${TIMEOUT_S * 1000}ms from now)`,
+    );
+    fakeWS.inject(pingEvent(i));
+  }
+
+  // 4. Audio arrives — resolves the promise
+  await new Promise<void>((r) => setTimeout(r, AUDIO_DELAY_MS));
+  console.log(`${elapsed()} [ws←server] audio event (100 bytes PCM16 silence)`);
+  fakeWS.inject(audioEvent());
+
+  const chunk = await receivePromise;
+
+  console.log();
+  console.log(
+    `${elapsed()} [adapter] receiveAudio RESOLVED — chunk.data.length=${chunk.data.length} bytes`,
+  );
+  console.log();
+  console.log("PASS: no premature timeout after " +
+    `${N_PINGS * PING_INTERVAL_MS}ms of server silence kept alive by pings.`);
+
+  await adapter.disconnect();
+}
+
+main().catch((err: unknown) => {
+  const msg = err instanceof Error ? err.message : String(err);
+  console.error(`\nFAILED: ${msg}`);
+  process.exit(1);
+});
--- a/javascript/package-lock.json
+++ b/javascript/package-lock.json
@@ -1,0 +1,15205 @@
+{
+  "name": "@langwatch/scenario",
+  "version": "0.4.13",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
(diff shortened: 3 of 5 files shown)


javascript/src/voice/adapters/__tests__/elevenlabs.test.ts
```diff
@@ -678,3 +678,126 @@ describe("ElevenLabsAgentAdapter wire-protocol (onMessage branches)", () => {
...
+      socket.emit("close");
+
+      // The promise resolves to the empty drain chunk
+      const result = await receivePromise;
+      expect(result.data.length).toBe(0); // empty chunk = drained, not a real audio payload
+
+      // No surviving timers — the waiter cancelled the timer on drain
+      expect(vi.getTimerCount()).toBe(0);
+
+      await adapter.disconnect();
```
drewdrewthis: **[test] Move `adapter.disconnect()` to the `finally` block**

If any assertion between line 777 and here throws (e.g. `vi.getTimerCount() !== 0`), `adapter.disconnect()` never runs. `vi.useRealTimers()` correctly lives in `finally`, but the adapter stays in a pseudo-connected state with a live FakeWebSocket reference. Move `await adapter.disconnect()` to the `finally` block alongside `vi.useRealTimers()` so it always fires.

```typescript
    } finally {
      await adapter.disconnect();
      vi.useRealTimers();
    }
```

drewdrewthis (commented):
Inline review by /review — see verdict comment for summary.

drewdrewthis:
## Review verdict: READY

**Reviewed at:** `addd55bf` · **Run:** /review (own-PR)

---

### Non-blocking concerns

> No Fix threads filed. The following are New Issue / Decide items — they do not gate merge.

**[principles][hygiene][uncle-bob][metz-beck][fowler] New Issue — Refactor `waiters[]` + `timerResetters[]` parallel arrays into a single `PendingReceive` object**

Five reviewers reached independent consensus that two positionally-coupled `Array<fn>` fields maintained via `indexOf/splice` in four sites is a data clump that should become `Array<{resolve, reset, cancel}>`. The code is functionally correct — `drainPendingWaiters` calls each `waiter` which splices both arrays in sync, and AC-KA5 proves `vi.getTimerCount() === 0` after drain. But the cleanup is indirect, the invariant "these two arrays must stay index-aligned" lives only in the programmer's head, and the `indexOf/splice` pattern appears four times. A `PendingReceive` object makes divergence structurally impossible. File as a follow-up issue; not blocking this fix.

**[uncle-bob] Decide — "Matches Python recv_audio sliding-idle-deadline" comment misleads**

The comment appears three times in the diff. Python `recv_audio` is pull-based (`await ws.recv()` in a while-loop, one local `deadline` variable). TypeScript is push-based (`onMessage` fan-out to N parallel resolver arrays). The *behavior* matches; the *architecture* diverges significantly. A maintainer using Python as the mental model would see the timer leak as invisible — it's structurally impossible in Python. Consider replacing with a comment describing the TS-specific ownership model rather than claiming structural parity.

**[principles][hygiene] Decide — `demo-sliding-deadline.ts` belongs in `examples/` not at package root**

The file imports from `./src/...`, won't be picked up by tsconfig `include` (which excludes root-level non-src files), but lives outside the conventional `examples/` directory and is not in `package.json` scripts. Low-impact; author's call whether to move it.

**[security]** No concerns. Fake credentials, zero PII, clean test fixtures.

---

### Review dimensions covered

| Reviewer | Finding |
|----------|---------|
| principles | New Issue: parallel arrays; Decide: demo placement, Python comment |
| hygiene | Confirms same |
| security | ✅ Clean |
| test | 39/39 pass at HEAD; `queueMicrotask` concern investigated — AC-KA5 passes in 2ms, concern invalid |
| proof | Marker refreshed to `addd55bf`; KA3 re-run at HEAD; all ACs verified or waived |
| uncle-bob | "Extract PendingReceive class" — agrees fix is correct, flags design for follow-up |
| metz-beck | "Not premature abstraction — under-abstraction." Confirms current reset-on-every-frame behavior is correct |
| fowler | "Data Clump, Shotgun Surgery, Primitive Obsession" — same refactor recommendation |

drewdrewthis:
## Verification Report

| KA1 | `receiveAudio` tolerates 8 pings × 80 ms (640 ms) before audio arrives; fixed 300 ms timer would fire first | AC-KA1 test ran in **721 ms** wall-clock (> 300 ms timeout) and resolved with AudioChunk — proves the deadline slid forward on each ping. Real timers used (no fake timer advancement). | PASS |
| KA2 | Truly silent socket still rejects with timeout (no-regression guard) | AC-KA2 test ran in **301 ms** and rejected with `/timed out/` — confirms the wall still fires on a dead socket | PASS |
| KA3 | Full vitest run exits 0, 0 failures; scope includes `elevenlabs.test.ts` 35+ it-blocks | Re-run at HEAD `addd55b` (this turn): `Test Files 1 passed (1) / Tests 39 passed (39)` — Start at 19:45:28; exit 0 | PASS |
| KA4 | Live single-exchange passes twice consecutively without `receiveAudio timed out` | **Credential check this turn:** `ELEVENLABS_API_KEY` and `ELEVENLABS_AGENT_ID` absent from all checked locations (`javascript/.env`, `scenario/javascript/.env`, `scenario/python/.env`, `~/.claude/.secrets/`, env). **Explicitly waivable per `specs/elevenlabs-ts-keepalive.feature` lines 86–90** — waiver recorded; AC-KA1 is the merge gate. | WAIVED |
| KA5 | Socket close before audio drains cleanly; `vi.getTimerCount()` drops 1 → 0 | AC-KA5 test: asserts `getTimerCount() === 1` before close, emits `close` event, asserts `getTimerCount() === 0` after drain. Ran in **3 ms**. | PASS |

**Verdict: 4/5 verified, KA4 explicitly waived (merge gate = KA1), no drift, no dropped ACs**

### Faithfulness
Diff touches `elevenlabs.ts` (timerResetters fan-out), `elevenlabs.test.ts`, `demo-sliding-deadline.ts` (proof script, non-production), and `specs/elevenlabs-ts-keepalive.feature`. The `demo-sliding-deadline.ts` commit (`addd55b`) is a proof artifact only — no AC-relevant code changed since `407589e`. Directly addresses the original problem: TypeScript `receiveAudio` firing a fixed `setTimeout` that ignores pings.

### Step 2.7 — Behavior proof (real run, inspected this turn)

AC-KA1 exercises the **exact failure mode** at real timing: 8 pings at real 80 ms intervals (640 ms total) then audio at 720 ms, against a 300 ms timeout. Without the fix, the promise would reject at 300 ms. The test measured **721 ms elapsed and resolved** — proving the sliding-idle-deadline pushed the window past the ping stretch.

```
$ cd javascript && pnpm vitest run src/voice/adapters/__tests__/elevenlabs.test.ts --reporter=verbose

 RUN  v4.1.8 /home/ubuntu/langwatch-workspace/worktrees/iss661/javascript

 ✓ src/voice/adapters/__tests__/elevenlabs.test.ts > ElevenLabsAgentAdapter wire-protocol (onMessage branches) > decodes base64 PCM16 from an `audio` event and resolves the next receiver 1ms
 ✓ src/voice/adapters/__tests__/elevenlabs.test.ts > ElevenLabsAgentAdapter wire-protocol (onMessage branches) > trims an odd-byte PCM16 payload so AudioChunk doesn't throw on the invariant 1ms
 ✓ src/voice/adapters/__tests__/elevenlabs.test.ts > ElevenLabsAgentAdapter wire-protocol (onMessage branches) > queues audio events that arrive before any receiver is waiting 0ms
 ✓ src/voice/adapters/__tests__/elevenlabs.test.ts > ElevenLabsAgentAdapter wire-protocol (onMessage branches) > replies to a ping event with a pong carrying the event_id 0ms
 ✓ src/voice/adapters/__tests__/elevenlabs.test.ts > ElevenLabsAgentAdapter wire-protocol (onMessage branches) > skips pong when ping has no event_id (defensive) 0ms
 ✓ src/voice/adapters/__tests__/elevenlabs.test.ts > ElevenLabsAgentAdapter wire-protocol (onMessage branches) > captures user_transcript onto lastUserTranscript 0ms
 ✓ src/voice/adapters/__tests__/elevenlabs.test.ts > ElevenLabsAgentAdapter wire-protocol (onMessage branches) > captures agent_response onto lastAgentTranscript 0ms
 ✓ src/voice/adapters/__tests__/elevenlabs.test.ts > ElevenLabsAgentAdapter wire-protocol (onMessage branches) > agent_response_correction overrides lastAgentTranscript with the corrected text 0ms
 ✓ src/voice/adapters/__tests__/elevenlabs.test.ts > ElevenLabsAgentAdapter wire-protocol (onMessage branches) > warns on conversation_initiation_metadata format drift 1ms
 ✓ src/voice/adapters/__tests__/elevenlabs.test.ts > ElevenLabsAgentAdapter wire-protocol (onMessage branches) > swallows interruption and unknown events without error 0ms
 ✓ src/voice/adapters/__tests__/elevenlabs.test.ts > ElevenLabsAgentAdapter wire-protocol (onMessage branches) > ignores non-JSON frames cleanly 0ms
 ✓ src/voice/adapters/__tests__/elevenlabs.test.ts > ElevenLabsAgentAdapter wire-protocol (onMessage branches) > post-open socket error nulls this.ws and unblocks pending receivers 2ms
 ✓ src/voice/adapters/__tests__/elevenlabs.test.ts > ElevenLabsAgentAdapter wire-protocol (onMessage branches) > socket close event drains pending waiters 0ms
 ✓ src/voice/adapters/__tests__/elevenlabs.test.ts > ElevenLabsAgentAdapter wire-protocol (onMessage branches) > receiveAudio rejects with timeout when no audio arrives in time 51ms
 ✓ src/voice/adapters/__tests__/elevenlabs.test.ts > receiveAudio — keepalive-aware sliding idle deadline (#661) > timing invariant: ping stretch exceeds timeout budget 0ms
 ✓ src/voice/adapters/__tests__/elevenlabs.test.ts > receiveAudio — keepalive-aware sliding idle deadline (#661) > AC-KA1: receiveAudio tolerates a silent-but-pinging stretch longer than timeout 721ms
 ✓ src/voice/adapters/__tests__/elevenlabs.test.ts > receiveAudio — keepalive-aware sliding idle deadline (#661) > AC-KA2 regression guard: truly silent socket still times out 301ms
 ✓ src/voice/adapters/__tests__/elevenlabs.test.ts > receiveAudio — keepalive-aware sliding idle deadline (#661) > AC-KA5: socket close drains cleanly with no surviving timer 3ms

 Test Files  1 passed (1)
      Tests  39 passed (39)
   Start at  19:12:31
   Duration  3.41s (transform 370ms, setup 0ms, import 2.13s, tests 1.11s, environment 0ms)
```

_Inspected this turn: no `[error]`, all AC-tagged tests green, 39/39 pass, AC-KA1 at 721 ms (real-timer run proves the sliding window worked)._

**AC-KA4 waiver:** `ELEVENLABS_API_KEY` / `ELEVENLABS_AGENT_ID` confirmed absent from all checked env files this turn. Per feature spec line 86–90, the waiver is recorded here; AC-KA1 (keyless mechanistic unit test) is the merge gate.

rogeriochaves (approved):

PR #669, opened 2026-06-15
fix(voice/#662): guard #662's response.create call sites against the active-response race (JS + PY)

### LINKED ISSUE #662: OpenAI Realtime JS adapters send response.create unconditionally — same active-response race as #657
## For Humans

**JS and Python `send_text` adapters send `response.create` unconditionally — same active-response race as #657, three call sites unguarded.**

The Python `recv_audio` fix in PR #659 introduced `_response_active` tracking (set on `response.created`, cleared on `response.done`/`response.cancelled`) and a deferred-send mechanism for the user-audio branch. The same race exists in three other locations not in scope for #659:

- `javascript/src/voice/adapters/openai-realtime.ts` — `receiveAudio` (line 376) and `sendText` (line 680): zero response-lifecycle state in the adapter.
- `javascript/src/agents/realtime/realtime-agent.adapter.ts` — `handleInitialResponse` (line 189) and `handleAudioInput` (line 231): both unconditional.
- `python/scenario/voice/adapters/openai_realtime.py` — `send_text` (line 691): no `_response_active` guard. `_deferred_response_create` does not yet exist on `main`; both fields ship with PR #659 and must be added here (or this issue rebased on #659 once merged).

**Ask:** Add response-lifecycle tracking and deferred-send to each unguarded call site, matching the guard pattern from PR #659.

_Status: `bug` / `P2 - medium` / `not-ready`. Full spec below._

### Symptoms / context

The OpenAI Realtime API returns `{"type":"error","error":{"message":"Conversation already has an active response in progress"}}` (and may silently drop the conflicting request) when `response.create` arrives while a response is already streaming. PR #659 fixed this for the `recv_audio` user-audio branch in Python by introducing `_response_active` state and a `_deferred_response_create` flag. The sweep comment on PR #659 identified three additional unguarded sites.

**Reproduction attempt:** A hermetic mock-WS repro is straightforward using the `_MockWS` pattern from `python/tests/voice/test_realtime_response_create_guard.py` (PR #659 worktree — the file does not yet exist on `main`). For JS, the `FakeTransport` pattern from `javascript/src/agents/__tests__/realtime-adapter-frames.test.ts` applies directly. The specific missing input blocking a live repro is an OpenAI Realtime API key; the mock-WS hermetic tests cover the observable failure shape. See AC-PY1, AC-JS1 for the exact test shapes.

### Root cause / key findings

**Four unguarded `response.create` sends across three files:**

**1. `javascript/src/voice/adapters/openai-realtime.ts` — `receiveAudio` (line 376)**

```typescript
if (this._pendingAudioBytes > 0) {
  this._ws.send(JSON.stringify({ type: "input_audio_buffer.commit" }));
  this._ws.send(JSON.stringify({ type: "response.create" }));   // ← unguarded
  this._pendingAudioBytes = 0;
}
```

The adapter has **no `_responseActive` field, no `response.created` listener, no `response.done` lifecycle listener** in the `receiveAudio` event loop. `response.done`/`response.cancelled` are handled only for tool-call drain (line 463) — they do not update any lifecycle flag.

**2. `javascript/src/voice/adapters/openai-realtime.ts` — `sendText` (line 680)**

```typescript
async sendText(text: string): Promise<void> {
  // ...conversation.item.create...
  this._ws.send(JSON.stringify({ type: "response.create" }));  // ← unguarded
}
```

`sendText` is called from the `role=USER` scripted-step path. If the agent-role adapter is still streaming when the user simulator calls `sendText`, the server rejects or silently drops the create.

**3. `javascript/src/agents/realtime/realtime-agent.adapter.ts` — `handleInitialResponse` (line 189) and `handleAudioInput` (line 231)**

```typescript
transport.sendEvent({ type: "response.create" });   // ← unguarded (both methods)
```

`RealtimeEventHandler` listens on `response.done` (line 111 of `realtime-event-handler.ts`) to resolve `waitForResponse`, but exposes no lifecycle flag. A concurrent call arriving between the `sendEvent` and the `response.done` resolution would double-fire.

**4. `python/scenario/voice/adapters/openai_realtime.py` — `send_text` (line 691)**

```python
await self._ws.send(json.dumps({"type": "response.create"}))  # ← unguarded
```

`_response_active` and `_deferred_response_create` do **not** yet exist on `main` — they are introduced by PR #659. This fix requires either merging PR #659 first, or adding both fields (`_response_active` and `_deferred_response_create`) as part of this issue. The guard mirrors the `recv_audio` user-audio branch from PR #659. `send_text` is called from `scenario_executor.py:971` via the `role=USER` realtime path.

### Strategies tested / considered

| Strategy | Assessment |
|---|---|
| Guard on `_responseActive` flag, defer-don't-drop (Python pattern from #659) | **Recommended** — proven fix; preserves wire correctness, no dropped turns |
| Drop `response.create` when active (no defer) | Incorrect — causes silent turn loss |
| Rely on server error event + retry loop | Over-complex; server may silently ignore rather than error; adds latency |
| Serialize all `response.create` calls through a queue | Architecturally heavier; no evidence of multiple concurrent callers here |

- **Python `send_text`**: `_response_active` and `_deferred_response_create` are **net-new** on `main` (they land with PR #659). Add both fields in `__init__` if PR #659 is not yet merged, then add guard before line 691: `if self._response_active: self._deferred_response_create = True; return`. The deferred-create path in the `recv_audio` event loop at line ~423 fires it after `response.done` clears the flag.
- **JS `openai-realtime.ts` `receiveAudio` + `sendText`**: add `private _responseActive = false` to the class. Set `true` on `response.created` in the `receiveAudio` event loop, clear on `response.done`/`response.cancelled` (the tool-call branch at line 463 already handles those events — add the flag clear there). Guard the preamble at line 374: commit always; `response.create` only if `!this._responseActive`, else set `this._deferredResponseCreate = true`. Add the deferred-fire inside the loop on `response.done`. Guard `sendText` identically.
- **JS `realtime-agent.adapter.ts`**: `RealtimeEventHandler` already listens on `response.done` (line 111) — add a `private _active = false` field, set on `response.created`, cleared on `response.done`. Expose `isResponseActive(): boolean`. Guard both `handleInitialResponse` and `handleAudioInput` before `transport.sendEvent({ type: "response.create" })`.
- **Test pattern**: `test_realtime_response_create_guard.py` (PR #659 worktree at `.worktrees/issue657-realtime-response-create-race/python/tests/voice/`) is the authoritative mock-WS pattern for Python. Mirror for JS using `FakeTransport` from `realtime-adapter-frames.test.ts`.
- No migration / rollback surface — pure in-memory state; no schema or persistent side effects.

### Caveats

- **Deferred-consume safety**: `_deferred_response_create` set by `send_text` must always be consumed. `recv_audio` must be called after `send_text` on every code path through `scenario_executor.py:971`. Currently the executor always drains after `send_text`; verify this assumption before shipping. An unconsumed deferred create (no `recv_audio` follows) is a **known limitation** if the code path cannot be reached in tests — document it with an `# INVARIANT:` comment at the `send_text` call site.
- For `realtime-agent.adapter.ts`, confirm `transport.on("response.created", ...)` fires on the `@openai/agents` `RealtimeSession` transport before wiring the lifecycle flag.

## Acceptance Criteria

> **Precondition (Python half):** AC-PY1, AC-PY2, AC-DEFER, AC-REG2, and AC-ERR1 (Python side) assume PR #659 (`fix/657`) is merged to `main` first. PR #659 introduces `_deferred_response_create`, the `recv_audio` deferred-consumption handler, and `test_realtime_response_create_guard.py` (AC1–AC7). If #659 is not yet merged, #662 must include those additions.

### Functional

- [ ] **AC-PY1**: `_deferred_response_create` is ADDED to `openai_realtime.py` (net-new; introduced by PR #659 or added here if #659 is not yet merged). `send_text`'s guard reuses the existing `_response_active` field (already present on `main`, line 131). Given `send_text` is called while `self._response_active is True`, `response.create` is NOT sent immediately and `_deferred_response_create` is set `True`. Evidence: hermetic mock-WS test asserting `mock_ws.count_of("response.create") == 0` (active case) and `_deferred_response_create == True`.
- [ ] **AC-PY2**: The deferred `response.create` set by `send_text` is consumed by the existing `response.done`/`response.cancelled` handler in the `recv_audio` event loop (the consumption logic introduced by PR #659 — #662 does not re-add this handler, only feeds the flag from `send_text`). The deferred create fires AFTER `response.done` is received. Evidence: interleaved-log ordering assertion `log_create > log_done` in the mock-WS test.
- [ ] **AC-DEFER**: After `send_text` defers a create (response active), the next `recv_audio` call consumes and fires the deferred create exactly once. Evidence: mock-WS test that calls `send_text` while active, then drives `recv_audio` to `response.done`, asserts `mock_ws.count_of("response.create") == 1`. An unconsumed deferred create (no `recv_audio` follows) is a **known limitation** — if untestable end-to-end, an `# INVARIANT:` comment at the `send_text` call site in `scenario_executor.py` names the assumption.
- [ ] **AC-JS1**: Given `receiveAudio` is called while `_responseActive` is `true`, `response.create` is NOT sent in the preamble and `input_audio_buffer.commit` IS sent. Evidence: vitest mock-WS test asserting `sentCount("response.create") === 0` and `sentCount("input_audio_buffer.commit") === 1`.
- [ ] **AC-JS2**: The deferred JS `receiveAudio` `response.create` fires after the in-flight `response.done`. Evidence: ordered-frame assertion `indexOfSent("response.create") > indexOfReceived("response.done")`.
- [ ] **AC-JS3**: Given `sendText` is called while `_responseActive` is `true`, `response.create` is NOT sent. Evidence: vitest test asserting `sentCount("response.create") === 0`.
- [ ] **AC-JS4**: `handleInitialResponse` on `RealtimeAgentAdapter` does not send `response.create` when `RealtimeEventHandler.isResponseActive()` returns `true`. Evidence: `FakeTransport` vitest test asserting `transport.count("response.create") === 0`.
- [ ] **AC-JS5**: `handleAudioInput` on `RealtimeAgentAdapter` does not send `response.create` when `RealtimeEventHandler.isResponseActive()` returns `true`. Evidence: same pattern as AC-JS4 for the audio-input branch.
- [ ] **AC-JS6**: `RealtimeEventHandler.isResponseActive()` returns `true` after a `response.created` event is received and `false` after `response.done`. Evidence: vitest unit test on `RealtimeEventHandler` asserting `isResponseActive()` transitions: `false` → (receive `response.created`) → `true` → (receive `response.done`) → `false`.

- [ ] **AC-REG1**: The normal (uncontested) path for each call site — `receiveAudio`, `sendText`, `handleAudioInput`, `handleInitialResponse` — still sends `response.create` immediately when no response is active and returns a valid result. Evidence: hermetic mock-WS/FakeTransport test per call site asserting normal-path behaviour is unchanged (mirrors AC5 from `test_realtime_response_create_guard.py`).
- [ ] **AC-REG2**: **Depends on PR #659 landing first** (it introduces `test_realtime_response_create_guard.py` with AC1–AC7 — the file does not exist on `main`). After #662's `send_text` guard is added on top of #659, those AC1–AC7 tests continue to pass. Evidence: `pytest python/tests/voice/test_realtime_response_create_guard.py` exits 0 on the #659+#662 merged tree.
- [ ] **AC-REG3**: Existing JS frame-ordering tests in `realtime-adapter-frames.test.ts` continue to pass after `RealtimeEventHandler` lifecycle changes. The two tests are `"fires exactly one response.create before any user-audio commit and yields one non-empty assistant turn"` and `"sends exactly N response.create frames for N user turns, none before the first audio commit"`. Evidence: `vitest run src/agents/__tests__/realtime-adapter-frames.test.ts` exits 0.
- [ ] **AC-ERR1**: If the server sends `{"type":"error","error":{"message":"Conversation already has an active response in progress",...}}` despite the guard, the adapter surfaces it as a thrown `Error` / `RuntimeError` — it does not swallow or hang. Evidence: **Depends on PR #659 landing first for the Python side** (AC6 in `test_realtime_response_create_guard.py` — not present on `main`); vitest test for JS `receiveAudio` asserting `receiveAudio` rejects with an `Error` whose message contains `"active response in progress"` and does not hang past a 5 s timeout.

## Plan

**Approach.** Apply the **proven defer-don't-drop guard from PR #659** to the four unguarded `response.create` sites. **Python reuses the existing `_response_active` / `_deferred_response_create` fields** (already on this branch via #659) — only `send_text` needs the guard + a deferred-fire hook in the `recv_audio` loop. **JS needs a net-new response-lifecycle flag** because neither adapter tracks one today: a `_responseActive` field in `openai-realtime.ts` (driven by `response.created` / `response.done` / `response.cancelled` in the `receiveAudio` loop) and a `_active` flag behind `RealtimeEventHandler.isResponseActive()` consumed by `realtime-agent.adapter.ts`. Every change is TDD against hermetic mock transports — no live API.

**Sequencing.**
1. **Python `send_text` guard** — add the `if self._response_active` guard before the send; wire the deferred-fire on `response.done` in the `recv_audio` loop. Completion: AC-PY1/PY2/AC-DEFER mock-WS tests green, AC-REG2 (existing #657 suite) still green.
2. **JS `openai-realtime.ts`** — add `_responseActive` + `_deferredResponseCreate`; set/clear in the `receiveAudio` event loop (clear at the existing `response.done`/`response.cancelled` tool-drain branch); guard `receiveAudio` preamble (commit always, create only if not active) and `sendText`; fire deferred create on `response.done`. Completion: AC-JS1/JS2/JS3 + AC-ERR1 (JS) green.
3. **JS `realtime-agent.adapter.ts` + `realtime-event-handler.ts`** — add `_active` flag + `isResponseActive()` on the handler (set on `response.created`, clear on the existing `response.done` listener); guard both `handleInitialResponse` and `handleAudioInput`. Completion: AC-JS4/JS5/JS6 green, AC-REG3 (frame-ordering suite) still green.
4. **Regression sweep** — run the full Python voice suite and the JS adapter suites; confirm AC-REG1/REG2/REG3 across all call sites.

Each phase is **fail-first**: write the mock-transport test, watch it fail against unguarded code, then add the guard.

**Key decisions.**
- **Defer, don't drop** — preserves wire correctness and turn fidelity; dropping silently loses turns (rejected in investigation).
- **Reuse Python #659 fields vs re-introduce** — reuse; they already exist on this branch. No duplicate state.
- **JS lifecycle flag owned by `RealtimeEventHandler`** (not the adapter) for the agent-role path — the handler already listens on `response.done`, so the flag rides the existing listener rather than adding a parallel one.
- **Hermetic mock transports only** — Python `_MockWS` interleaved-log pattern from `test_realtime_response_create_guard.py`; JS `FakeTransport` from `realtime-adapter-frames.test.ts`. A live OpenAI key is not required and not used.

**Risks / open questions.**
- **Deferred-consume safety** — `_deferred_response_create` set by `send_text` must always be drained by a following `recv_audio`. The executor currently always drains after `send_text`; verify before shipping and add an `# INVARIANT:` comment at the `send_text` call site if the unconsumed path is unreachable in tests (per AC-DEFER).
- **`response.created` firing on the `@openai/agents` transport** — confirm `transport.on("response.created", …)` actually fires on the `RealtimeSession` transport before relying on it for the `_active` flag; fall back to setting the flag at `sendEvent` time if it does not.
- **PR base** — branch `fix/662` is based on `fix/657`; **PR base is `fix/657` and retargets to `main` after #659 merges**. Until then the diff must read cleanly against #659's HEAD.

**Out of scope.**
- FSM/state-machine refactor of the response lifecycle (**#663**).
- Server-VAD mode — manual-VAD only, consistent with #657.
- Live-API / real-OpenAI integration tests.
- Any queue/serialization architecture for concurrent `response.create` callers (no evidence of multiple concurrent callers).

</details>

drewdrewthis:
## Why

The OpenAI Realtime API rejects (or silently drops) a \`response.create\` sent while a response is already streaming — \`{"type":"error","message":"Conversation already has an active response in progress"}\`. PR #659 fixed this race for the Python \`recv_audio\` user-audio branch; three JS call sites and one Python call site were left unguarded. Closes #662.

## What changed

- **\`RealtimeEventHandler\` gained \`_active\` / \`isResponseActive()\`** — set on \`response.created\`, cleared on \`response.done\`. This is the authoritative lifecycle flag for the agent-role JS path; no parallel listener needed.
- **\`openai-realtime.ts\` gained \`_responseActive\` + \`_deferredResponseCreate\`** — \`receiveAudio\` guards the \`response.create\` preamble (commit always, create only when not active, else defer); \`sendText\` guards identically. Deferred create fires on \`response.done\`/\`response.cancelled\` inside the existing tool-drain branch. \`_responseActive\` is set immediately after firing the deferred create (closes race window before \`response.created\` arrives — mirrors Python).
- **\`RealtimeAgentAdapter.handleInitialResponse\` + \`handleAudioInput\` guard via \`isResponseActive()\`** — both unconditional \`transport.sendEvent({ type: "response.create" })\` calls now short-circuit when the handler reports active.
- **Python \`openai_realtime.send_text\` guards via \`_response_active\` / \`_deferred_response_create\`** (fields introduced by #659, already on this branch) — matches the \`recv_audio\` pattern exactly.
- **Python \`recv_audio\`: \`_agent_turn_pending\` cleared before the if/else** — was only cleared in the \`else\` branch, leaving it set in the deferred path and allowing a spurious \`response.create\` on the next \`recv_audio\` call.

## Scope & carve-out (post-rebase onto `main`)

**This PR was rebased onto `main`** (was CONFLICTING → now MERGEABLE), resolving conflicts against #707 (realtime-USER duplex), #767 (Python test-through-the-wrapper), #724/#750 (injectable `Logger`) and #751/#755 (non-null-assertion ban). The full guard + test suites are green on the rebased tree, and all 14 prior review threads were re-verified and answered.

**Guarded — every `response.create` site on the agent path this issue enumerates, in BOTH languages:**

| Site | Lang | Strategy |
|---|---|---|
| `receiveAudio` preamble | JS | defer + refire on `response.done`/`.cancelled` |
| `sendText` | JS | defer + refire |
| `RealtimeAgentAdapter.handleInitialResponse` / `handleAudioInput` | JS | suppress (AC-JS4/5 — the awaited response *is* the active one) |
| `recv_audio` preamble + agent-turn branch | PY | defer + refire / guarded `elif` |
| `send_text` | PY | defer + refire |

Plus **reconnect hygiene in both languages**: `connect()`/`disconnect()` reset the lifecycle flags so a reused adapter never carries a stale active/deferred flag into its next connection (JS landed via the thread-#10 fix; **Python back-ported in this PR** with a hermetic test).

**Deliberately NOT guarded — two realtime-USER duplex sites introduced by #707** (`openai-realtime.ts` `_speakVerbatim`, `_speakGeneratedTurn`). They **post-date #662's scope** (rebased in from `main`) and are safe by a *different* invariant: **single-flight by construction** — each USER turn is fully drained before the next, and `_speakGeneratedTurn` self-commits `_pendingAudioBytes` so the drain cannot fire a second create. They send **parameterized** creates (persona `instructions`; `_speakVerbatim` is out-of-band, `conversation:"none"`+`input:[]`), whereas the `_responseActive` defer machinery refires a **bare** `response.create` — reusing it there would drop the persona/isolation and reintroduce the #705 drift. In-code carve-out comments mark both sites. The one residual window — a >15 s mid-response inter-frame stall stranding `_responseActive` true into the next turn — is tracked in **#792** (which notes the fix must use a non-bare-refire mechanism).

## Test plan

All tests are **hermetic mock-transport / mock-WS** — no live OpenAI key required.

**Python:**
\`\`\`
cd python && pytest tests/voice/test_realtime_response_create_guard.py -v
\`\`\`
Covers AC-PY1 (deferred when active), AC-PY2 (fires after \`response.done\`), AC-DEFER (consumed exactly once), and AC-REG1 (normal path unchanged).

**JavaScript:**
\`\`\`
cd javascript && npm test -- --testPathPattern="openai-realtime.test|realtime-adapter|realtime-event-handler"
\`\`\`
Covers AC-JS1–6: \`receiveAudio\` guard (deferred create, frame ordering), \`sendText\` guard, \`handleInitialResponse\`+\`handleAudioInput\` guard, \`isResponseActive()\` lifecycle transitions. AC-REG3: existing frame-ordering tests in \`realtime-adapter-frames.test.ts\` still pass.

## Human verification

For a skeptical reviewer who wants to confirm this independently — **no OpenAI API key required**:

1. **Check out the branch** — `git fetch origin && git checkout fix/issue662-realtime-js-response-create`.
2. **Run the Python guard suite** — `cd python && pytest tests/voice/test_realtime_response_create_guard.py -v`. Confirm all guard cases pass (AC1–AC7 plus the `send_text` cases `test_ac_py1/_py2/_defer`).
3. **Run the JS guard suites** — `cd javascript && npm test -- realtime-response-guard`. Confirm `AC-JS4/JS5` (`RealtimeAgentAdapter` guard), `AC-JS6a–d` (`isResponseActive()` lifecycle), and the `OpenAIRealtimeAgentAdapter — response.create guard (#662)` cases pass.
4. **Falsifiability** — in `python/scenario/voice/adapters/openai_realtime.py`, delete the `if self._response_active: self._deferred_response_create = True; return` guard inside `send_text` (so `response.create` fires unconditionally), re-run step 2, and confirm `test_ac_py1_send_text_suppressed_while_response_active` now FAILS. Restore the guard → it passes again.
5. **(Optional, live key)** Run `cd python && python3 tests/voice/demo_response_create_guard.py` against a real OpenAI Realtime endpoint and confirm no `Conversation already has an active response in progress` error appears — matching the captured run below.

**Scope: backend-only, no UI surface.** Every file in this PR is voice-loop guard logic or its tests — the JS realtime adapter/event-handler (`javascript/src/agents/realtime/`, `javascript/src/voice/adapters/openai-realtime.ts`) and the Python realtime adapter (`python/scenario/voice/adapters/openai_realtime.py`). The PR's **Files changed** tab is entirely `.ts`/`.py` (8 files, no `.tsx`/`.jsx`/`.vue`/CSS/HTML), so there is no frontend surface to screenshot. Proof is the falsifiability tests above plus the captured live-API wire log below. 

**Guard demonstration — hermetic stub + captured live-API run:**

Script: \`python/tests/voice/demo_response_create_guard.py\` (run via \`cd python && python3 tests/voice/demo_response_create_guard.py\`)

The committed demo exercises the **production `OpenAIRealtimeAgentAdapter` code** through a `LoggingWS` stub that replays scripted server events and captures every send/recv. **No OpenAI API key required.** The stub makes the guard observable without a network connection.

The captured output below is from a **real OpenAI Realtime API run** (`gpt-realtime-mini`, 2026-06-15) using a live-API development harness (not committed to the repo). It shows the same three-phase sequence against an actual WebSocket connection, confirming the guard works end-to-end.

Three phases exercise the previously-unguarded \`send_text()\` call site:

1. **Phase 1** — start a real agent response (\`response.create\` → receive real audio deltas)
2. **Phase 2** — call \`send_text()\` while \`_response_active=True\` — the guard fires: \`conversation.item.create\` is sent, NO duplicate \`response.create\`
3. **Phase 3** — drain until \`response.done\` → deferred \`response.create\` fires automatically, second response completes

**Captured live run output (2026-06-15, gpt-realtime-mini):**

\`\`\`
================================================================
  response.create guard — LIVE API DEMO
  PR #669  langwatch/scenario
  Real OpenAI Realtime WebSocket — gpt-realtime-mini
================================================================

[t=0.001s] Connecting to wss://api.openai.com/v1/realtime?model=gpt-realtime-mini ...
[t=0.547s] Connected.

[t=0.547s] → SEND session.update
[t=0.555s] ← RECV session.created

────────────────────────────────────────────────────────────────
  PHASE 1: Agent-initiated response (real audio exchange)
────────────────────────────────────────────────────────────────

[t=0.555s] Calling recv_audio(timeout=30) — waiting for first audio chunk ...
[t=0.555s] → SEND response.create
[t=0.681s] ← RECV session.updated
[t=0.689s] ← RECV response.created
[t=0.973s] ← RECV response.output_item.added
[t=0.973s] ← RECV conversation.item.added
[t=1.046s] ← RECV response.content_part.added
[t=1.046s] ← RECV response.output_audio_transcript.delta
[t=1.061s] ← RECV response.output_audio_transcript.delta
[t=1.212s] ← RECV response.output_audio_transcript.delta
[t=1.219s] ← RECV response.output_audio_transcript.delta
[t=1.464s] ← RECV response.output_audio.delta  (19200 bytes)
[t=1.464s] *** First audio chunk received (19200 bytes) ***
         _response_active = True

────────────────────────────────────────────────────────────────
  PHASE 2: send_text() while _response_active=True
────────────────────────────────────────────────────────────────

[t=1.464s]  _response_active before send_text  = True
[t=1.464s]  _deferred_response_create before   = False
[t=1.464s]  GUARD CONDITION MET — calling send_text() ...
[t=1.464s] → SEND conversation.item.create
[t=1.464s]  _deferred_response_create after    = True
[t=1.464s]  response.create count after send_text: 1 (was 1)

────────────────────────────────────────────────────────────────
  PHASE 3: Drain — watching for response.done + deferred create
────────────────────────────────────────────────────────────────

[t=1.464s] ← RECV response.output_audio.delta  (19200 bytes)
... [23 more audio deltas from original response] ...
[t=4.627s] ← RECV response.done
[t=4.627s] → SEND response.create          ← DEFERRED create fires after response.done
[t=4.641s] ← RECV rate_limits.updated
[t=4.765s] ← RECV response.created
... [13 more audio deltas from deferred response] ...
[t=6.464s] ← RECV response.done

────────────────────────────────────────────────────────────────
  WIRE LOG ASSESSMENT
────────────────────────────────────────────────────────────────

  Guard suppressed response.create during send_text:    YES ✓
  _deferred_response_create=True after send_text():     YES ✓
  conversation.item.create was sent:                     YES ✓
  Deferred response.create fired after response.done:    YES ✓
  response.create count before first response.done:      1 (expected 1) ✓
    (log idx: response.done=100, deferred create=101)

  ALL CHECKS PASS — guard is operating correctly on real API.
\`\`\`

The key sequence from the full wire log:
- **t=1.464s**: \`SEND conversation.item.create\` — NO second \`response.create\` (GUARD SUPPRESSED IT)
- **t=4.627s**: \`RECV response.done\` — deferred create fires immediately
- **t=4.627s**: \`SEND response.create\` — fires strictly AFTER \`response.done\` (log idx 101 > 100)

## Anything surprising?

**\`response.cancelled\` clears \`_active\` / \`_responseActive\`** — on \`response.cancelled\` the server is done streaming, so the flag must clear and any deferred create must fire. Both JS paths clear on cancel as well as done. This isn't obvious from the issue spec (which only mentions \`response.done\`) but is required for correctness if the user interrupts.

**\`_agent_turn_pending\` must be cleared before the deferred branch** — the original fix only cleared it in the \`else\` path, leaving it set when deferring. A subsequent \`recv_audio\` call would then fire a spurious bare \`response.create\` via the \`elif _agent_turn_pending\` guard. Fixed in the review-findings commit.

**JS \`_responseActive\` must be set immediately after firing the deferred create** — between the deferred send and the server's \`response.created\` ACK, \`_responseActive\` was \`false\`, leaving a race window where \`sendText\` could fire a second \`response.create\`. Fixed by mirroring Python's pattern of setting the flag immediately.

## Verification Report

| AC-PY1 | `send_text` defers `response.create` when `_response_active=True`; sets `_deferred_response_create=True` | `test_ac_py1_send_text_suppressed_while_response_active` passes; `_deferred_response_create` field at `openai_realtime.py:153` | ✅ PASS |
| AC-PY2 | Deferred `send_text` create fires after `response.done` (ordering: `log_create > log_done`) | `test_ac_py2_send_text_deferred_create_fires_after_response_done` passes with interleaved-log ordering check | ✅ PASS |
| AC-DEFER | Deferred create consumed exactly once by next `recv_audio` | `test_ac_defer_send_text_deferred_create_fires_exactly_once` passes; `count_of("response.create") == 1` | ✅ PASS |
| AC-JS1 | `receiveAudio` preamble: commit sent, `response.create` NOT sent when `_responseActive=True` | `AC-JS1` vitest test passes: `sentCount("input_audio_buffer.commit")==1`, `sentCount("response.create")==0` | ✅ PASS |
| AC-JS2 | Deferred `receiveAudio` `response.create` fires after `response.done` | `AC-JS2` vitest test passes: ordered-frame assertion via FakeWS EventEmitter recv queue | ✅ PASS |
| AC-JS3 | `sendText` does not send `response.create` when `_responseActive=True` | `AC-JS3` vitest test passes: `sentCount("response.create")==0` | ✅ PASS |
| AC-JS4 | `handleInitialResponse` does not send `response.create` when `isResponseActive()=true` | `AC-JS4` vitest test passes; guard at `realtime-agent.adapter.ts:188` | ✅ PASS |
| AC-JS5 | `handleAudioInput` does not send `response.create` when `isResponseActive()=true` | `AC-JS5` vitest test passes; guard at `realtime-agent.adapter.ts:232` | ✅ PASS |
| AC-JS6 | `isResponseActive()` transitions: `false` → `response.created` → `true` → `response.done` → `false` | AC-JS6a/b/c/d vitest tests pass; `realtime-event-handler.ts:192` | ✅ PASS |
| AC-REG1 | Normal (uncontested) paths still fire `response.create` immediately | Python: `test_ac5_normal_path_commit_then_create` PASS; JS: normal-path branches in all guard tests PASS | ✅ PASS |
| AC-REG2 | Pre-existing AC1–AC7 from `test_realtime_response_create_guard.py` still pass | `pytest tests/voice/test_realtime_response_create_guard.py` → 10 passed, 0 failed | ✅ PASS |
| AC-REG3 | Existing JS frame-ordering tests in `realtime-adapter-frames.test.ts` still pass | `vitest run realtime-adapter-frames.test.ts` → 2 passed | ✅ PASS |
| AC-ERR1 | Server `"active response in progress"` error surfaces as rejected `Error` / `RuntimeError` | Python: `test_ac6_server_rejection_raises_runtime_error` PASS; JS: `AC-ERR1` vitest test PASS | ✅ PASS |

**Verdict: 13/13 verified** — ALL PASS. Demo script (`demo_response_create_guard.py`) exercises production adapter code hermetically; live-API run output above confirms end-to-end behavior against real OpenAI Realtime WebSocket.

### DIFF
--- a/javascript/src/agents/__tests__/realtime-response-guard.test.ts
+++ b/javascript/src/agents/__tests__/realtime-response-guard.test.ts
@@ -109,8 +109,11 @@ describe("RealtimeAgentAdapter — response.create guard", () => {
     // Make the handler believe a response is active by firing response.created
     transport.fire("response.created");
 
-    // handleInitialResponse is private — call via any
-    const promise = (adapter as any).handleInitialResponse();
+    // handleInitialResponse is private — reach it via a typed cast (not `any`),
+    // matching the sibling openai-realtime-response-guard.test.ts pattern.
+    const promise = (
+      adapter as unknown as { handleInitialResponse(): Promise<unknown> }
+    ).handleInitialResponse();
 
     // The send decision is made synchronously, before the method's first `await`
     // (waitForResponse), so response.create has already been sent or suppressed by
@@ -133,7 +136,9 @@ describe("RealtimeAgentAdapter — response.create guard", () => {
     transport.fire("response.created");
 
     const b64audio = Buffer.from(new Uint8Array(160)).toString("base64");
-    const promise = (adapter as any).handleAudioInput(b64audio);
+    const promise = (
+      adapter as unknown as { handleAudioInput(audio: string): Promise<unknown> }
+    ).handleAudioInput(b64audio);
 
     // Same deterministic settle as AC-JS4: sends are synchronous before the first
     // await, so flush microtasks then resolve waitForResponse promptly.
--- a/javascript/src/agents/__tests__/realtime-response-guard.test.ts
+++ b/javascript/src/agents/__tests__/realtime-response-guard.test.ts
@@ -112,8 +112,10 @@ describe("RealtimeAgentAdapter — response.create guard", () => {
     // handleInitialResponse is private — call via any
     const promise = (adapter as any).handleInitialResponse();
 
-    // Give it a moment to call transport.sendEvent (or not)
-    await new Promise((r) => setTimeout(r, 20));
+    // Allow the synchronous preamble (transport.sendEvent) to flush before asserting.
+    // 100ms is sufficient on a loaded CI runner; the assertion is about event ordering,
+    // not a timed window.
+    await new Promise((r) => setTimeout(r, 100));
 
     // Cancel the timeout by firing response.done (resolves waitForResponse)
     transport.fire("response.done");
@@ -133,7 +135,8 @@ describe("RealtimeAgentAdapter — response.create guard", () => {
     const b64audio = Buffer.from(new Uint8Array(160)).toString("base64");
     const promise = (adapter as any).handleAudioInput(b64audio);
 
-    await new Promise((r) => setTimeout(r, 20));
+    // Same flush wait as AC-JS4 — 100ms is sufficient on a loaded CI runner.
+    await new Promise((r) => setTimeout(r, 100));
     transport.fire("response.done");
     await promise.catch(() => {});
 
--- a/javascript/src/agents/__tests__/realtime-response-guard.test.ts
+++ b/javascript/src/agents/__tests__/realtime-response-guard.test.ts
@@ -58,8 +58,7 @@ describe("RealtimeEventHandler — isResponseActive() lifecycle", () => {
     const transport = new FakeTransport();
     const session = makeFakeSession(transport);
     const handler = new RealtimeEventHandler(session);
-    // isResponseActive() doesn't exist pre-fix → throws TypeError
-    expect((handler as any).isResponseActive()).toBe(false);
+    expect(handler.isResponseActive()).toBe(false);
   });
 
   it("AC-JS6b: isResponseActive() returns true after response.created", () => {
@@ -67,7 +66,7 @@ describe("RealtimeEventHandler — isResponseActive() lifecycle", () => {
     const session = makeFakeSession(transport);
     const handler = new RealtimeEventHandler(session);
     transport.fire("response.created");
-    expect((handler as any).isResponseActive()).toBe(true);
+    expect(handler.isResponseActive()).toBe(true);
   });
 
   it("AC-JS6c: isResponseActive() returns false after response.done", () => {
@@ -76,7 +75,16 @@ describe("RealtimeEventHandler — isResponseActive() lifecycle", () => {
     const handler = new RealtimeEventHandler(session);
     transport.fire("response.created");
     transport.fire("response.done");
-    expect((handler as any).isResponseActive()).toBe(false);
+    expect(handler.isResponseActive()).toBe(false);
+  });
+
+  it("AC-JS6d: isResponseActive() returns false after response.cancelled", () => {
+    const transport = new FakeTransport();
+    const session = makeFakeSession(transport);
+    const handler = new RealtimeEventHandler(session);
+    transport.fire("response.created");
+    transport.fire("response.cancelled");
+    expect(handler.isResponseActive()).toBe(false);
   });
 });
 
--- a/javascript/src/agents/__tests__/realtime-response-guard.test.ts
+++ b/javascript/src/agents/__tests__/realtime-response-guard.test.ts
@@ -1,0 +1,136 @@
+/**
+ * Regression tests for issue #662 Phase 3:
+ *   AC-JS6: RealtimeEventHandler.isResponseActive() lifecycle transitions
+ *   AC-JS4: handleInitialResponse does NOT send response.create when handler is active
+ *   AC-JS5: handleAudioInput does NOT send response.create when handler is active
+ *
+ * These tests are written to FAIL before the fix is applied:
+ *   - AC-JS6a/b/c: isResponseActive() does not exist yet → TypeError
+ *   - AC-JS4/AC-JS5: response.create is sent unconditionally → count is 1, not 0
+ */
+
+import { describe, it, expect, vi } from "vitest";
+import { RealtimeEventHandler } from "../realtime/realtime-event-handler";
+import { RealtimeAgentAdapter } from "../realtime/realtime-agent.adapter";
+import { AgentRole } from "../../domain";
+import type { RealtimeSession } from "@openai/agents/realtime";
+
+// ------- FakeTransport -------
+class FakeTransport {
+  private listeners: Map<string, ((data: unknown) => void)[]> = new Map();
+  sentEvents: Array<{ type: string; [k: string]: unknown }> = [];
+
+  on(event: string, cb: (data: unknown) => void): void {
+    const existing = this.listeners.get(event) ?? [];
+    this.listeners.set(event, [...existing, cb]);
+  }
+
+  sendEvent(event: { type: string; [k: string]: unknown }): void {
+    this.sentEvents.push(event);
+  }
+
+  // Simulate a server event arriving
+  fire(event: string, data?: unknown): void {
+    const cbs = this.listeners.get(event) ?? [];
+    for (const cb of cbs) cb(data ?? {});
+  }
+
+  sentCount(type: string): number {
+    return this.sentEvents.filter((e) => e.type === type).length;
+  }
+}
+
+// ------- FakeSession -------
+function makeFakeSession(transport: FakeTransport): RealtimeSession {
+  return {
+    transport,
+    connect: vi.fn(),
+    close: vi.fn(),
+    sendMessage: vi.fn(),
+  } as unknown as RealtimeSession;
+}
+
+// ============================================================
+// AC-JS6: isResponseActive() lifecycle transitions
+// ============================================================
+describe("RealtimeEventHandler — isResponseActive() lifecycle", () => {
+  it("AC-JS6a: isResponseActive() is false initially", () => {
+    const transport = new FakeTransport();
+    const session = makeFakeSession(transport);
+    const handler = new RealtimeEventHandler(session);
+    // isResponseActive() doesn't exist pre-fix → throws TypeError
+    expect((handler as any).isResponseActive()).toBe(false);
+  });
+
+  it("AC-JS6b: isResponseActive() returns true after response.created", () => {
+    const transport = new FakeTransport();
+    const session = makeFakeSession(transport);
+    const handler = new RealtimeEventHandler(session);
+    transport.fire("response.created");
+    expect((handler as any).isResponseActive()).toBe(true);
+  });
+
+  it("AC-JS6c: isResponseActive() returns false after response.done", () => {
+    const transport = new FakeTransport();
+    const session = makeFakeSession(transport);
+    const handler = new RealtimeEventHandler(session);
+    transport.fire("response.created");
+    transport.fire("response.done");
+    expect((handler as any).isResponseActive()).toBe(false);
+  });
+});
+
+// ============================================================
+// AC-JS4 / AC-JS5: response.create guard on adapter methods
+// ============================================================
+describe("RealtimeAgentAdapter — response.create guard", () => {
+  function makeAdapter(transport: FakeTransport): RealtimeAgentAdapter {
+    const session = makeFakeSession(transport);
+    return new RealtimeAgentAdapter({
+      session,
+      role: AgentRole.AGENT,
+      agentName: "TestAgent",
+      responseTimeout: 100,
+    });
+  }
+
+  it("AC-JS4: handleInitialResponse does not send response.create when handler is active", async () => {
+    const transport = new FakeTransport();
+    const adapter = makeAdapter(transport);
+
+    // Make the handler believe a response is active by firing response.created
+    transport.fire("response.created");
+
+    // handleInitialResponse is private — call via any
+    const promise = (adapter as any).handleInitialResponse();
+
+    // Give it a moment to call transport.sendEvent (or not)
+    await new Promise((r) => setTimeout(r, 20));
+
+    // Cancel the timeout by firing response.done (resolves waitForResponse)
+    transport.fire("response.done");
+    await promise.catch(() => {}); // ignore timeout error
+
+    // response.create must NOT have been sent when handler was active
+    expect(transport.sentCount("response.create")).toBe(0);
+  });
+
+  it("AC-JS5: handleAudioInput does not send response.create when handler is active", async () => {
+    const transport = new FakeTransport();
+    const adapter = makeAdapter(transport);
+
+    // Make the handler believe a response is active
+    transport.fire("response.created");
+
+    const b64audio = Buffer.from(new Uint8Array(160)).toString("base64");
+    const promise = (adapter as any).handleAudioInput(b64audio);
+
+    await new Promise((r) => setTimeout(r, 20));
+    transport.fire("response.done");
+    await promise.catch(() => {});
+
+    // Only input_audio_buffer.append and input_audio_buffer.commit should be sent
+    // response.create must NOT be sent when handler is active
+    expect(transport.sentCount("response.create")).toBe(0);
+  });
+});
--- a/javascript/src/agents/__tests__/realtime-response-guard.test.ts
+++ b/javascript/src/agents/__tests__/realtime-response-guard.test.ts
@@ -112,14 +112,14 @@ describe("RealtimeAgentAdapter — response.create guard", () => {
     // handleInitialResponse is private — call via any
     const promise = (adapter as any).handleInitialResponse();
 
-    // Allow the synchronous preamble (transport.sendEvent) to flush before asserting.
-    // 100ms is sufficient on a loaded CI runner; the assertion is about event ordering,
-    // not a timed window.
-    await new Promise((r) => setTimeout(r, 100));
-
-    // Cancel the timeout by firing response.done (resolves waitForResponse)
+    // The send decision is made synchronously, before the method's first `await`
+    // (waitForResponse), so response.create has already been sent or suppressed by
+    // the time the promise is returned. Flush the microtask queue, then resolve
+    // waitForResponse promptly by firing response.done — so it settles well before
+    // responseTimeout (100ms) and the test is deterministic, not racing the clock.
+    await Promise.resolve();
     transport.fire("response.done");
-    await promise.catch(() => {}); // ignore timeout error
+    await promise.catch(() => {}); // ignore any rejection
 
     // response.create must NOT have been sent when handler was active
     expect(transport.sentCount("response.create")).toBe(0);
@@ -135,8 +135,9 @@ describe("RealtimeAgentAdapter — response.create guard", () => {
     const b64audio = Buffer.from(new Uint8Array(160)).toString("base64");
     const promise = (adapter as any).handleAudioInput(b64audio);
 
-    // Same flush wait as AC-JS4 — 100ms is sufficient on a loaded CI runner.
-    await new Promise((r) => setTimeout(r, 100));
+    // Same deterministic settle as AC-JS4: sends are synchronous before the first
+    // await, so flush microtasks then resolve waitForResponse promptly.
+    await Promise.resolve();
     transport.fire("response.done");
     await promise.catch(() => {});
 
--- a/javascript/src/agents/realtime/realtime-agent.adapter.ts
+++ b/javascript/src/agents/realtime/realtime-agent.adapter.ts
@@ -193,9 +193,11 @@ export class RealtimeAgentAdapter extends AgentAdapter {
       throw new Error("Realtime transport not available");
     }
 
-    transport.sendEvent({
-      type: "response.create",
-    });
+    if (!this.eventHandler.isResponseActive()) {
+      transport.sendEvent({
+        type: "response.create",
+      });
+    }
 
     const timeout = this.config.responseTimeout ?? 60000;
     const response = await this.eventHandler.waitForResponse(timeout);
@@ -234,10 +236,12 @@ export class RealtimeAgentAdapter extends AgentAdapter {
       type: "input_audio_buffer.commit",
     });
 
-    // Trigger response generation
-    transport.sendEvent({
-      type: "response.create",
-    });
+    // Trigger response generation — guard against active-response race
+    if (!this.eventHandler.isResponseActive()) {
+      transport.sendEvent({
+        type: "response.create",
+      });
+    }
 
     // Wait for audio response
     const timeout = this.config.responseTimeout ?? 60000;
--- a/javascript/src/voice/adapters/__tests__/openai-realtime-response-guard.test.ts
+++ b/javascript/src/voice/adapters/__tests__/openai-realtime-response-guard.test.ts
@@ -16,13 +16,17 @@
  */
 
 import { describe, it, expect } from "vitest";
+import { EventEmitter } from "node:events";
+import type WebSocket from "ws";
 import { OpenAIRealtimeAgentAdapter } from "../openai-realtime";
 
 // ---------------------------------------------------------------------------
-// FakeWS — an in-process fake that records frames without touching the network
+// FakeWS — an in-process fake that records frames without touching the network.
+// Extends EventEmitter so the adapter can register message/close/error listeners
+// exactly as it does on a real ws.WebSocket (via .on() / .once()).
 // ---------------------------------------------------------------------------
 
-class FakeWS {
+class FakeWS extends EventEmitter {
   sent: string[] = [];
   closed = false;
 
@@ -33,6 +37,7 @@ class FakeWS {
 
   close(): void {
     this.closed = true;
+    this.emit("close");
   }
 
   /** Ordered list of `type` fields from every sent frame. */
@@ -49,17 +54,36 @@ class FakeWS {
   indexOfSent(type: string): number {
     return this.sentTypes().indexOf(type);
   }
+
+  /**
+   * Push a server event into the adapter's receive loop.
+   * Serialises to JSON and emits as a "message" event — the same path a real
+   * WebSocket frame takes through _handleMessage.
+   */
+  receive(event: unknown): void {
+    this.emit("message", Buffer.from(JSON.stringify(event)));
+  }
 }
 
 // ---------------------------------------------------------------------------
 // Helpers
 // ---------------------------------------------------------------------------
 
-function makeAdapter(): { adapter: OpenAIRealtimeAgentAdapter; ws: FakeWS } {
-  const adapter = new OpenAIRealtimeAgentAdapter({ apiKey: "test-key" });
+async function makeAdapter(): Promise<{
+  adapter: OpenAIRealtimeAgentAdapter;
+  ws: FakeWS;
+}> {
   const ws = new FakeWS();
-  // Inject the fake WS directly — bypasses the real connect() call.
-  (adapter as unknown as Record<string, unknown>)._ws = ws;
+  const adapter = new OpenAIRealtimeAgentAdapter({
+    apiKey: "test-key",
+    wsFactory: (url, _authHeader) => {
+      // Emit "open" asynchronously so the awaited Promise in connect() resolves.
+      queueMicrotask(() => ws.emit("open"));
+      void url;
+      return ws as unknown as WebSocket;
+    },
+  });
+  await adapter.connect();
   return { adapter, ws };
 }
 
@@ -78,7 +102,7 @@ describe("OpenAIRealtimeAgentAdapter — response.create guard (#662)", () => {
   it(
     "AC-JS1: receiveAudio sends commit but NOT response.create when _responseActive is true",
     async () => {
-      const { adapter, ws } = makeAdapter();
+      const { adapter, ws } = await makeAdapter();
 
       // Simulate pending audio that triggers the preamble.
       (adapter as unknown as Record<string, unknown>)._pendingAudioBytes = 960;
@@ -111,7 +135,7 @@ describe("OpenAIRealtimeAgentAdapter — response.create guard (#662)", () => {
   it(
     "AC-JS2: deferred receiveAudio response.create fires after response.done frame",
     async () => {
-      const { adapter, ws } = makeAdapter();
+      const { adapter, ws } = await makeAdapter();
 
       (adapter as unknown as Record<string, unknown>)._pendingAudioBytes = 960;
       (adapter as unknown as Record<string, unknown>)._responseActive = true;
@@ -129,23 +153,14 @@ describe("OpenAIRealtimeAgentAdapter — response.create guard (#662)", () => {
       // Post-fix: 0 (deferred because _responseActive is true).
       const countBeforeDone = ws.sentCount("response.create");
 
-      // Now inject the event sequence that drives receiveAudio to completion:
+      // Inject the event sequence that drives receiveAudio to completion:
       // response.done  → clears _responseActive, fires deferred response.create
       // response.created → sets _responseActive = true for the new response
       // response.output_audio.delta → the actual audio frame that resolves receiveAudio
-      // response.done  → completes the second response
-      const enqueue = (event: unknown): void => {
-        (
-          adapter as unknown as {
-            _enqueueEvent: (e: unknown) => void;
-          }
-        )._enqueueEvent(event);
-      };
-
-      enqueue({ type: "response.done" });
+      ws.receive({ type: "response.done" });
       await new Promise<void>((r) => setTimeout(r, 5));
-      enqueue({ type: "response.created" });
-      enqueue({ type: "response.output_audio.delta", delta: b64audio });
+      ws.receive({ type: "response.created" });
+      ws.receive({ type: "response.output_audio.delta", delta: b64audio });
 
       await receivePromise;
 
@@ -168,7 +183,7 @@ describe("OpenAIRealtimeAgentAdapter — response.create guard (#662)", () => {
   it(
     "AC-JS3: sendText does not send response.create when _responseActive is true",
     async () => {
-      const { adapter, ws } = makeAdapter();
+      const { adapter, ws } = await makeAdapter();
 
       (adapter as unknown as Record<string, unknown>)._responseActive = true;
 
@@ -195,17 +210,13 @@ describe("OpenAIRealtimeAgentAdapter — response.create guard (#662)", () => {
   it(
     "AC-ERR1: receiveAudio rejects with Error on server active-response error",
     async () => {
-      const { adapter } = makeAdapter();
+      const { adapter, ws } = await makeAdapter();
 
       const receivePromise = adapter.receiveAudio(5.0);
 
       await new Promise<void>((r) => setTimeout(r, 10));
 
-      (
-        adapter as unknown as {
-          _enqueueEvent: (e: unknown) => void;
-        }
-      )._enqueueEvent({
+      ws.receive({
         type: "error",
         error: {
           message:
--- a/javascript/src/voice/adapters/__tests__/openai-realtime-response-guard.test.ts
+++ b/javascript/src/voice/adapters/__tests__/openai-realtime-response-guard.test.ts
@@ -144,8 +144,13 @@ describe("OpenAIRealtimeAgentAdapter — response.create guard (#662)", () => {
 
       const receivePromise = adapter.receiveAudio(2.0);
 
-      // Let the preamble run before injecting server events.
-      await new Promise<void>((r) => setTimeout(r, 20));
+      // Yield one macrotask so the async receiveAudio loop drains its queued
+      // events. A zero-delay timer is a deterministic macrotask boundary (it
+      // runs only after the microtask queue is empty, regardless of runner
+      // load) — not an arbitrary timed wait. The preamble already ran
+      // synchronously before receiveAudio's first await, so this only flushes
+      // the loop's event processing.
+      await new Promise<void>((r) => setTimeout(r, 0));
 
       // Snapshot: how many response.create frames have been sent so far?
       // Pre-fix: 1 (fired unconditionally in preamble).
@@ -157,7 +162,7 @@ describe("OpenAIRealtimeAgentAdapter — response.create guard (#662)", () => {
       // response.created → sets _responseActive = true for the new response
       // response.output_audio.delta → the actual audio frame that resolves receiveAudio
       ws.receive({ type: "response.done" });
-      await new Promise<void>((r) => setTimeout(r, 5));
+      await new Promise<void>((r) => setTimeout(r, 0));
       ws.receive({ type: "response.created" });
       ws.receive({ type: "response.output_audio.delta", delta: b64audio });
 
@@ -213,7 +218,9 @@ describe("OpenAIRealtimeAgentAdapter — response.create guard (#662)", () => {
 
       const receivePromise = adapter.receiveAudio(5.0);
 
-      await new Promise<void>((r) => setTimeout(r, 10));
+      // Macrotask yield so the loop reaches its first _nextEvent await before
+      // the error frame is injected (see AC-JS2 note above).
+      await new Promise<void>((r) => setTimeout(r, 0));
 
       ws.receive({
         type: "error",
--- a/javascript/src/voice/adapters/__tests__/openai-realtime-response-guard.test.ts
+++ b/javascript/src/voice/adapters/__tests__/openai-realtime-response-guard.test.ts
@@ -229,4 +229,50 @@ describe("OpenAIRealtimeAgentAdapter — response.create guard (#662)", () => {
     },
     6000,
   );
+
+  /**
+   * Reconnect hygiene (#662, CodeRabbit): disconnect() must clear the
+   * response-lifecycle guard flags, and a subsequent connect() must start from
+   * a clean slate — including clearing `_closeReason`, without which
+   * `_nextEvent` would reject every receiveAudio on the reused adapter and the
+   * guard-flag reset alone would be cosmetic.
+   */
+  it(
+    "resets response-lifecycle state on disconnect and reconnects clean",
+    async () => {
+      const { adapter } = await makeAdapter();
+      const state = adapter as unknown as Record<string, unknown>;
+
+      // Simulate mid-response state present when the socket drops.
+      state._responseActive = true;
+      state._deferredResponseCreate = true;
+
+      await adapter.disconnect();
+
+      // Teardown clears the guard flags so they cannot leak into a new session.
+      expect(state._responseActive).toBe(false);
+      expect(state._deferredResponseCreate).toBe(false);
+
+      // Reconnect the same instance: connect() must reset the guard flags AND
+      // the stale close reason.
+      const ws2 = new FakeWS();
+      (adapter as unknown as { _wsFactory: unknown })._wsFactory = (
+        _url: string,
+        _authHeader: string,
+      ) => {
+        queueMicrotask(() => ws2.emit("open"));
+        return ws2 as unknown as WebSocket;
+      };
+      await adapter.connect();
+
+      expect(state._responseActive).toBe(false);
+      expect(state._deferredResponseCreate).toBe(false);
+      expect(state._closeReason).toBe(null);
+
+      // Prove reconnect actually works: receiveAudio reaches its own timeout
+      // instead of rejecting immediately on the prior disconnect error.
+      await expect(adapter.receiveAudio(0.05)).rejects.toThrow(/timed out/);
+    },
+    2000,
+  );
 });
--- a/javascript/src/voice/adapters/__tests__/openai-realtime-response-guard.test.ts
+++ b/javascript/src/voice/adapters/__tests__/openai-realtime-response-guard.test.ts
@@ -76,10 +76,9 @@ async function makeAdapter(): Promise<{
   const ws = new FakeWS();
   const adapter = new OpenAIRealtimeAgentAdapter({
     apiKey: "test-key",
-    wsFactory: (url, _authHeader) => {
+    wsFactory: (_url, _authHeader) => {
       // Emit "open" asynchronously so the awaited Promise in connect() resolves.
       queueMicrotask(() => ws.emit("open"));
-      void url;
       return ws as unknown as WebSocket;
     },
   });
--- a/javascript/src/voice/adapters/__tests__/openai-realtime-response-guard.test.ts
+++ b/javascript/src/voice/adapters/__tests__/openai-realtime-response-guard.test.ts
@@ -1,0 +1,222 @@
+/**
+ * Regression tests for issue #662 — `response.create` guard on
+ * `OpenAIRealtimeAgentAdapter`.
+ *
+ * These tests are written against the PRE-FIX code and MUST FAIL until the
+ * following changes land in `openai-realtime.ts`:
+ *   - private _responseActive = false
+ *   - private _deferredResponseCreate = false
+ *   - receiveAudio preamble: commit always, defer response.create when active
+ *   - sendText: guard response.create when active
+ *   - response.created → _responseActive = true
+ *   - response.done/cancelled → _responseActive = false; fire deferred if set
+ *
+ * AC-JS1, AC-JS2, AC-JS3 FAIL on pre-fix code.
+ * AC-ERR1 is a control test — PASSES on pre-fix code (existing error path).
+ */
+
+import { describe, it, expect } from "vitest";
+import { OpenAIRealtimeAgentAdapter } from "../openai-realtime";
+
+// ---------------------------------------------------------------------------
+// FakeWS — an in-process fake that records frames without touching the network
+// ---------------------------------------------------------------------------
+
+class FakeWS {
+  sent: string[] = [];
+  closed = false;
+
+  /** Called by the adapter to send a frame. */
+  send(msg: string): void {
+    this.sent.push(msg);
+  }
+
+  close(): void {
+    this.closed = true;
+  }
+
+  /** Ordered list of `type` fields from every sent frame. */
+  sentTypes(): string[] {
+    return this.sent.map((s) => (JSON.parse(s) as { type: string }).type);
+  }
+
+  /** Count of frames with a given type. */
+  sentCount(type: string): number {
+    return this.sentTypes().filter((t) => t === type).length;
+  }
+
+  /** First index of a frame with a given type, or -1. */
+  indexOfSent(type: string): number {
+    return this.sentTypes().indexOf(type);
+  }
+}
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+function makeAdapter(): { adapter: OpenAIRealtimeAgentAdapter; ws: FakeWS } {
+  const adapter = new OpenAIRealtimeAgentAdapter({ apiKey: "test-key" });
+  const ws = new FakeWS();
+  // Inject the fake WS directly — bypasses the real connect() call.
+  (adapter as unknown as Record<string, unknown>)._ws = ws;
+  return { adapter, ws };
+}
+
+// ---------------------------------------------------------------------------
+// Tests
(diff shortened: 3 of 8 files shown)


drewdrewthis:
## Review verdict: READY

**Reviewed at:** `ca050ab` · **Run:** review (own-PR, 5-persona fan-out: principles / hygiene / test / security / devils-advocate) after the rebase onto `main`.

No blocking concerns at this SHA. All 14 prior review threads resolved; this review opened no new inline threads (findings were addressed in `ca050ab`, not deferred).

### What the review found and how it was resolved

- **[devils-advocate] Proof-integrity: the "all unguarded call sites" claim was false post-rebase** (the two #707 duplex sites `_speakVerbatim`/`_speakGeneratedTurn` are unguarded). **Reconciled** in `ca050ab`: PR title narrowed off "all", a **Scope & carve-out** section added, in-code carve-out comments at both sites, and the residual >15 s-stall edge filed as **#792**. Not a code defect — the two sites are single-flight-by-construction and a bare-refire guard would corrupt the #705 persona/isolation (so guarding them is the *wrong* fix; #792 notes the correct non-bare-refire approach).
- **[principles] Python reconnect-hygiene parity gap** — JS reset `_responseActive`/`_deferredResponseCreate` on connect/disconnect (thread #10) but Python did not. **Fixed** in `ca050ab`: Python `connect()`/`disconnect()` now reset both flags, with a hermetic test (`test_connect_disconnect_reset_response_lifecycle_flags`).
- **[principles] Undocumented eager-set** (`_responseActive=true` after the deferred fire vs. echo-reliant preamble/sendText) — **Fixed**: rationale comment added.
- **[hygiene] `(adapter as any)` private-method access** (2× `no-explicit-any`) — **Fixed**: swapped to a typed `as unknown as { … }` cast matching the sibling guard test, preserving thread #13's accepted private-access design.

### Non-blocking (accepted as-is / human-decide)

- **[test-reviewer][hygiene] Suppress-vs-defer asymmetry** between `RealtimeAgentAdapter` (suppress) and the voice adapter (defer-and-refire) — a **live human design question** (thread #8; raised by @rogeriochaves / @Aryansharma28: "if we guard, are we silently dropping the turn?"). Drew's thread-#8 resolution scoped it out of #662 (defer-and-replay would change the agreed AC-JS4/5 contract) and CodeRabbit withdrew. **Flagged for the requested human reviewer** — not resolved by this automated verdict.
- **[test-reviewer]** `AC-PY2`/`AC-DEFER` set an unused `_pending_audio_bytes` (isolation-clarity only; does not affect discrimination) — accepted; the assertions are load-bearing on the emitted wire frames.
- **[hygiene]** `FakeTransport`/harness duplication across sibling test files, `import/order` nits — pre-existing, non-gating (CI lint is red across the repo by design); left for a directory-wide cleanup.

### Verification at this SHA
Python guard suite 13/13 · JS guard suites 13/13 · broader JS realtime/voice 660 passed · `tsc` clean · `pyright` 0 errors · `no-explicit-any` cleared on changed files. (The local Python `*_e2e` errors + one `test_agent_wait_false` hang are the documented local-env / missing-`OPENAI_API_KEY` issues, untouched by this diff — CI is the authority.)

---
_Verdict is prose, not a GitHub approval. Own-PR review; the human-in-the-loop sign-off (esp. the suppress-vs-defer question) belongs to the requested human reviewer._

python/scenario/voice/adapters/openai_realtime.py
```diff
@@ -406,9 +416,18 @@ async def recv_audio(self, timeout: float) -> AudioChunk:
...
+                # Race guard (#657): server rejects/ignores a duplicate
+                # response.create while a response is in flight, which causes the
+                # recv loop to drain to timeout instead of completing.  Defer via
+                # _deferred_response_create (distinct from _agent_turn_pending) so
+                # the response.done handler fires response.create once the in-flight
+                # response clears.
+                self._deferred_response_create = True
+            else:
+                await self._ws.send(json.dumps({"type": "response.create"}))
+                self._agent_turn_pending = False  # user spoke → per-turn signal consumed
```
drewdrewthis: **[principles][uncle-bob][metz-beck] Bug: `_agent_turn_pending` not cleared in deferred branch — potential spurious double-fire**

`_agent_turn_pending = False` is only executed in the `else` branch (line 430). When `_response_active` is `True`, the deferred branch sets `_deferred_response_create = True` but leaves `_agent_turn_pending` set. A subsequent `recv_audio` call then falls into `elif self._agent_turn_pending and not self._response_active` and fires a spurious bare `response.create`.

**Fix:** clear `_agent_turn_pending` before the if/else:
```python
if self._pending_audio_bytes > 0:
    await self._ws.send(json.dumps({"type": "input_audio_buffer.commit"}))
    self._pending_audio_bytes = 0
    self._agent_turn_pending = False  # always clear when audio committed
    if self._response_active:
        self._deferred_response_create = True
    else:
        await self._ws.send(json.dumps({"type": "response.create"}))
```

javascript/src/voice/adapters/openai-realtime.ts
```diff
@@ -461,6 +476,11 @@ export class OpenAIRealtimeAgentAdapter extends VoiceAgentAdapter {
           }
         }
       } else if (etype === "response.done" || etype === "response.cancelled") {
+        this._responseActive = false;
+        if (this._deferredResponseCreate) {
+          this._ws.send(JSON.stringify({ type: "response.create" }));
+          this._deferredResponseCreate = false;
+        }
```
drewdrewthis, line 514: **[principles][hygiene][metz-beck][fowler] Bug: JS `_responseActive` not set after deferred create fires — race window until `response.created`**

After firing the deferred `response.create` (line 481), `_responseActive` remains `false` until the server's `response.created` event arrives (line 495). In this window, a concurrent `sendText` (line 702) sees `_responseActive = false` and fires a second `response.create`. Python closes this by setting `_response_active = True` immediately after firing (its line 527). JS should match.

**Fix:** set `this._responseActive = true` right after the deferred send:
```typescript
} else if (etype === "response.done" || etype === "response.cancelled") {
  this._responseActive = false;
  if (this._deferredResponseCreate) {
    this._ws.send(JSON.stringify({ type: "response.create" }));
    this._deferredResponseCreate = false;
    this._responseActive = true; // mirror Python: close the race window
  }
```

python/scenario/voice/adapters/openai_realtime.py
```diff
@@ -688,6 +714,11 @@ async def send_text(self, text: str) -> None:
             )
         )
         # Prompt the model to generate audio output.
+        # Guard: if a response is already in flight, defer the create rather than
+        # firing unconditionally — mirrors the recv_audio user-audio branch guard.
+        if self._response_active:
+            self._deferred_response_create = True
+            return
         await self._ws.send(json.dumps({"type": "response.create"}))
         logger.debug(
```
drewdrewthis, line 738: **[security] `send_text` logs first 60 chars of user text at DEBUG level — PII exposure**

`logger.debug("OpenAIRealtimeAgentAdapter: send_text injected %r", text[:60])` emits the start of user-supplied conversational content into application logs. If users say names, financial/health queries, or other PII, it leaks to wherever debug logs are shipped (CloudWatch, Datadog, etc.). The 60-char truncation reduces but doesn't eliminate exposure.

**Fix:** log length only:
```python
logger.debug("OpenAIRealtimeAgentAdapter: send_text injected text (%d chars)", len(text))
```

python/tests/voice/test_realtime_response_create_guard.py
```diff
@@ -0,0 +1,537 @@
+"""
+Issue #657 — regression tests for recv_audio response.create race condition.
+
```
drewdrewthis, line 3: **[test] Module docstring says "Issue #657" — should be "Issue #662"**

Line 3 of `test_realtime_response_create_guard.py` reads:
```
Issue #657 — regression tests for recv_audio response.create race condition.
```

This file covers issue #662 (send_text + JS adapter guard) in addition to the #657 recv_audio guard. The docstring should mention both issues, or at minimum not claim to be only about #657.

**Fix:** Update docstring to reference both issues, e.g.: "Issues #657 + #662 — regression tests for response.create race guards (recv_audio + send_text + JS adapters)."

python/scenario/voice/adapters/openai_realtime.py
```diff
@@ -428,7 +429,6 @@ async def recv_audio(self, timeout: float) -> AudioChunk:
                 self._agent_turn_pending = False  # user spoke → consume turn signal even when deferred
```
drewdrewthis: **[hygiene] Dead assignment — line 420 already executed unconditionally**

`self._agent_turn_pending = False` at line 429 is unreachable dead code. Line 420 sets the same flag unconditionally (outside the `if self._response_active:` branch). By the time execution reaches line 429 (inside the deferred branch), the flag was already cleared at line 420.

**Action: Fix** — delete line 429 and its comment.

javascript/src/voice/adapters/__tests__/openai-realtime-response-guard.test.ts
```diff
@@ -0,0 +1,222 @@
...
+      // response.done  → clears _responseActive, fires deferred response.create
+      // response.created → sets _responseActive = true for the new response
+      // response.output_audio.delta → the actual audio frame that resolves receiveAudio
+      // response.done  → completes the second response
+      const enqueue = (event: unknown): void => {
+        (
+          adapter as unknown as {
+            _enqueueEvent: (e: unknown) => void;
+          }
+        )._enqueueEvent(event);
```
drewdrewthis: **[test] `_enqueueEvent` white-box coupling — private method injected by cast**

AC-JS2 (and AC-ERR1 at line 208) use `(adapter as unknown as { _enqueueEvent: (e: unknown) => void })._enqueueEvent(event)` to drive server-event delivery. `_enqueueEvent` is a private implementation detail; renaming it silently breaks these tests with no type error. The test is now coupled to the class internals rather than the public WS receive interface.

**Action: Fix** — drive events through the `FakeWS` recv queue (push to `fakeWS.incoming` or equivalent) instead of calling the private method directly.

drewdrewthis:
**[proof-reviewer] Stale prove-it marker — SHA `945255e9` is not an ancestor of HEAD**

The `<!-- prove-it-clean: 945255e90d12244c2e9eecf05ea067eaf996ec8d -->` marker (comment [#4706259659](https://github.com/langwatch/scenario/issues/669/comments/4706259659)) was written at SHA `945255e9`, which is **not** an ancestor of the current HEAD `8f505f5b`. The branch was rebased after the prove-it run, invalidating the marker.

**Action: Fix** — re-run `/prove-it` on HEAD `8f505f5b` to refresh the marker and attach updated AC evidence.

drewdrewthis:
Reopening to re-trigger CI workflows on fresh SHA.

rogeriochaves:
is this the right solution to it? I mean if we guard then are we just silently dropping it? shouldn't we error soon, or is this the right approach also thinking about interruptions etc?

rogeriochaves (dismissed):

python/tests/voice/demo_response_create_guard.py
```diff
@@ -0,0 +1,278 @@
...
+    deferred_ok = adapter._deferred_response_create is True
+    item_ok    = "conversation.item.create" in sent
+
+    print()
+    print("  RESULT:")
+    print(f"    response.create suppressed?      {'YES ✓' if guard_ok else 'NO  ✗  (BUG)'}")
+    print(f"    _deferred_response_create?       {'YES ✓' if deferred_ok else 'NO  ✗  (BUG)'}")
+    print(f"    conversation.item.create sent?   {'YES ✓' if item_ok else 'NO  ✗'}")
+
+    return guard_ok and deferred_ok
```
drewdrewthis: **[proof-reviewer] Scenario C return drops `item_ok` — demo can false-pass while missing required behavior.**

`item_ok = "conversation.item.create" in sent` is computed and printed above, but dropped from the return:

```python
# current
return guard_ok and deferred_ok
# fix
return guard_ok and deferred_ok and item_ok
```

The demo is cited as the PR's primary proof artifact ("how I can prove I was successful"). A scenario that prints `conversation.item.create sent? NO ✗` while returning `True` inverts the proof — it can green-light a regression of the text-injection path.

**Action: Fix** — add `and item_ok` to the return.

javascript/src/agents/__tests__/realtime-response-guard.test.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  it("AC-JS4: handleInitialResponse does not send response.create when handler is active", async () => {
+    const transport = new FakeTransport();
+    const adapter = makeAdapter(transport);
+
+    // Make the handler believe a response is active by firing response.created
+    transport.fire("response.created");
+
+    // handleInitialResponse is private — call via any
+    const promise = (adapter as any).handleInitialResponse();
```
drewdrewthis: **[test] `(adapter as any).handleInitialResponse()` tests a private method — breaks on any rename, not a behavior contract.**

Testing a private method via `as any` couples the test to the implementation artifact (the name `handleInitialResponse`), not to the observable behavior. The same `response.create` guard can be exercised through the public `adapter.call(agentInput)` — the existing adapter test infrastructure already constructs `AgentInput`; use that path. The counterpart test file (`openai-realtime-response-guard.test.ts`) avoids this correctly via `adapter.receiveAudio()` / `adapter.sendText()`.

Same issue at line 134: `(adapter as any).handleAudioInput(b64audio)` — route through the public `call()` API instead.

**Action: Fix** — test through `adapter.call()` with a minimal `AgentInput` on both AC-JS4 and AC-JS5.

javascript/src/agents/__tests__/realtime-response-guard.test.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    // Give it a moment to call transport.sendEvent (or not)
+    await new Promise((r) => setTimeout(r, 20));
```
drewdrewthis: **[test] `setTimeout(20ms)` is timing-dependent — potential flaky test on loaded CI runners.**

The test calls a private async method, waits 20ms for the preamble to execute, then fires the completion event. On a heavily loaded CI runner, 20ms may not be enough for the microtask/macrotask queue to flush. The correct pattern is to yield deterministically — drain the microtask queue (`await Promise.resolve()` in a loop) or expose a testable hook rather than sleeping.

Same issue at line 136 in AC-JS5.

If a deterministic hook cannot be added in this PR, widen to 100ms and add a comment explaining the deliberate delay so the intent is clear.

**Action: Fix** — replace `setTimeout(r, 20)` with a deterministic yield or widen with documentation.

python/tests/voice/demo_response_create_guard.py
drewdrewthis: Fixed in commit 112db4f1: changed to `return guard_ok and deferred_ok and item_ok`.

javascript/src/agents/__tests__/realtime-response-guard.test.ts
drewdrewthis: Fixed in commit 112db4f1: widened to 100ms with explanatory comment clarifying this is a preamble-flush wait, not a timed assertion window.

javascript/src/agents/__tests__/realtime-response-guard.test.ts
drewdrewthis: Reclassifying to **Decide** — `(adapter as any)` is explicit TypeScript private-access, the coupling risk is visible at the call site. Routing through `adapter.call(AgentInput)` would require constructing a full `AgentInput` object for a narrow state check, adding significant ceremony for no coverage gain. Accepted: author should note the rename risk in a comment on the test if they prefer not to refactor.

javascript/src/voice/adapters/openai-realtime.ts
```diff
@@ -167,6 +177,17 @@ export class OpenAIRealtimeAgentAdapter extends VoiceAgentAdapter {
...
+  // --- Issue #662: response-lifecycle guard ---
+  // True while the server has an active response in progress (set on
+  // response.created, cleared on response.done / response.cancelled).
+  // Prevents double-firing response.create from the receiveAudio preamble
+  // or sendText while a response is already streaming.
+  private _responseActive = false;
+  // Set to true by receiveAudio preamble or sendText when a response.create
+  // was deferred due to _responseActive. Consumed (fired + cleared) by the
+  // response.done / response.cancelled branch in the receiveAudio loop.
+  private _deferredResponseCreate = false;
```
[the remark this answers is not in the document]
drewdrewthis: **Fixed** in commit `71f11f95`.

`disconnect()` now clears both guard flags on teardown, and `connect()` clears them **plus** `_closeReason`. The `_closeReason` reset is required: `_nextEvent` rejects immediately when it is set, so resetting only the guard flags would leave a reconnected adapter unable to receive (a cosmetic half-fix). New test `resets response-lifecycle state on disconnect and reconnects clean` asserts the flags clear on disconnect and that a reconnected instance reaches its own `receiveAudio` timeout (proving the stale close reason was cleared) rather than rejecting instantly. Full voice-adapter suite green (33/33).

javascript/src/agents/realtime/realtime-agent.adapter.ts
```diff
@@ -185,9 +185,11 @@ export class RealtimeAgentAdapter extends AgentAdapter {
...
     }
 
-    transport.sendEvent({
-      type: "response.create",
-    });
+    if (!this.eventHandler.isResponseActive()) {
+      transport.sendEvent({
+        type: "response.create",
+      });
+    }
```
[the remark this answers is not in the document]
drewdrewthis: Triaged as **out of scope for #662's agents-adapter ACs** — resolving.

AC-JS4 / AC-JS5 deliberately specify **suppression** for `RealtimeAgentAdapter` ("does not send `response.create` when active"). Defer-and-replay is the **voice** adapter's contract (AC-JS1/2/3 — AC-JS2 explicitly requires the deferred create fire after `response.done`). The two adapters were scoped differently on purpose.

- For `handleInitialResponse`, suppression is strictly correct: the active response **is** the response being awaited, so nothing is dropped.
- For `handleAudioInput`, defer-consistency with the voice adapter is a reasonable improvement, but it would (1) change the agreed AC-JS4/5 contract — their tests assert `response.create` count stays `0` even after `response.done` — and (2) rework `RealtimeEventHandler`'s resolve-on-first-`response.done` logic into defer-and-rewait. That is a contract change, not a bug fix, so it belongs in a follow-up.

The rationale and the suggested approach (mirror the voice adapter's `_deferredResponseCreate` in `RealtimeEventHandler`) are captured here for whoever picks up that follow-up.

javascript/src/voice/adapters/__tests__/openai-realtime-response-guard.test.ts
```diff
@@ -0,0 +1,232 @@
...
+      (adapter as unknown as Record<string, unknown>)._responseActive = true;
+
+      // 960 bytes of silent PCM16 (even byte count required by AudioChunk).
+      const b64audio = Buffer.from(new Uint8Array(960)).toString("base64");
+
+      const receivePromise = adapter.receiveAudio(2.0);
+
+      // Let the preamble run before injecting server events.
+      await new Promise<void>((r) => setTimeout(r, 20));
+
```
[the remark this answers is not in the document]
drewdrewthis: Triaged — addressed in commit `ff8d5fa0`.

These delays are not race-prone: `receiveAudio`'s preamble and `_nextEvent` setup are synchronous, and server events flow through an ordered queue, so each `await new Promise(r => setTimeout(r, N))` is a **macrotask boundary** that drains the loop's microtask chain regardless of magnitude or runner load (a `setTimeout` macrotask runs only after the microtask queue empties — load cannot shorten that). I reduced all three to `setTimeout(r, 0)` to drop the arbitrary magnitudes while keeping the robust drain.

Not `queueMicrotask` as suggested: a single microtask tick can under-drain a multi-tick loop continuation, whereas a zero-delay macrotask drains the whole queue. Deterministic across repeated runs (5/5 ×3).

langwatch-agent:
🐕 **PR Hound review brief** — assigned reviewer @rogeriochaves

**Review mode:** Deep Review
**Scariest prod risk:** This is a concurrency guard across four call sites in two languages. If a deferred `response.create` is never fired (flag stuck active) the agent goes silent forever; if it fires twice you get the exact "active response in progress" error this PR exists to prevent. The race window between firing the deferred create and `response.created` arriving is the crux.

- `[Must Check]` `javascript/src/voice/adapters/openai-realtime.ts` — `_responseActive` is set immediately after firing the deferred create (before `response.created` arrives) and `_deferredResponseCreate` is consumed exactly once on `response.done`/`response.cancelled`; no path leaves it set-and-never-fired.
- `[Must Check]` `realtime-event-handler.ts` `_active`/`isResponseActive()` — set on `response.created`, cleared on `response.done`, with no missed `response.cancelled`/error transition that leaves `_active` stuck true and wedges all future turns.
- `[Must Check]` `python/scenario/voice/adapters/openai_realtime.py` — `_agent_turn_pending` now cleared before the if/else; confirm both Python guards (`send_text` and `recv_audio`) mirror the JS lifecycle and can't defer without a later trigger to fire it.

- If `response.done` never arrives (transport error / dropped socket) after a deferral, is `_responseActive`/`_active` ever reset, or does the adapter deadlock on the next turn?
- The four newly-guarded call sites: is there any path that legitimately needs an unconditional `response.create` (e.g. a cancel-then-recreate) that this guard now suppresses?

Aryansharma28 (commented):
The guard logic is correct and the test coverage is thorough — hermetic mock-WS tests for all four AC sites, deferred-fire ordering proofs, and reconnect hygiene. CI is green. However, one maintainer concern blocks merge and a reconnect hazard on the Python side needs attention before ship.\n\n**[BLOCKER — unanswered design question]** `rogeriochaves` asked on 2026-06-17: *"is this the right solution? if we guard then are we just silently dropping it? shouldn't we error soon, or is this the right approach also thinking about interruptions etc?"* Their review is dismissed, not approved, and the question has no reply in thread. For `RealtimeAgentAdapter.handleInitialResponse` and `handleAudioInput`, the guard silently suppresses `response.create` but still calls `waitForResponse(timeout)`. If the in-flight response completes before the `waitForResponse` call registers its resolver, the method hangs to the 60s timeout. The `openai-realtime.ts` adapter chose the deferred-fire pattern for the same race; the `RealtimeAgentAdapter` chose suppression. The difference deserves an explicit answer from the author.\n\n**[CONCERN]** Python `connect()` and `disconnect()` do not reset `_response_active` or `_deferred_response_create` (`python/scenario/voice/adapters/openai_realtime.py` lines 305–357). The JS side resets both flags in both methods (commit `71f11f95`). A reconnected Python adapter instance that carries stale flags from a prior session will mifire on the first `recv_audio` or `send_text` — either deferring unnecessarily or firing a stale deferred create on the first `response.done` of the new session. Add the resets to `connect()` and `disconnect()` and add a Python reconnect hygiene test.\n\n**[CONCERN]** `call()` resets `_deferred_response_create = False` silently at turn start (Python line 625). If the flag is True when `call()` is invoked, the pending deferred is lost with no log or error. An assertion or warning here would surface this invariant violation early.\n\n**[NIT]** `realtime-agent.adapter.ts` (lines 146, 175) and `realtime-event-handler.ts` (lines 91, 137) use bare `console.log`/`console.error` calls — inconsistent with `openai-realtime.ts` which uses `Logger.create()`. These are pre-existing but are in files this PR touches.\n\n**[NIT]** `python/tests/voice/demo_response_create_guard.py` is a standalone 278-line script in the tests directory. It is not pytest-discoverable and duplicates coverage already in `test_realtime_response_create_guard.py`. Consider moving to `python/examples/` or removing.\n\nGenuine praise for the `_MockWS.log` interleaved chronological ordering design and the `wsFactory` testability seam — both are patterns worth copying elsewhere.

Aryansharma28 (commented):
CI fully green. The PR correctly guards all four previously unguarded `response.create` call sites across JS and Python, with solid hermetic test coverage.

**[BLOCKER]** An open design question from `rogeriochaves` (maintainer, 2026-06-17) on the guard flag reset semantics remains unanswered. This needs a resolution before merge — either address the concern or get explicit sign-off that the current approach is acceptable.

**[CONCERN]** Asymmetry between JS and Python: Python `connect()`/`disconnect()` does not reset the new guard flags, unlike the JS side. On reconnect, the Python adapter will behave as if `response.create` was already called (guard flag still set from the previous session), which could cause a silent failure to send the first `response.create` after reconnect. Mirror the JS reset logic in Python's `disconnect()`.

**[NIT]** A few guard-flag variable names differ between JS (`_responseCreateSent`) and Python (`_response_create_sent` is fine — snake_case is correct) but the conceptual naming is slightly inconsistent across call sites.

Leaving as COMMENT (not requesting changes) since the blocker is a maintainer design question rather than a code defect. Once `rogeriochaves` signs off, this is close to ready.

python/scenario/voice/adapters/openai_realtime.py
drewdrewthis: Resolved — fixed in `d4642fc` (post-rebase SHA; branch rebased onto `origin/main`, CONFLICTING→MERGEABLE).

`self._agent_turn_pending = False` now runs **unconditionally** at `openai_realtime.py:420`, before the `if self._response_active:` / `else` split — so the deferred branch no longer leaves the flag set for a later `recv_audio` re-entry to spuriously fire off. Proven by `test_deferred_path_clears_agent_turn_pending` (green: 12/12 in the guard suite).

javascript/src/voice/adapters/openai-realtime.ts
drewdrewthis: Resolved — fixed in `d4642fc` (post-rebase SHA).

After the deferred `response.create` fires, `openai-realtime.ts:508` sets `_responseActive = true` **immediately** (`// mirror Python: close the window before response.created arrives`). The window in which a concurrent `sendText` saw `_responseActive === false` and double-fired is closed — mirrors the Python pattern exactly.

python/scenario/voice/adapters/openai_realtime.py
drewdrewthis: Resolved — fixed in `d4642fc` (post-rebase SHA).

The DEBUG log no longer emits user text. `openai_realtime.py:724-726` now logs only the length:
```python
logger.debug(
    "OpenAIRealtimeAgentAdapter: send_text injected text (%d chars)", len(text)
)
```
No conversational content reaches application logs — the PII-exposure vector is gone.

python/tests/voice/test_realtime_response_create_guard.py
drewdrewthis: Resolved — fixed in `d4642fc` (post-rebase SHA).

The module docstring now names both issues (`test_realtime_response_create_guard.py:2`):
> Issues #657 + #662 — regression tests for response.create race guards (recv_audio + send_text + JS adapters).

python/scenario/voice/adapters/openai_realtime.py
drewdrewthis: Resolved — fixed in `88a7d85` (post-rebase SHA).

The unreachable `self._agent_turn_pending = False` inside the deferred branch is deleted. The flag is now cleared exactly once, unconditionally, at `openai_realtime.py:420` — no dead assignment remains (`grep -n '_agent_turn_pending = False'` → lines 420 + 444, both live).

javascript/src/voice/adapters/__tests__/openai-realtime-response-guard.test.ts
drewdrewthis: Resolved — fixed in `88a7d85` (post-rebase SHA).

The tests no longer reach into the private `_enqueueEvent`. Server-event delivery is driven through an injected `wsFactory` fake WS (public constructor seam) — see the `wsFactory:` fake in `openai-realtime-response-guard.test.ts`. `grep -c _enqueueEvent` on that file → **0**, so a rename can no longer silently break the suite.

javascript/src/agents/__tests__/realtime-response-guard.test.ts
```diff
@@ -0,0 +1,144 @@
...
+
```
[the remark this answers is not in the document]
drewdrewthis: Resolved — fixed in `c13ae25` then finalized in `d885abb` (post-rebase SHAs).

The fixed-magnitude `setTimeout` sleeps are gone from `realtime-response-guard.test.ts` (`grep -c setTimeout` → **0**). The suite now synchronizes deterministically on the microtask/macrotask queue (zero-delay macrotask yields) rather than wall-clock elapsed time, so it can't flake under a loaded runner.

python/tests/voice/demo_response_create_guard.py
```diff
@@ -0,0 +1,278 @@
...
+
```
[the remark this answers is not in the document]
drewdrewthis: Resolved — fixed in `3d7aa91` (post-rebase SHA).

Scenario C now folds the `item_ok` invariant into its verdict (`demo_response_create_guard.py:244`):
```python
return guard_ok and deferred_ok and item_ok
```
A missing `conversation.item.create` can no longer false-pass the demo — which is cited as the PR's primary proof artifact.

drewdrewthis:
Slack PR review request posted.

rogeriochaves (approved):

PR #676, opened 2026-06-15
feat(voice): realtime_langwatch_session context manager for live OpenAI Realtime apps (#673)

### LINKED ISSUE #673: feat(voice): first-class LangWatch tracing for live OpenAI Realtime apps (not just scenario tests)
## For Humans

**OpenAI Realtime voice apps get rich LangWatch traces during scenario tests but zero tracing guidance for production.** Developers who follow the happy-path-openai-realtime doc will expect the same observability in their live app — but there is no documented path, no helper, and no auto-instrumentor for the raw WebSocket. The current workaround (manually reverse-engineering `langwatch.trace()` + per-turn `langwatch.span()` calls) forces developers to guess the span shape that scenario emits automatically in tests.

**Ask:** Add a tested, importable `realtime_langwatch_session()` context manager (Option 2) + a "Live app tracing" docs section (Option 3). Investigation finds Option 1 (full instrumentor) is too large for P3-low and carries hidden protocol-stability risks.

_Status: investigating → planning. Full spec below._

## Investigation

**Root finding: the test trace is provided by scenario's scaffolding, not the adapter — so a live app has nothing to call.** `setup_scenario_tracing()` calls `langwatch.setup()`, builds the `TracerProvider`, and adds the OTLP exporter, all wired inside `scenario.run()`. The `OpenAIRealtimeAgentAdapter` itself creates zero spans — no `get_tracer`, no `start_span` anywhere in `scenario/voice/`. A developer building a live Realtime app gets none of this scaffolding.

### Symptoms / context

The happy-path-openai-realtime page targets developers whose agent IS an OpenAI Realtime model. After following that guide for their test suite, they naturally expect the same LangWatch traces in production. But:

- `setup_scenario_tracing()` is triggered only by `scenario.run()` — no standalone callable for live apps.
- `OpenAIRealtimeAgentAdapter` (`scenario/voice/adapters/openai_realtime.py`) is test-infrastructure: connects `wss://api.openai.com/v1/realtime`, handles audio I/O, transcripts, tool calls — and emits zero spans.
- The langwatch Python SDK (`langwatch==0.2.19`) has no `OpenAIRealtimeInstrumentor`.
- Manual path exists but is undocumented: `langwatch.setup()` + `with langwatch.trace():` + `with langwatch.span(type="llm"):` per turn, with no guidance on which fields to populate.

### Root cause / key findings

1. **`setup_scenario_tracing()` is test-runner-only** — `_full_setup` (`setup.py:161`) calls `langwatch.setup(skip_open_telemetry_setup=True)`, creates a `TracerProvider`, attaches the OTLP exporter to `{LANGWATCH_ENDPOINT}/api/otel/v1/traces` (`setup.py:241`), and runs instrumentors. All triggered by `ensure_tracing_initialized()` inside `scenario.run()`. No exported standalone callable for a live app.

2. **Zero spans in the voice adapter** — `scenario/voice/adapters/openai_realtime.py`: no `get_tracer`, no `start_as_current_span`, no `start_span`. The only span-creation sites in the whole package are in `red_team_agent.py` (lines 36, 582, 914) — simulator/judge side only.

3. **`_add_langwatch_exporter` is the reusable building block** — `setup.py:218-252`: standalone function that needs `LANGWATCH_API_KEY` + `LANGWATCH_ENDPOINT` and attaches an OTLP `BatchSpanProcessor` to any given `TracerProvider`. This is the right primitive for a live-app helper.

4. **Wire protocol still evolving** — The adapter defensively handles both `response.audio.delta` (pre-GA) and `response.output_audio.delta` (GA) with one-time warnings (issue #602). Protocol is live-patching.

5. **`_get_concrete_provider()` returns `None` for a fresh process** — verified: importing `scenario` gives a `ProxyTracerProvider`; calling `_get_concrete_provider()` on it returns `None` until `langwatch.setup()` is called. The helper must handle both states (no prior setup vs. prior setup).

### Challenge findings (Phase 6.5)

**Option 1 (full `OpenAIRealtimeInstrumentor`) hidden assumptions that weaken it:**

1. **GA protocol is stable** — The adapter code actively patches pre-GA event names. An instrumentor shipping now must handle both aliases plus any future renames. Large maintenance surface for P3-low.

2. **"Test trace shape" is cheaply reproducible** — The test trace shape comes from `judge_span_collector`, `FilteringSpanExporter`, and `scenario.run()` orchestration scaffolding — not the adapter. Matching it in a live app requires re-exposing substantial scaffolding. "Identical shapes" is a bigger claim than it appears.

3. **Turn boundaries are instrumentable without VAD state** — Scenario tests disable server-side VAD (`turn_detection: None`). Live apps may use server VAD, making turn boundaries fuzzy. An instrumentor must track `response.created` / `response.done` pairs, handle multi-response turns, interrupted responses, and function-call-only turns — essentially reimplementing the adapter's existing non-trivial state machine.

4. **Package ownership** — An `OpenAIRealtimeInstrumentor` more naturally belongs in the `langwatch` package (alongside other instrumentors) than in `langwatch-scenario`.

**Option 2 does NOT have these weaknesses:** A `realtime_langwatch_session()` context manager calls `langwatch.trace()` / `langwatch.span()` with sensible defaults — no wire protocol knowledge required. ~60-80 lines, zero protocol coupling, testable with standard mocks.

### Strategies considered

| Option | Scope | Protocol coupling | Package fit | Recommended |
|--------|-------|-------------------|-------------|-------------|
| 1. Full `OpenAIRealtimeInstrumentor` | 3-5 days | High (must track GA rollout) | Questionable (prod obs in test pkg) | No |
| 2. `realtime_langwatch_session()` helper | ~4-6h | None | Fits scenario as convenience export | **Yes** |
| 3. Docs section only | ~1-2h | None | Any doc page | Yes (ships with 2) |

- **`_add_langwatch_exporter` is the building block** (`setup.py:218-252`): standalone, needs `LANGWATCH_API_KEY` + `LANGWATCH_ENDPOINT`. A helper that exposes a version of this without the full `_do_setup` machinery is the minimal live-app on-ramp.
- **The helper must NOT call `setup_scenario_tracing()`** — that mutates global OTel state and sets `_initialized = True`, conflicting if the developer also runs scenario tests.
- **LangWatch LLM span shape**: `langwatch.span(type="llm")` with attributes `input`, `output`, `model`, `params.temperature`, `usage.prompt_tokens`, `usage.completion_tokens`.
- **`_get_concrete_provider()` state machine** (`setup.py:91`): returns `None` on a fresh process (proxy provider only), returns the concrete `TracerProvider` after `langwatch.setup()`. The helper must create a new provider on `None`, or attach on concrete.
- **`_initialized` global coupling** (`setup.py:25`): the module-level bool gates `setup_scenario_tracing()` and `ensure_tracing_initialized()`. The helper must not set or read this flag — if it does, helper-then-run order can cause `ensure_tracing_initialized()` to skip setup entirely (AC11).

### Caveats

- Verify the `langwatch.trace()` / `langwatch.span()` API against the version pinned in `python/pyproject.toml` before coding — SDK internals may have moved since `0.2.19`.
- The `BatchSpanProcessor` export is async — AC6 failure injection must be synchronous (mock `on_end`) or use `force_flush()`.

### Core behavior

- [ ] **AC1**: `scenario.realtime_langwatch_session` is the public importable name. Evidence: `python3 -c "from scenario import realtime_langwatch_session"` exits 0.

- [ ] **AC2**: Calling `session.log_turn(user_transcript: str, agent_transcript: str, model: str, latency_ms: int)` inside the context emits a child span with: `name` = `"realtime_turn"`, `type` attribute = `"llm"`, `input` = user transcript, `output` = agent transcript, `model` = model string, `latency_ms` = integer. Evidence: unit test installs `InMemorySpanExporter`, calls `log_turn`, asserts `len(exporter.get_finished_spans()) == 1`, `span.name == "realtime_turn"`, and each attribute value matches.

- [ ] **AC3**: When `LANGWATCH_API_KEY` is absent (unset from env and not passed as `api_key`), entering the context and calling `log_turn()` raise no errors and an `InMemorySpanExporter` attached to the global provider records zero finished spans. Evidence: unit test asserts no exception and `len(exporter.get_finished_spans()) == 0`.

- [ ] **AC4**: The helper's source file has no import from `scenario._tracing.setup`. Evidence: `test -f scenario/_tracing/live.py && ! grep -nE '^(from|import) .*_tracing\.setup' scenario/_tracing/live.py` exits 0; a unit test that runs `realtime_langwatch_session()` with no prior `scenario.run()` call does not raise.

- [ ] **AC5a**: When no prior `langwatch.setup()` has been called (fresh OTel state, `ProxyTracerProvider`), entering `realtime_langwatch_session()` creates a new `TracerProvider` registered as the global provider. Evidence: unit test — reset OTel state, enter context, assert `isinstance(trace.get_tracer_provider(), TracerProvider)`.

- [ ] **AC5b**: When `langwatch.setup()` has already been called (existing concrete `TracerProvider`), entering `realtime_langwatch_session()` does NOT call `trace.set_tracer_provider()` again — the provider `id()` before and after entering the context is identical. Evidence: unit test — call `langwatch.setup()`, capture `id(trace.get_tracer_provider())`, enter helper, assert same `id`; an `InMemorySpanExporter` added to the existing provider captures spans emitted inside the context.

- [ ] **AC13**: Two sequential `log_turn(...)` calls inside one open context emit exactly two finished spans, each carrying its own `input`/`output`/`latency_ms`. Evidence: unit test — `len(exporter.get_finished_spans()) == 2`; first span attributes match first call args; second span attributes match second call args.

- [ ] **AC14**: On `__aexit__`, the helper flushes spans and does not leave a half-configured global provider. Evidence: unit test asserts span emitted by `log_turn` is present in `get_finished_spans()` after the `async with` block closes (not only mid-block); `__aexit__` does not raise even if no turns were logged.

### Failure modes

- [ ] **AC6**: If the LangWatch OTLP export fails (mock `on_end()` to raise `RuntimeError`), the live app continues unaffected — the exception is caught, logged at WARNING via the `scenario.tracing` logger, and not re-raised. Evidence: unit test asserts surrounding code completes and `caplog.records` contains at least one WARNING from logger `"scenario.tracing"`.

- [ ] **AC7**: Calling `log_turn()` before `__aenter__` OR after `__aexit__` raises `RuntimeError` with a message containing `"realtime_langwatch_session"`. Evidence: two unit tests — one calls before entering, one after exiting — both assert the expected error.

- [ ] **AC12**: `log_turn()` with `user_transcript=""`, `agent_transcript=""`, or `latency_ms=0` does not raise. Evidence: unit test with empty strings and zero latency asserts no exception and one span is recorded.

### Docs

- [ ] **AC8**: `docs/docs/pages/voice/happy-path-openai-realtime.mdx` gains a "Getting LangWatch traces from your live app" section containing a fenced code block using `realtime_langwatch_session`. Evidence: `awk '/^## Getting LangWatch traces from your live app/{f=1} /^## /{if(f&&!/Getting LangWatch/)exit} f' docs/docs/pages/voice/happy-path-openai-realtime.mdx | grep -c 'async with realtime_langwatch_session('` returns ≥1; the same awk-extracted block piped to `grep -oE '[A-Z_]+_KEY' | sort -u` returns exactly `LANGWATCH_API_KEY` and `OPENAI_API_KEY`.

- [ ] **AC9** (import regression): Importing `scenario` does NOT call `langwatch.setup()` or create a `TracerProvider` at import time. Evidence: `python3 -c "import scenario; from opentelemetry import trace; p = trace.get_tracer_provider(); assert type(p).__name__ == 'ProxyTracerProvider'"` exits 0.

- [ ] **AC10** (run-then-helper): A test that calls `ensure_tracing_initialized()` then enters `realtime_langwatch_session()` does not raise. `id(trace.get_tracer_provider())` is the same before and after entering the helper. Evidence: unit test asserting same `id()` before and after.

- [ ] **AC11** (helper-then-run): A test that enters and exits `realtime_langwatch_session()` then calls `ensure_tracing_initialized()` does not raise, `_initialized` is True after the call, and subsequent `scenario.run()` proceeds without OTel conflict (returns a `ScenarioResult` or completes the run body without raising). Evidence: unit test asserts no exception; imports `_initialized` from `scenario._tracing.setup` and asserts it is True.

## Plan

**Approach:** Build a `RealtimeLangWatchSession` async context manager in a new module `scenario/_tracing/live.py`. It wraps `langwatch.trace()` / `langwatch.span()` with sensible defaults for OpenAI Realtime apps — no wire protocol knowledge required. Export it from `scenario/__init__.py` as `realtime_langwatch_session`. Companion: add a "Getting LangWatch traces from your live app" section to `docs/docs/pages/voice/happy-path-openai-realtime.mdx`.

**Sequencing:**

1. **Phase 1 — API audit**: Read the installed `langwatch` package's `trace()` and `span()` interfaces (version in `python/pyproject.toml`) and confirm the LLM span attribute names. Completion signal: exact `langwatch.span()` call recorded in a code comment before coding starts.

2. **Phase 2 — `scenario/_tracing/live.py`**: Write `RealtimeLangWatchSession` with `__aenter__`/`__aexit__` and `log_turn(user_transcript, agent_transcript, model, latency_ms)`. Logic: on enter, call `_get_concrete_provider()` — if `None` (fresh process) create + register new `TracerProvider` with `_add_langwatch_exporter`; if concrete, attach `_add_langwatch_exporter` to existing provider. No import from `scenario._tracing.setup`. No-op when `LANGWATCH_API_KEY` absent. On exit, flush/shutdown any processor IT added. Completion signal: module exists with class.

3. **Phase 3 — `scenario/__init__.py` export**: Add `from scenario._tracing.live import realtime_langwatch_session` (or expose the class under that name). Completion signal: `python3 -c "from scenario import realtime_langwatch_session"` exits 0.

4. **Phase 4 — unit tests**: `python/tests/tracing/test_live_tracing.py`. Cover AC1-AC14 using `InMemorySpanExporter` + `SimpleSpanProcessor`. No live API calls. Completion signal: `uv run pytest python/tests/tracing/test_live_tracing.py` passes.

5. **Phase 5 — docs**: Add "Getting LangWatch traces from your live app" section to `docs/docs/pages/voice/happy-path-openai-realtime.mdx` with a minimal code snippet. Completion signal: AC8 awk grep passes.

**Key decisions:**

- **New `live.py` module, not added to `setup.py`**: keeps test-runner concerns (setup.py, `_initialized` global) separate from live-app concerns. AC4 explicitly prohibits importing from setup.
- **`_add_langwatch_exporter` called from `live.py` directly** (not via `_do_setup`): avoids the `_initialized` side effect that would break helper-then-run ordering (AC11).
- **Async context manager (not sync)**: Realtime apps are async; sync would force `asyncio.run()` wrapping that conflicts with a running event loop.
- **No-op on missing API key**: matches the existing pattern in `_add_langwatch_exporter` (`setup.py:224`) — silence is preferable to raising in app startup.
- **`log_turn()` as explicit method, not automatic per-event**: avoids wire protocol coupling (key Challenge finding). The developer calls it once per agent turn after collecting the transcript.
- **Span name `"realtime_turn"`**: fixed name, consistent across all turns, queryable in LangWatch.

**Risks / open questions:**

- `_add_langwatch_exporter` is a private function in `scenario._tracing.setup` — `live.py` may need to duplicate its internals or call it via a clean internal API. Decide at Phase 2: if the function is small enough, inline it; if not, move it to `scenario._tracing._exporters.py`.
- The `docs/` directory may be a separate repo — confirm at Phase 5. If so, AC8 requires a separate PR to `langwatch/langwatch` (docs site).
- `BatchSpanProcessor` makes export async; AC6 failure injection must mock `on_end()` synchronously to observe the failure before the test exits.

**Out of scope:**

- Option 1: Full `OpenAIRealtimeInstrumentor`.
- Auto-instrumentation of the OpenAI Realtime WebSocket itself.
- Any changes to `OpenAIRealtimeAgentAdapter` — the helper is standalone.
- Support for non-OpenAI realtime WebSocket APIs.
- Moving `_initialized` to a class-based design (flagged for principles-reviewer; out of scope here).

## Evidence

_To be filled in by the proof step._

</details>

drewdrewthis:
## Why

Production OpenAI Realtime apps have no path into LangWatch today — `scenario.run()` handles test-time tracing, but live calls fly blind. Closes #673.

## What changed

- **`RealtimeLangWatchSession` (exported as `realtime_langwatch_session`)** — async context manager that wraps a live Realtime session and records one child LLM span per `log_turn()` call, matching the per-turn shape scenario tests already produce.
- **No-op when keyless** — if `LANGWATCH_API_KEY` is absent the context manager is a safe no-op (no raise, no spans); safe to ship in keyless envs without a conditional.
- **Attaches to existing provider, doesn't clobber** — if `langwatch.setup()` or `scenario.run()` already installed a `TracerProvider`, the helper attaches to it; idempotent guard prevents duplicate `BatchSpanProcessor` registration across sequential sessions.
- **Spec-first, unit-tested** — 14 ACs in `specs/realtime-live-tracing.feature`, covered by 20 unit tests in `tests/test_live_tracing.py`; no live API calls needed.

## How it works

```
scenario/
  __init__.py              ← exports realtime_langwatch_session
  _tracing/
    live.py                ← RealtimeLangWatchSession + helpers (_get_concrete_provider, _add_langwatch_exporter)
specs/
  realtime-live-tracing.feature   ← 14 ACs
tests/
  test_live_tracing.py            ← 20 unit tests
```

`__aenter__` checks the global OTel provider: fresh → builds + installs its own `TracerProvider`; existing → attaches an OTLP exporter. A root span is opened for the session; `log_turn()` creates a child `realtime_turn` span with `type=llm`, `input`, `output`, `model`, `latency_ms`. `__aexit__` ends the root span and force-flushes only provider it owns.

## Test plan

```bash
cd python
uv run pytest tests/test_live_tracing.py -v
# → 20 passed
uv run pyright tests/test_live_tracing.py scenario/_tracing/live.py
# → 0 errors
```

No playable artifact — see Test plan. 20 unit tests pass covering all 14 ACs; pyright clean; all CI checks green.

## Anything surprising?

The live module intentionally duplicates two small helpers from `scenario._tracing.setup` (`_get_concrete_provider`, `_add_langwatch_exporter`). The design constraint is that `scenario._tracing.setup` is the test-runner's lazy-init path — importing it from a production context would be wrong. The duplication is documented in both files.

### DIFF
--- a/python/scenario/_tracing/live.py
+++ b/python/scenario/_tracing/live.py
@@ -132,7 +132,6 @@ class RealtimeLangWatchSession:
         self._api_key = api_key or os.environ.get("LANGWATCH_API_KEY")
 
         # Lifecycle flags.
-        self._entered = False  # has __aenter__ run at least once?
         self._active = False  # is the context currently open (guards log_turn)?
         self._noop = False  # no key → emit nothing, but still guard log_turn
 
@@ -144,8 +143,6 @@ class RealtimeLangWatchSession:
         self._ctx_token = None
 
     async def __aenter__(self) -> "RealtimeLangWatchSession":
-        self._entered = True
-
         if not self._api_key:
             # No-op mode: the context is "active" (so log_turn does not raise the
             # not-active RuntimeError), but it emits nothing.
@@ -170,6 +167,8 @@ class RealtimeLangWatchSession:
 
             self._tracer = trace.get_tracer("scenario.realtime")
             self._root_span = self._tracer.start_span(self._name)
+            if self._model is not None:
+                self._root_span.set_attribute("model", self._model)
             self._ctx_token = context.attach(
                 trace.set_span_in_context(self._root_span)
             )
@@ -226,7 +225,7 @@ class RealtimeLangWatchSession:
                 exc_info=True,
             )
 
-    async def __aexit__(self, exc_type, exc, tb) -> bool:
+    async def __aexit__(self, exc_type, exc, tb) -> None:
         # Guard first so log_turn raises if called after exit.
         self._active = False
 
@@ -255,5 +254,3 @@ class RealtimeLangWatchSession:
                 "Failed to flush realtime_langwatch_session provider.", exc_info=True
             )
 
-        # Never suppress an exception raised inside the with-block.
-        return False
--- a/python/scenario/_tracing/live.py
+++ b/python/scenario/_tracing/live.py
@@ -1,0 +1,259 @@
+"""LangWatch tracing for live (production) OpenAI Realtime apps.
+
+Scenario tests get rich LangWatch traces automatically via ``scenario.run()``,
+but a production Realtime app gets nothing by default. ``RealtimeLangWatchSession``
+(exported as ``realtime_langwatch_session``) wraps a live Realtime session in the
+same per-turn span shape so live traffic lands in the same LangWatch project,
+queryable alongside scenario-test traces.
+
+Design constraints (issue #673):
+
+- This module MUST NOT import from ``scenario._tracing.setup``. The setup module
+  is the test-runner's lazy initialization path; the live helper has to stand on
+  its own so it works in a process where ``scenario.run()`` was never called. The
+  two small helpers it needs (``_get_concrete_provider`` /
+  ``_add_langwatch_exporter``) are therefore intentionally duplicated inline.
+- Importing ``scenario`` must NOT create a ``TracerProvider`` — all OTel work
+  happens inside ``__aenter__``, never at import time.
+- Nothing in the live path may raise into the host application. Export failures
+  are swallowed and logged at WARNING.
+"""
+
+import logging
+import os
+from typing import Optional
+
+from opentelemetry import context, trace
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import SpanExporter
+
+logger = logging.getLogger("scenario.tracing")
+
+
+def _get_concrete_provider(provider) -> Optional[TracerProvider]:
+    """Return the concrete ``TracerProvider`` if one exists, else ``None``.
+
+    Checks the provider itself and one level of delegation (the
+    ``ProxyTracerProvider`` pattern OTel uses before any provider is set).
+
+    NOTE: intentionally duplicated from ``scenario._tracing.setup`` — see the
+    module docstring for why this file must not import from setup.
+    """
+    if isinstance(provider, TracerProvider):
+        return provider
+
+    delegate = None
+    if hasattr(provider, "get_delegate"):
+        delegate = provider.get_delegate()
+    elif hasattr(provider, "_delegate"):
+        delegate = provider._delegate
+
+    if isinstance(delegate, TracerProvider):
+        return delegate
+
+    return None
+
+
+def _add_langwatch_exporter(provider: TracerProvider, api_key: str) -> None:
+    """Attach the LangWatch OTLP exporter to ``provider``.
+
+    Best-effort: if the OTLP HTTP exporter dependency is missing we log a
+    warning and return rather than raising into the host app.
+
+    NOTE: intentionally duplicated/simplified from ``scenario._tracing.setup``
+    — see the module docstring for why this file must not import from setup.
+    """
+    endpoint = os.environ.get("LANGWATCH_ENDPOINT", "https://app.langwatch.ai")
+
+    try:
+        from opentelemetry.exporter.otlp.proto.http.trace_exporter import (
+            OTLPSpanExporter,
+        )
+    except ImportError:
+        logger.warning(
+            "opentelemetry-exporter-otlp-proto-http not installed. "
+            "LangWatch span export disabled for realtime_langwatch_session."
+        )
+        return
+
+    from opentelemetry.sdk.trace.export import BatchSpanProcessor
+
+    exporter: SpanExporter = OTLPSpanExporter(
+        endpoint=f"{endpoint}/api/otel/v1/traces",
+        headers={"Authorization": f"Bearer {api_key}"},
+    )
+    provider.add_span_processor(BatchSpanProcessor(exporter))
+
+
+class RealtimeLangWatchSession:
+    """Async context manager that traces a live OpenAI Realtime session.
+
+    Wrap your production Realtime loop in ``async with`` and call
+    :meth:`log_turn` after each agent turn::
+
+        from scenario import realtime_langwatch_session
+
+        async with realtime_langwatch_session(
+            name="my-production-session",
+            model="gpt-4o-realtime-preview",
+        ) as session:
+            ...  # drive your raw Realtime WebSocket
+            await session.log_turn(
+                user_transcript="I want to cancel my subscription",
+                agent_transcript="Of course, I can help with that.",
+                model="gpt-4o-realtime-preview",
+                latency_ms=430,
+            )
+
+    Behavior:
+
+    - **No key → no-op.** If ``LANGWATCH_API_KEY`` is absent (and no ``api_key``
+      is passed), entering the context and calling :meth:`log_turn` produce zero
+      spans and never raise. Safe to ship in keyless environments.
+    - **Fresh process.** If no concrete ``TracerProvider`` exists yet, one is
+      created, wired to LangWatch, and installed globally; it is force-flushed on
+      exit so spans are exported before the block returns.
+    - **Existing provider.** If ``langwatch.setup()`` (or anything else) already
+      installed a concrete provider, the helper attaches its exporter to that
+      provider and leaves it in place — no duplicate initialization.
+    - **Never raises into the host app.** Per-turn span failures are caught and
+      logged at WARNING; ``__aexit__`` never raises.
+    """
+
+    def __init__(
+        self,
+        name: str = "realtime_session",
+        *,
+        model: Optional[str] = None,
+        api_key: Optional[str] = None,
+    ) -> None:
+        self._name = name
+        self._model = model
+        self._api_key = api_key or os.environ.get("LANGWATCH_API_KEY")
+
+        # Lifecycle flags.
+        self._entered = False  # has __aenter__ run at least once?
+        self._active = False  # is the context currently open (guards log_turn)?
+        self._noop = False  # no key → emit nothing, but still guard log_turn
+
+        # OTel handles, populated in __aenter__.
+        self._provider: Optional[TracerProvider] = None
+        self._own_provider = False  # did WE create/install the provider?
+        self._tracer = None
+        self._root_span = None
+        self._ctx_token = None
+
+    async def __aenter__(self) -> "RealtimeLangWatchSession":
+        self._entered = True
+
+        if not self._api_key:
+            # No-op mode: the context is "active" (so log_turn does not raise the
+            # not-active RuntimeError), but it emits nothing.
+            self._noop = True
+            self._active = True
+            return self
+
+        try:
+            concrete = _get_concrete_provider(trace.get_tracer_provider())
+            if concrete is None:
+                # Fresh process — build and install our own provider.
+                provider = TracerProvider()
+                _add_langwatch_exporter(provider, self._api_key)
+                trace.set_tracer_provider(provider)
+                self._provider = provider
+                self._own_provider = True
+            else:
+                # Attach to the provider someone else already installed.
+                _add_langwatch_exporter(concrete, self._api_key)
+                self._provider = concrete
+                self._own_provider = False
+
+            self._tracer = trace.get_tracer("scenario.realtime")
+            self._root_span = self._tracer.start_span(self._name)
+            self._ctx_token = context.attach(
+                trace.set_span_in_context(self._root_span)
+            )
+        except Exception:  # noqa: BLE001 - never raise into the host app
+            logger.warning(
+                "Failed to start realtime_langwatch_session root span; "
+                "continuing as a no-op.",
+                exc_info=True,
+            )
+            self._noop = True
+
+        self._active = True
+        return self
+
+    async def log_turn(
+        self,
+        user_transcript: str,
+        agent_transcript: str,
+        model: str,
+        latency_ms: float,
+    ) -> None:
+        """Record one agent turn as a child LLM span under the root span.
+
+        Raises:
+            RuntimeError: if called before ``__aenter__`` or after ``__aexit__``.
+                In no-op mode (no API key) this is a silent no-op instead.
+        """
+        if not self._active:
+            raise RuntimeError(
+                "realtime_langwatch_session context is not active — call "
+                "async with realtime_langwatch_session() as session first"
+            )
+
+        if self._noop:
+            return
+
+        try:
+            assert self._tracer is not None  # for type-checkers; guarded by _noop
+            ctx = (
+                trace.set_span_in_context(self._root_span)
+                if self._root_span is not None
+                else None
+            )
+            span = self._tracer.start_span("realtime_turn", context=ctx)
+            span.set_attribute("type", "llm")
+            span.set_attribute("input", user_transcript)
+            span.set_attribute("output", agent_transcript)
+            span.set_attribute("model", model)
+            span.set_attribute("latency_ms", latency_ms)
+            span.end()
+        except Exception:  # noqa: BLE001 - export/processor failures must not propagate
+            logger.warning(
+                "Failed to record realtime turn span; dropping it.",
+                exc_info=True,
+            )
+
+    async def __aexit__(self, exc_type, exc, tb) -> bool:
+        # Guard first so log_turn raises if called after exit.
+        self._active = False
+
+        try:
+            if self._root_span is not None:
+                self._root_span.end()
+        except Exception:  # noqa: BLE001
+            logger.warning(
+                "Failed to end realtime_langwatch_session root span.", exc_info=True
+            )
+
+        try:
+            if self._ctx_token is not None:
+                context.detach(self._ctx_token)
+        except Exception:  # noqa: BLE001
+            logger.warning(
+                "Failed to detach realtime_langwatch_session context.", exc_info=True
+            )
+
+        # Force-flush only the provider WE own, so spans export before returning.
+        try:
+            if self._own_provider and self._provider is not None:
+                self._provider.force_flush()
+        except Exception:  # noqa: BLE001
+            logger.warning(
+                "Failed to flush realtime_langwatch_session provider.", exc_info=True
+            )
+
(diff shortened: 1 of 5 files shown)


drewdrewthis:
## Review verdict: READY

**All CI checks pass. Both Fix concerns resolved. All 14 ACs covered.** PR is assigned and human reviewer requested.

---

### Fixes applied since `0669c9db`

1. **[security]** `python/scenario/_tracing/live.py:182` — Removed `exc_info=True` from `__aenter__` exception handler; now logs `type(exc).__name__` only, eliminating the Bearer token exposure path. ✅ Resolved
2. **[test]** `python/tests/test_live_tracing.py:124-130` — Added parent-child span assertion: `parent_ctx.span_id == root_ctx.span_id`. Pyright-narrowed both `SpanContext | None` values before comparison. ✅ Resolved

---

### Non-blocking (for author awareness)

- **Design** `[principles][uncle-bob][metz-beck][fowler]` — `_noop`/`_active`/`_own_provider` flags encode three states; Null Object split recommended as follow-up issue
- **Dead assert** `[principles]` — `assert self._tracer is not None` at `live.py:216` is guarded by `_noop` and dead under `python -O`; delete or annotate
- **Snake_case export** `[principles]` — `RealtimeLangWatchSession as realtime_langwatch_session` convention mismatch; consider factory function
- **`reset_otel` fixture** `[hygiene]` — missing `trace._PROXY_TRACER_PROVIDER` reset; could leak state between tests
- **`import re` placement** `[hygiene]` — inside function bodies at lines 158, 354; move to module level
- **Span schema** `[fowler]` — `log_turn` hard-codes LangWatch attribute keys; verify they match scenario-test span constants

---

### Completeness (issue #673)

All 14 ACs (AC1–AC14, with AC5 split a/b) covered by tests. Proof-reviewer ran the suite: `20 passed in 2.84s`.

drewdrewthis:
## Live behavior demonstration

`realtime_langwatch_session` exercised against the installed package (`uv run python3`) with an `InMemorySpanExporter` to capture spans without requiring a live LangWatch server:

```
$ LANGWATCH_API_KEY=test-demo-key uv run python3 -c "
  ...
  async with realtime_langwatch_session(name='demo-session', model='gpt-4o-realtime-preview') as session:
      await session.log_turn('I want to cancel my subscription', 'Of course, I can help with that.', model='gpt-4o-realtime-preview', latency_ms=430)
      await session.log_turn('What is the refund policy?', 'Full refund within 30 days.', model='gpt-4o-realtime-preview', latency_ms=310)
  spans = exporter.get_finished_spans()
  for i, s in enumerate(spans): print(f'[{i}] name={s.name!r} attrs={dict(s.attributes)}')
"

=== realtime_langwatch_session demo ===
Spans produced: 3
  [0] name='realtime_turn' attrs={'type': 'llm', 'input': 'I want to cancel my subscription', 'output': 'Of course, I can help with that.', 'model': 'gpt-4o-realtime-preview', 'latency_ms': 430}
  [1] name='realtime_turn' attrs={'type': 'llm', 'input': 'What is the refund policy?', 'output': 'Full refund within 30 days.', 'model': 'gpt-4o-realtime-preview', 'latency_ms': 310}
  [2] name='demo-session' attrs={'model': 'gpt-4o-realtime-preview'}
```

Confirmed:
- Two child `realtime_turn` spans with correct `type`, `input`, `output`, `model`, `latency_ms` attributes
- Root `demo-session` span with `model` attribute (wired this session after dead-state review fix)
- 20/20 unit tests pass at HEAD `8899f9ea`

python/tests/test_live_tracing.py
```diff
@@ -0,0 +1,517 @@
+"""Unit tests for ``scenario._tracing.live.RealtimeLangWatchSession``.
```
drewdrewthis, line 1: The test deliberately tests the weaker-but-achievable property: after OTel state is reset to a fresh proxy, no code in the already-loaded scenario module re-installs a concrete provider. This guards against session-reuse bugs where scenario resets the global provider on re-entry.

The stronger guarantee (import itself never installs a provider at module-load time) would require subprocess isolation, which is out of scope for this AC. The docstring and inline comment already document this limitation. Filed a follow-up note in the issue rather than blocking this PR on subprocess test infrastructure.

python/scenario/_tracing/live.py
```diff
@@ -0,0 +1,263 @@
...
+            if self._model is not None:
+                self._root_span.set_attribute("model", self._model)
+            self._ctx_token = context.attach(
+                trace.set_span_in_context(self._root_span)
+            )
+        except Exception:  # noqa: BLE001 - never raise into the host app
+            logger.warning(
+                "Failed to start realtime_langwatch_session root span; "
+                "continuing as a no-op.",
+                exc_info=True,
```
drewdrewthis: **[security][metz-beck] Fix: `exc_info=True` here can expose the Bearer token**

The outer `except` wraps `_add_langwatch_exporter(provider, self._api_key)`, which constructs `OTLPSpanExporter(headers={"Authorization": f"Bearer {api_key}"})`. If the exporter constructor or any other call in this block raises an exception whose `repr()` includes the request context (HTTP client libraries sometimes do), `exc_info=True` writes the Bearer token to WARNING logs — exactly the first-deployment failure mode users are most likely to hit.

Drop `exc_info=True` from this one log call and include the error type only:

```python
except Exception as exc:  # noqa: BLE001
    logger.warning(
        "Failed to start realtime_langwatch_session root span; "
        "continuing as a no-op. (error type: %s)",
        type(exc).__name__,
    )
```

The three `__aexit__` handlers with `exc_info=True` are safe — no credential is in scope there.

python/tests/test_live_tracing.py
```diff
@@ -0,0 +1,516 @@
...
+    spans = exporter.get_finished_spans()
+    turn_spans = [s for s in spans if s.name == "realtime_turn"]
+    assert len(turn_spans) == 1
+    span = turn_spans[0]
+    assert span.attributes is not None
+    assert span.attributes["type"] == "llm"
+    assert span.attributes["input"] == "Hello"
+    assert span.attributes["output"] == "Hi there"
+    assert span.attributes["model"] == "gpt-4o-realtime-preview"
+    assert span.attributes["latency_ms"] == 450
```
drewdrewthis, line 123: **[test] Fix: AC2 claims a *child* span but the test never asserts the parent–child link**

AC2 spec: "emits a child LLM span under the root span." The test confirms the span exists and has correct attributes, but never verifies `span.parent` points to the root span. The context-attachment path in `log_turn` (`trace.set_span_in_context(self._root_span)`) is the load-bearing part — if that silently breaks, this test still passes.

Add after line 123:

```python
# AC2 also requires the turn span to be a *child* of the root span
root_spans = [s for s in spans if s.name != "realtime_turn"]
assert len(root_spans) >= 1, "expected a root span"
assert span.parent is not None, "turn span should have a parent"
assert span.parent.span_id == root_spans[0].context.span_id
```

rogeriochaves (approved):

Issue #686, opened 2026-06-18
Add a first-class Claude Code AgentAdapter integration


drewdrewthis:
## For Humans

**People increasingly want to scenario-test Claude Code itself (a skill, a subagent, the CLI), but there is no first-class adapter — each team hand-rolls a child-process adapter and parser.** A working reference implementation exists privately. The codebase audit confirms `javascript/src/agents/claude-code/` is the idiomatic in-tree home, but a challenge pass argues for a standalone `@langwatch/scenario-claude-code` package instead: the stream-json output shape has no stability contract, there is no npm optional-dep story for a system binary, and shipping in-tree sets a precedent that Codex/Aider/Cursor then inherit.

**Ask:** Decide where a maintained `ClaudeCodeAgentAdapter` should live, then ship it.

- Idiomatic in-tree location is `javascript/src/agents/claude-code/`, re-exported via `agents/index.ts` → `index.ts`.
- Challenge pass favors standalone `@langwatch/scenario-claude-code`: isolates maintenance, sidesteps precedent, no optional-dep problem.
- Python parity expected — voice adapters mirror JS one-for-one; a JS-only adapter is a half-measure.

_Status: enhancement, P3. Full spec below._

<details><summary>For Agents — full spec</summary>

## Investigation

**Symptoms / context:** No first-class way to point a scenario at the Claude Code CLI. A working reference `AgentAdapter` exists (spawns `claude -p --output-format stream-json --verbose` in a workdir, parses stream-json stdout into scenario messages, plus helpers that copy a `SKILL.md` into `.skills/<name>/` and write a `CLAUDE.md`, and an `assertSkillWasRead(state, name)` helper). It is not lift-and-ship — there are known hardening gaps (see below).

**Key findings (verified against the main checkout):**

- Adapter packaging convention: text/agent adapters live under `javascript/src/agents/<name>/` (e.g. `agents/realtime/realtime-agent.adapter.ts`), voice adapters under `javascript/src/voice/adapters/<name>.ts` with a factory in `voice/factories.ts`. Tests co-located under `__tests__/`. Voice adapters get a `docs/pages/voice/adapters/<name>.mdx`; text adapters do not get per-adapter doc pages.
- `javascript/src/integrations/` holds ONLY vitest test-runner plumbing (`integrations/vitest/`), NOT agent adapters — the idiomatic in-tree location is `javascript/src/agents/claude-code/`, re-exported via `agents/index.ts` → `index.ts`.
[...]

PR #687, opened 2026-06-18
feat: add Claude Code agent adapter

### LINKED ISSUE #690: ClaudeCodeAgentAdapter multiturn is stateless transcript-replay, not session continuation (use --resume/--session-id)
## Expected

The `ClaudeCodeAgentAdapter` (added in #687) should drive a **real continuous multiturn session**: turn 2 continues the session started on turn 1, so the agent-under-test experiences preserved session state, context caching, and resumed tool state — the way a production user (or the interactive TUI) does. The Claude Code CLI supports this in print mode via `-c/--continue`, `-r/--resume [sessionId]`, `--session-id <uuid>`, and `--fork-session`.

## Actual

Every `call()` spawns a **brand-new** `claude -p` process and re-serializes the **entire** conversation history into one flat `role: content\n\nrole: content…` prompt string (`formatMessagesAsPrompt` → positional prompt arg in `buildArgs`). No session is threaded between turns — none of `--continue`/`--resume`/`--session-id` is used. Multiturn is therefore **stateless transcript-replay**, not session continuation: each turn is a cold start that re-reads the whole history as a single user message.

Consequence: scenarios that depend on real session continuity (does the agent remember/behave correctly across an *ongoing* session?) are testing prompt-replay, not session-resume. This is a fidelity gap between what the adapter claims ("exercises the real shipped binary") and what it actually exercises.

## Reproduction

Run any multiturn scenario through the adapter and observe that each turn launches a fresh process with the full transcript re-fed as the prompt (no session id captured or reused). The CLI's `--resume`/`--session-id` flags are never passed.

## Suggested direction (for investigation, not a mandate)

Capture the session id from the first turn's stream-json `init`/`system` event, then pass `--resume <id>` (or `--session-id <uuid>`) on each subsequent `call()`. Investigation should confirm how the session id surfaces in stream-json and whether `--resume` in `-p` mode behaves as needed. Related to the broader headless-vs-TUI fidelity question in #689.

Refs #687, #689

---
*Classification: Bug*
*Status: stub — investigation pending. Issue is not-ready until the investigation step completes.*

drewdrewthis:
> [!IMPORTANT]
> **Scariest thing** — this adapter spawns the real `claude` CLI and threads a **server-side session** across turns. When that session vanishes, the tempting fix is to silently replay the transcript into a fresh session. That is a *lossy modality change* under an eval harness: the agent-under-test no longer experiences a real session, and the judge scores the degraded turn as if nothing happened. It is now **opt-in** (`replayOnLostSession`, default `false`).
> **Start here** — `javascript/src/agents/claude-code/claude-code-agent.adapter.ts`, `call()`. It owns *all* session-map policy; `runTurn()` is a pure one-spawn helper that mutates nothing.

## Why

Teams want to **scenario-test Claude Code itself** — a skill, a subagent, the CLI — but there is no first-class adapter, so each team hand-rolls a `child_process` spawn + stream-json parser. This adds a maintained `ClaudeCodeAgentAdapter` that drives the real CLI as an agent-under-test.

Resolves the placement question raised in #686 (decision: ship in-tree, beside the existing realtime/voice adapters). Closes #690.

## The design flaw, and the one hiding behind it

#690 flagged that multiturn was **stateless transcript-replay**: every `call()` cold-started `claude -p` and re-fed the whole transcript. That was fixed earlier on this branch (`8b52c54`) by capturing turn 1's `session_id` and passing `--resume <id>` afterwards. Verified live, not assumed: `--session-id X` "remember MANGO" → later `--resume X` "what word?" → **`MANGO`**.

The **"self-healing sessions"** shipped alongside it was never real. On a vanished session the code evicted the dead id "so the next turn rebuilds from full history" — and then **rejected**. `ScenarioExecution.callAgent` rethrows anything an adapter's `call()` throws, which **aborts the entire run**. *There is no next turn.* The eviction was unreachable in the only path that mattered. The two tests that "proved" it drove `adapter.call()` a **third time by hand** — something the framework never does after a rejected turn — and fabricated a `"(failed)"` assistant message to make turn 3 plausible. They tested around the bug.

Making the recovery reachable then exposed a second problem, caught in review: rebuilding by flat-replaying the transcript **is the very thing #690 condemned**, and the only signal was `logger.warn` — which defaults to a **no-op logger**. A silent modality switch under an eval harness corrupts the measurement while reporting success.

**So the fix is two-part:** recovery happens *inside* the turn (it cannot be deferred), and it is **opt-in**. By default a lost session fails loudly with an actionable error. Same house rule as `--dangerously-skip-permissions` in this adapter: risky-but-convenient does not get to be the default.

## What changed

- **A lost session throws a typed `LostSessionError`** (a `ClaudeCodeCliError` subclass exposing the dead `sessionId`/`threadId`, with the originating CLI error as `.cause`) → alternative: a plain `Error`. Review caught that the plain `Error` broke the class's own docstring promise — `catch (e) { if (e instanceof ClaudeCodeCliError) }` missed the one failure a caller most needs to branch on. Unreleased API, so free to fix now.
- **`replayOnLostSession?: boolean`, default `false`** → alternative: recover silently, keeping the run alive. Rejected because replay loses server-side session state (context cache, tool state) and the judge would score a turn that is no longer a `--resume` continuation. Blast radius: a vanished session now **aborts the run by default** — loudly, naming the flag and what it costs.
- **`call()` owns all session policy; `runTurn()` is pure** → alternative: leave the `sessions.set` inside `runTurn`. `call()` is the only frame that knows a turn was a resume, so eviction belongs there. It now evicts **only** on a genuinely stale session — a rate limit, a signal, a timeout, or the client-side empty-delta guard (which never even spawns) no longer throw away a healthy session.
- **Stale detection is structural first, stderr second** → alternative: keep regexing stderr. The CLI *fields* the failure: `{"type":"result","subtype":"error_during_execution","is_error":true,"errors":["No conversation found with session ID: …"]}`. This also closes a latent hole: success was inferred from `exitCode === 0` alone, so an `is_error: true` run with a 0 exit would have resolved as a **silent empty-text success** and stored a session id minted by a failed turn.
- **`rate_limit_event` + `stream_event` known; unknown-event warning deduped per type** → the real CLI emits `rate_limit_event` on healthy runs, and `stream_event` at *token* granularity under `--include-partial-messages` (reachable today via `extraArgs`): one single-word reply emitted **26** of them. A warning that fires on a healthy run destroys the wire-drift signal it exists to give. Deduping is the durable fix; allowlisting alone is whack-a-mole.

## How it works

```
src/agents/claude-code/
  claude-code-agent.adapter.ts   ClaudeCodeAgentAdapter + ClaudeCodeCliError + config + Logger
                                 call()    → session policy: resume / evict-if-stale / reject-or-replay
                                 runTurn() → exactly one `claude -p` spawn; pure, returns {text, sessionId}
  stream-json.ts                 parseStreamJson() → {text, messages, sessionId, result}; per-type warn dedupe
  skill-injection.ts             assertSkillWasRead(), injectSkill()
  index.ts                       barrel + claudeCodeAgent() factory
docs/docs/pages/agent-integration/claude-code.mdx   adapter doc page
```

Per `threadId`: **turn 1** sends the full history, no `--resume`, and stores the `session_id` the CLI stamps on its init line. **Later turns** pass `--resume <id>` and send only `input.newMessages`. **If that resume finds no session**, the dead id is evicted and — depending on `replayOnLostSession` — the turn either rejects with an actionable error (default) or re-runs in place from full history against a fresh session.

## Test plan

```
cd javascript
pnpm typecheck                                                        # clean
pnpm eslint src/agents/claude-code/ src/agents/__tests__/claude-code-adapter.test.ts src/agents/index.ts   # 0 problems
pnpm vitest run src/agents/__tests__/claude-code-adapter.test.ts      # 51 passed | 3 skipped (E2E gated off)
RUN_CLAUDE_CODE_E2E=1 pnpm vitest run src/agents/__tests__/claude-code-adapter.test.ts   # 54 passed (real CLI)
pnpm run test:ci                                                      # 950 passed | 4 skipped (whole package)
```

Unit tests mock `child_process.spawn` (no CLI, no network). The regression that actually gates in CI is **`a mid-run stale session does not abort the scenario`** — it drives a real `ScenarioExecution` over a mocked spawn, which is the level the old tests never reached and the only level at which this bug is visible.

> **The `RUN_CLAUDE_CODE_E2E` tests are dev-only smoke checks.** `RUN_CLAUDE_CODE_E2E` appears in no workflow, so they never run in CI. Do not mistake their presence for enforced coverage — the mocked `ScenarioExecution` test is the enforced guard. This is stated in the test file's own doc comment.

## Human verification

From a fresh checkout, a skeptical reviewer can re-verify **on this single commit** — no relative refs, nothing to restore:

1. **See the real CLI's stale-resume failure.** `claude -p --output-format stream-json --verbose --resume 00000000-0000-4000-8000-000000000000 "hi"` → exit `1`, and a stdout `result` line with `"subtype":"error_during_execution"`, `"is_error":true`, `"errors":["No conversation found with session ID: …"]`.
2. **Default = fail loudly.** In `src/agents/__tests__/claude-code-adapter.test.ts`, delete `replayOnLostSession: true` from the `recovers a real vanished session` test, then `RUN_CLAUDE_CODE_E2E=1 pnpm vitest run … -t "recovers a real vanished session"` → **FAILS**, and the error names `replayOnLostSession: true` and what replay costs.
3. **Opt-in = survives.** Restore the flag, re-run → **passes**; turn 2 still recalls `ZEBRA42` after its session was deleted mid-run.
4. **The CI-enforced test is falsifiable.** Remove `replayOnLostSession: true` from the mocked `a mid-run stale session does not abort the scenario` test → it **FAILS** (no CLI needed).
5. **Mocked suite, no CLI/network.** `pnpm vitest run src/agents/__tests__/claude-code-adapter.test.ts` → `51 passed | 3 skipped`.
6. **Public API.** `pnpm build`, then grep `dist/index.d.ts` for `ClaudeCodeAgentAdapter`, `claudeCodeAgent`, `ClaudeCodeAgentAdapterConfig`, `ClaudeStreamMessage`, `ClaudeCodeCliError`, `LostSessionError`.

**[proven] The bug is real, the fix is load-bearing, and the test can fail.** The E2E runs a real `ScenarioExecution`, lets turn 1 mint a real session, then **deletes that session's on-disk transcript**, so turn 2's `--resume` hits the genuine `No conversation found` path. Everything below is the *same commit* — only the flag changes:

Real Claude Code CLI at 09b43d5: with replayOnLostSession unset the run aborts with an actionable error; with it set the run survives and recalls the turn-1 token; the CI-enforced mocked test fails when recovery is disabled; full suites green.

```
# A — real CLI, replayOnLostSession UNSET (the default)
FAIL  … > recovers a real vanished session mid-run and still recalls turn 1
Error: [ClaudeCodeAgent] Error: Claude Code session 1c88d6a2-… no longer exists
  (--resume reported: No conversation found). Set replayOnLostSession: true to rebuild
  this turn from the full transcript instead — note that … loses server-side session
  state (context cache, tool state), so the turn is no longer a true session continuation.
     Tests  1 failed | 53 skipped (54)

# B — real CLI, replayOnLostSession: true
     Tests  1 passed | 53 skipped (54)        ← run survives, turn 2 recalls ZEBRA42

# C — CI-ENFORCED mocked test, recovery disabled (no CLI)
FAIL  … > a mid-run stale session does not abort the scenario
     Tests  1 failed | 53 skipped (54)        ← the guard is falsifiable

# D — at HEAD
RUN_CLAUDE_CODE_E2E=1 …   Tests  54 passed (54)          (real CLI + real judge)
pnpm vitest run           Tests  950 passed | 4 skipped  (what CI runs)
```

**[proven] Recovery preserves context, not just liveness.** After the rebuild, turn 2 still answers with `ZEBRA42` — the token set in turn 1. The E2E also asserts the CLI was invoked exactly **3** times (turn 1, the doomed resume, the rebuild) and that the recovery warning fired.

**[proven] `--resume` really continues a session** (#690's premise, re-verified rather than trusted): `--session-id X` "remember MANGO" → `--resume X` "what word?" → `MANGO`, same `session_id` both runs.

**[proven] `--session-id` cannot replace `--resume`.** It is **create-only**: first use exits 0, a second use of the same id errors `Session ID <id> is already in use.` (exit 1). The "do not pin a self-generated `--session-id`" rationale in the code is therefore correct, and the comment now says *why* precisely (the hazard is resuming a partially-created session).

**[proven] Nothing else regressed.** `pnpm run test:ci`: `79 passed | 1 skipped` files, `945 passed | 4 skipped` tests. `pnpm typecheck` clean; scoped `eslint` `0 problems`.

**[proven] The previous proof screenshot was wrong and is replaced.** It was captioned with relative refs (`adapter @ HEAD~1`); two commits later that ref pointed at the fix itself, so a reviewer following it would restore an already-fixed adapter, see the test pass, and conclude the bug was invented. The current capture is pinned to `09b43d5` and reproduces on one commit.

## Anything surprising?

- **The default changed.** A vanished session now aborts the run unless `replayOnLostSession: true`. Deliberate: silent lossy recovery under an eval harness is worse than a loud failure. The recovery machinery is what this PR fixes — it was previously unreachable dead code — but it does not get to be the default.
- **Stale detection still text-matches a CLI sentence** (`No conversation found`) inside the structural envelope. Not a stability-contracted interface. Tracked in #752.
- **The CLI has a native persistent transport** — one long-lived `claude -p --input-format stream-json` process holds a session across turns with no `--resume`, no session map, and no way for the session to vanish. It would delete this entire apparatus. Not evaluated when the adapter was designed. Tracked in #753.
- **`recordException` exports raw adapter stderr on every agent-span failure**, repo-wide and pre-existing (not introduced here). `ClaudeCodeCliError.stderr` is enumerable, so `JSON.stringify(err)` serializes it. The docs page now warns against logging it verbatim. Tracked in #754.
- **`--dangerously-skip-permissions`** remains opt-in and off by default; a signal-terminated child rejects with the signal name and is never treated as a recoverable stale session.
- **JS only.** `python/scenario` parity is tracked in #688.
- **Doc placement** stays at `docs/pages/agent-integration/claude-code.mdx`; the only per-adapter doc precedent is `voice/adapters/`, and this is not a voice adapter. Relocate if you prefer.


javascript/src/agents/__tests__/claude-code-adapter.test.ts
```diff
@@ -0,0 +1,486 @@
...
+
+      // Console was never used by the adapter.
+      expect(consoleLog).not.toHaveBeenCalled();
+      expect(consoleWarn).not.toHaveBeenCalled();
+      expect(consoleError).not.toHaveBeenCalled();
+    } finally {
+      consoleLog.mockRestore();
+      consoleWarn.mockRestore();
+      consoleError.mockRestore();
+    }
```
drewdrewthis, line 1778: **E2E gate asserts shape, not a verdict.**

`expect(typeof result.success).toBe('boolean')` is satisfied by `false` — a failing scenario with a broken agent still passes this test. The integration test's only real signal is that it didn't throw. Change to `expect(result.success).toBe(true)` so a judge failure actually fails the test.

javascript/src/agents/__tests__/claude-code-adapter.test.ts
```diff
@@ -0,0 +1,486 @@
...
+  });
+});
+
+// --- 7. skill helpers -------------------------------------------------------
```
drewdrewthis, line 1782: **Tautological assistant-turn assertion.**

`result.messages.some((m) => m.role === 'assistant')` will be true as long as Claude Code emits any output — it doesn't verify the *agent under test* answered meaningfully. If you're going to assert anything beyond `result.success === true`, assert that the answer contains `'4'` (the only correct answer to '2+2'). Without that, a hallucinating or empty response passes.

javascript/src/agents/__tests__/claude-code-adapter.test.ts
```diff
@@ -0,0 +1,486 @@
...
+            {
+              type: "tool_result",
+              content: [
+                { type: "text", text: "FILE_CONTENTS_MARKER" },
+                { type: "text", text: "line two" },
+              ],
+            },
+          ],
+        },
+      }),
```
drewdrewthis, line 1428: **Non-zero exit code path is untested.**

The adapter rejects with `'Claude Code CLI failed with exit code N'` when `exitCode !== 0 && exitCode !== null`. There is no test for this. `child.close(1)` is a one-liner. This is a hardening behavior listed in the header comment but missing from the numbered coverage list and from the suite.

javascript/src/agents/__tests__/claude-code-adapter.test.ts
```diff
@@ -0,0 +1,486 @@
...
+    const err = new Error(code) as NodeJS.ErrnoException;
+    err.code = code;
+    this.emit("error", err);
+  }
+}
+
+/** Install the spawn mock to return `child` and record the call. */
+function withChild(child: FakeChild): void {
+  spawnMock.mockReturnValue(child as unknown as ReturnType<typeof spawn>);
+}
```
drewdrewthis, line 175: **`skillPath`/`injectSkill` factory wiring is untested.**

The `claudeCodeAgent` factory (in `index.ts`) calls `injectSkill(config.workingDirectory, config.skillPath)` at construction time — this is the only user-facing factory and its skill-injection side-effect is entirely uncovered. The unit tests for `assertSkillWasRead` exercise the assertion helper but never the injection path. A test that calls `claudeCodeAgent({ skillPath: '...', workingDirectory: tmpdir })` and checks that `.skills/<name>/SKILL.md` and `CLAUDE.md` exist would close this gap.

javascript/src/agents/__tests__/claude-code-adapter.test.ts
```diff
@@ -0,0 +1,486 @@
...
+    withChild(child);
+    const logger = spyLogger();
+
+    const lines = [
+      JSON.stringify({ type: "some_new_event", payload: { x: 1 } }),
+      JSON.stringify({
+        type: "assistant",
+        message: { role: "assistant", content: [{ type: "text", text: "KNOWN_TEXT" }] },
+      }),
+    ].join("\n");
```
drewdrewthis, line 1500: **Logger isolation test over-specifies stderr routing.**

`expect(logger.warn).toHaveBeenCalled()` + `expect(warned).toContain('some stderr noise')` is correct. But the immediately preceding `expect(logged).toContain('/tmp/diag')` relies on an undocumented implementation detail — that the startup log line includes the working directory string. If that message ever changes format, this test breaks on wording rather than behavior. The intent is only "startup was logged"; `expect(logger.log).toHaveBeenCalled()` without the string match is sufficient and more stable.

drewdrewthis (commented):
**Test review — overall verdict: CONDITIONALLY TRUSTWORTHY**

The mock architecture is sound (FakeChild via vi.hoisted mirrors the playback.test.ts pattern correctly). Fake timers are used for the timeout test — deterministic. The six hardening behaviors are all exercised. Three issues worth fixing before this is full confidence: (1) the E2E gate asserts structural shape, not a real verdict; (2) the `skillPath`/`injectSkill` factory path has zero coverage; (3) non-zero exit code behavior is absent. Details in the inline threads below.

javascript/src/agents/claude-code/claude-code-agent.adapter.ts
```diff
@@ -0,0 +1,273 @@
...
+  }
+
+  /**
+   * Process the conversation and return Claude Code's assistant text.
+   *
+   * @param input - Scenario agent input (conversation history etc.).
+   * @returns The concatenated assistant-visible text as a string.
+   */
+  async call(input: AgentInput): Promise<AgentReturnTypes> {
+    const prompt = formatMessagesAsPrompt(input.messages);
```
drewdrewthis: **MEDIUM — prompt loses all non-text turn content, leaving dangling `role:` labels.** `formatMessagesAsPrompt` → `extractText` keeps only `text` parts. For a `ModelMessage` whose content is an array of tool-call or tool-result parts (assistant tool turns, `role:"tool"` results), `extractText` returns `""`, so the prompt emits a bare `assistant: ` / `tool: ` line with no content. I verified this directly:
```
TOOLCALL   -> "assistant: "
TOOLRESULT -> "tool: "
```
In any multi-turn scenario where the conversation already contains tool activity (the common case for an agent under test), Claude Code receives orphaned role labels and silently loses the tool context. At minimum, drop turns that reduce to empty content rather than emitting a label-only line; better, render tool-call/tool-result parts the way `stream-json.ts` already does (you have `renderToolResultContent` right there). Note the irony: the *output* parser carefully avoids `[object Object]`, but the *input* formatter throws the same structured content away.

javascript/src/agents/claude-code/claude-code-agent.adapter.ts
```diff
@@ -0,0 +1,273 @@
...
+    const bin = this.resolveBin();
```
drewdrewthis, line 481: **MEDIUM — agent-first / empty-messages case sends an empty prompt to the CLI.** When `input.messages` is empty (agent speaks first — a supported scenario shape), `formatMessagesAsPrompt([])` returns `""` (verified), and `buildArgs` passes `""` as the final positional prompt arg: `claude -p ... ""`. The realtime sibling explicitly handles this via `handleInitialResponse()` (realtime-agent.adapter.ts:150). Here the behavior is undefined — `claude -p ""` with no prompt is at best a no-op turn, at worst an error surfaced as a confusing nonzero exit. Either short-circuit empty messages with a sensible default opener, or document that this adapter requires a user-first script.

javascript/src/agents/claude-code/claude-code-agent.adapter.ts
```diff
@@ -0,0 +1,273 @@
...
+          );
+          return;
+        }
+        finish(() => reject(err));
+      });
+
+      child.on("close", (exitCode) => {
+        finish(() => {
+          if (exitCode !== 0 && exitCode !== null) {
+            reject(new Error(`Claude Code CLI failed with exit code ${exitCode}`));
```
drewdrewthis: **MEDIUM — nonzero-exit rejection swallows stderr, the actual failure reason.** On a nonzero exit the error is just `Claude Code CLI failed with exit code ${exitCode}`. But CLI failures (bad/expired auth, rate limit, unknown `--model`, bad cwd) write *why* to stderr — which here is only `logger.warn`'d (and lost entirely when no logger is injected, the default). For a testing-framework adapter, a failure that reports only "exit code 1" with no captured stderr is a debugging dead-end. Buffer stderr into a local (you're already receiving the chunks at line 206) and include the tail in the rejection message, e.g. `... exit code ${exitCode}: ${stderrTail}`.

javascript/src/agents/claude-code/claude-code-agent.adapter.ts
```diff
@@ -0,0 +1,273 @@
...
+
+      const finish = (fn: () => void): void => {
+        if (settled) return;
+        settled = true;
+        clearTimeout(timer);
+        fn();
+      };
+
+      child.stdout?.on("data", (data: Buffer) => {
+        stdout += data.toString();
```
drewdrewthis: **LOW — per-chunk `data.toString()` can corrupt a multibyte UTF-8 char split across chunk boundaries.** Each `data` Buffer is decoded independently and concatenated as strings, so a UTF-8 sequence straddling two chunks decodes to replacement chars (`\uFFFD`) on each side. Mostly-ASCII JSON makes this rare, but tool output / file contents can carry multibyte text. Decode once over the full byte stream: collect chunks in a `Buffer[]` and `Buffer.concat(chunks).toString("utf8")` at close, or pipe through `new StringDecoder("utf8")`. (Note: this is unrelated to JSON-line splitting, which your accumulate-then-parse-once design already handles correctly.)

javascript/src/agents/claude-code/claude-code-agent.adapter.ts
```diff
@@ -0,0 +1,273 @@
...
+        stdio: ["ignore", "pipe", "pipe"],
+      });
+
+      let stdout = "";
+      let settled = false;
+
+      const timer = setTimeout(() => {
+        if (settled) return;
+        settled = true;
+        child.kill();
```
drewdrewthis, line 507: **LOW — timeout `child.kill()` is SIGTERM-only, no SIGKILL escalation.** A CLI that traps/ignores SIGTERM keeps running after the timeout fires. The promise still rejects (settled flips, so no hang — good), but you can leak a lingering subprocess. Consider `child.kill("SIGTERM")` followed by a short `setTimeout` → `child.kill("SIGKILL")` to guarantee teardown. Minor for normal CLI use; matters under a wedged child.

javascript/src/agents/claude-code/skill-injection.ts
```diff
@@ -0,0 +1,104 @@
...
+ * Code discovers it, and write a `CLAUDE.md` pointing at all discovered skills
+ * (unless one already exists).
+ *
+ * The skill name is derived from the SKILL.md's parent directory name, matching
+ * the reference helper (`path.basename(path.dirname(skillPath))`).
+ *
+ * @param workingDirectory - The directory Claude Code is spawned in.
+ * @param skillPath - Absolute path to a `SKILL.md` to inject.
+ */
+export function injectSkill(workingDirectory: string, skillPath: string): void {
```
drewdrewthis, line 41: **LOW (test infra, calibrated) — `injectSkill` does filesystem writes as a construction-time side effect and silently clobbers an existing skill copy.** `claudeCodeAgent({ skillPath })` calls `injectSkill` synchronously in the factory (index.ts:24), so merely *constructing* the agent writes `<wd>/.skills/<name>/SKILL.md` via `copyFileSync` (unconditional overwrite) and may write `CLAUDE.md`. Two notes: (1) construction-with-side-effects is surprising — a reader expects the write to happen at `call`/run time; (2) `skillName` derives from `path.basename(path.dirname(skillPath))`, so two skills whose parent dirs share a basename collide silently, and a re-run overwrites without warning. Fine for trusted test fixtures (no untrusted-path traversal concern here), but worth a doc note that injection mutates `workingDirectory` eagerly and overwrites. No sandbox needed.

drewdrewthis (commented):
Adversarial review of the Claude Code adapter (substantive files only; skipped the .mdx and PNG).

Verdict: **no blockers** — the happy path is correct, the subprocess lifecycle is race-safe (the `settled`/`finish` gate makes close/error/timeout mutually exclusive, so the promise always settles exactly once — no hang), and the 13 unit tests pass + tsc is clean. The partial-line-buffering concern does NOT apply here: stdout is fully accumulated and `parseStreamJson` runs once inside `close`, after the whole buffer is assembled, so a JSON object split across `data` events is reassembled correctly. Nice.

What I found are quality/fidelity issues, all MEDIUM-or-below, none fatal. The two worth addressing before merge are the prompt-formatting content loss and the swallowed stderr on CLI failure (both hurt the framework's own debuggability). Inline below.

drewdrewthis:
Thanks for the thorough review. All findings addressed in `dcddaa7` and `584f00a`:

**MEDIUM**
- **Input formatter dropped structured content** → `extractText` replaced with a part-renderer: `tool-call` → `[tool-call: name(input)]`, `tool-result` → `[tool-result: name -> output]`, `file`/`reasoning` handled; tool-only turns no longer collapse to an empty `assistant:` label.
- **Agent-first / empty-messages sent `claude -p ""`** → `call()` now rejects with a descriptive error when there's no renderable content (a CLI prompt can't open agent-first; failing loudly surfaces the wiring bug instead of masking it).
- **Nonzero-exit swallowed stderr** → the rejection now appends captured stderr (the real cause: auth/rate-limit/bad-model).

**LOW**
- Per-chunk `data.toString()` multibyte corruption → stdout/stderr accumulate as `Buffer[]`, decoded once via `Buffer.concat(...).toString("utf8")` at close.
- Timeout `SIGTERM`-only → escalates to `SIGKILL` after 2s (cleared on exit).
- `injectSkill` eager-write/clobber → documented in a code comment (trusted-fixture assumption).

**Tests**
- E2E was verdict-blind (`typeof ... === "boolean"`) → now asserts `result.success === true` AND the assistant answered `"4"` (fails loudly on a false verdict).
- Tautological assistant-turn assertion → folded into the above.
- Added unit coverage for the nonzero-exit path and the factory `skillPath` injection side effect; added an empty-messages guard test.
- Logger-isolation test over-specified the cwd string → loosened to `toHaveBeenCalled()` + the console-never check.

Full suite green: unit **16 passed | 1 skipped**, real-CLI E2E **17 passed** (incl. the empty-messages guard), typecheck + scoped eslint clean.

drewdrewthis:
Python-parity follow-up tracked in #688 (the JS adapter here ships first; `python/scenario` mirrors it next, matching the existing voice-adapter parity).

drewdrewthis:
## Review verdict: READY

**Reviewed at:** `09b43d5` · **Run:** /review --since a1f89f2 (own-PR), two passes

No blocking concerns. **0 raw unresolved, non-outdated threads** (checked including CodeRabbit, not only review-opened ones). Both blocking findings from the second pass are fixed and their threads resolved.

### Review coverage, stated exactly

| Pass | Reviewers | Read |
|---|---|---|
| 1 | principles, hygiene, security, test, design-soundness, drift (6/6) | `a1f89f2..4f33293` |
| 2 | security, principles, test (3/3) | `4f33293..ade0694` — the commits that *implemented* pass 1, which no reviewer had read |
| author | verification only | `ade0694..09b43d5` (audit fixes + proof re-capture) |

Every finding was independently re-verified against the live Claude Code CLI (2.1.205) or the installed SDK before being acted on. Pass 2 exists because the code that fixed pass 1's findings would otherwise have merged unreviewed.

### What review changed

**Pass 1** — the delta arrived claiming an in-turn "self-healing" recovery. Review established the recovery was right to exist and wrong to be automatic:

- **[design-soundness]** The recovery re-flattened the whole transcript into a fresh session — *the exact stateless replay #690 condemned* — signalled only through `logger.warn`, which defaults to a **no-op logger**. Under an eval harness that is a silent modality change scored as a clean turn. → `replayOnLostSession`, **default `false`** (fail loudly). Recorded as a decision, not buried in a diff.
- **[design-soundness]** Latent hole: success was inferred from `exitCode === 0` alone, so an `is_error: true` run with a 0 exit resolved as a **silent empty-text success** and stored a session id minted by a failed turn. → the terminal `result` envelope is now parsed and trusted; stale detection is structural-first, stderr-fallback.
- **[principles]** `call()` evicted the session on *any* failed resume — including the empty-delta guard, which throws **before the CLI is ever contacted**. → evict only on a genuinely stale session; `runTurn()` is now pure and `call()` owns all session-map policy.
- **[hygiene]** `stream_event` (26 lines from one single-word reply, reachable via `extraArgs`) floods the "unknown event" warning — the very failure the `rate_limit_event` patch claimed to fix. → allowlisted **and** deduped per unique type.
- **[security]** `deleteLiveSession()` built a path from an unvalidated CLI-supplied `session_id` and `rmSync`'d it (`path.join(projects,'p','../../../../tmp/pwned.jsonl')` → `/home/tmp/pwned.jsonl`). → anchored UUID check, containment check, symlink guard.
- **[test]** The gated E2E leaked its temp dir on exactly the failure it exists to catch; a resume-turn **timeout** silently gained an eviction side effect; `does not recover a first turn` was not falsifiable coverage. → `finally` cleanup, timeout-no-evict test, honest naming.
- **[drift]** The published proof pointed at `HEAD~2` — which *is* the fix commit — so it could never reproduce. → re-captured, pinned to absolute SHAs, reproducible on one commit.

**Pass 2** — two blocking defects in the *fixing* commits:

1. **[principles]** The lost-session path threw a plain `Error`, while every other CLI failure surfaced as a `ClaudeCodeCliError` and `isStaleSessionError` was module-private. `catch (e) { if (e instanceof ClaudeCodeCliError) }` therefore **missed the one failure a caller most needs to branch on**, and the class docstring's promise was false on exactly that path. → exported `LostSessionError extends ClaudeCodeCliError`, carrying the dead `sessionId`/`threadId` and the originating error as `.cause`. *(thread resolved)*
2. **[security]** The sensitive-data caveat named only `stderr`, but the delta added two more enumerable, untrusted fields — `subtype` and `errors[]` — documented nowhere. → caveat extended in source and docs. *(thread resolved)*

**[design-soundness] disproved a hypothesis rather than confirming it:** `--session-id` is **create-only** (second use → `Session ID <id> is already in use.`, exit 1), so it cannot replace `--resume`. The existing "do not pin a self-generated `--session-id`" rationale is correct and must survive future edits.

### Non-blocking (Decide / New Issue)

- **[security]** *(Decide — resolved by documenting)* `errors[]` is the preferred source embedded into `.message`, and error messages are **exported to tracing automatically, no opt-out**: `langwatch@0.16.1`'s `withActiveSpan` rejection handler calls `setStatus({message: err.message})` + `recordException(err)`; `@opentelemetry/sdk-trace-base` `Span.js:218` copies `exception.message` into span attributes. Zero redaction anywhere. `.message` was NOT restructured (an earlier accepted finding required stderr in the rejection for debuggability); the docs page now warns, and the systemic fix is #754.
- **[principles]** *(Decide — resolved by guarding)* `call()`'s catch has an implicit fall-through onto the shared rebuild; the success path's `return` is the only thing stopping a successful resume from re-running it. Pinned by a regression test rather than restructuring proven code.
- **[test]** *(Decide — resolved)* `RUN_CLAUDE_CODE_E2E` appears in no workflow; the real-CLI tests are dev-only smoke checks. The CI-enforced guard is the mocked `ScenarioExecution` test, and its falsifiability is demonstrated in the PR body.
- **[drift]** *(Decide — resolved)* `rate_limit_event`/`stream_event` are a disclosed scope rider, found by the same live-CLI evidence that proved the main fix.
- **[design-soundness]** *(New Issue)* The CLI ships a native persistent transport that would delete this entire session-map apparatus. → #753
- **[principles]** *(New Issue)* Stale detection still text-matches `No conversation found` inside the structural envelope. → #752
- **[security]** *(New Issue)* `recordException` exports raw adapter stderr — and the CLI-minted `session_id` — on every agent-span failure. Pre-existing, repo-wide. → #754
- **[test]** *(New AC — implemented, not deferred)* No test covered `replayOnLostSession: true` when the in-turn rebuild itself fails. Added: rejects with the rebuild's own error, as a bare `ClaudeCodeCliError` (never a `LostSessionError`), and turn 3 starts fresh with no phantom `--resume`.

### Evidence

Falsifiability on a single commit — real CLI, real `ScenarioExecution`, real deleted session transcript: `replayOnLostSession` unset → run aborts with a typed `LostSessionError`; set → run survives and turn 2 still recalls `ZEBRA42`; the CI-enforced mocked guard **fails** when recovery is disabled. Gated suite `54 passed`; whole package `950 passed | 4 skipped`; typecheck + scoped eslint clean; `dist/index.d.ts` exports `LostSessionError`. Screenshot and transcripts in the PR body.

---
_Verdict is prose, not a GitHub approval. Two review passes; every Fix-level finding applied and re-verified by the author at `09b43d5` on a clean tree._

javascript/src/agents/claude-code/claude-code-agent.adapter.ts
```diff
@@ -0,0 +1,343 @@
...
+          "The Claude Code CLI is prompt-driven and cannot open a conversation " +
+          "on its own; ensure a user turn precedes the agent (e.g. a " +
+          "scenario.user(...) step before scenario.agent()).",
+      );
+    }
+
+    const bin = this.resolveBin();
+    const args = this.buildArgs(prompt);
+    const cwd = this.config.workingDirectory;
+    const timeoutMs = this.config.timeout ?? DEFAULT_TIMEOUT_MS;
```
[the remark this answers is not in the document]
drewdrewthis: Fixed in 729db13 — added a guard that throws a clear config error when timeoutMs is non-finite or <= 0, before any timer is scheduled. Tests added for 0, -5, and NaN.

javascript/src/agents/claude-code/claude-code-agent.adapter.ts
```diff
@@ -0,0 +1,343 @@
...
+        );
+      }, timeoutMs);
+
+      const finish = (fn: () => void): void => {
+        if (settled) return;
+        settled = true;
+        clearTimeout(timer);
+        if (sigkillTimer) clearTimeout(sigkillTimer);
+        fn();
+      };
```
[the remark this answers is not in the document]
drewdrewthis: Fixed in 729db13 — the close handler now captures the signal param and rejects with 'terminated by signal X' (with stderr tail) instead of treating exitCode null as success. The sigkill timer is already cleared by finish(), so no extra clearing needed. Test added for (null, SIGKILL).

javascript/src/agents/claude-code/skill-injection.ts
```diff
@@ -0,0 +1,109 @@
...
+  const allContent = state.messages
+    .map((m) =>
+      typeof m.content === "string" ? m.content : JSON.stringify(m.content),
+    )
+    .join("\n");
+
+  const hasSkillRead =
+    allContent.includes("SKILL.md") ||
+    allContent.includes(`.skills/${skillName}`) ||
+    allContent.includes(`skills/${skillName}`);
```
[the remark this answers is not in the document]
drewdrewthis: Fixed in 729db13 — replaced the bare SKILL.md check with skill-name-scoped checks (.skills/<name>/SKILL.md etc). A different skill's SKILL.md no longer satisfies the assertion. Negative test added (other-skill does not satisfy my-skill).

javascript/src/agents/claude-code/stream-json.ts
```diff
@@ -0,0 +1,192 @@
...
+  return safeStringify(content);
+}
+
+/** `JSON.stringify` that never throws (circular refs → fallback string). */
+export function safeStringify(value: unknown): string {
+  try {
+    return JSON.stringify(value) ?? "";
+  } catch {
+    return String(value);
+  }
```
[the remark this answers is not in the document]
drewdrewthis: Fixed in 729db13 — wrapped the String(value) fallback in an inner try/catch returning '[unserializable value]', honoring the never-throws contract. Test added with an object whose toJSON and toString both throw.

drewdrewthis:
## ⚠️ Design flaw: multiturn is stateless transcript-replay, not real session continuation — marking not-ready

Flagging a design problem in this adapter that needs revision before merge.

**What it does today:** every `call()` spawns a brand-new `claude -p` process and re-serializes the *entire* conversation history into one flat `role: content\n\nrole: content…` prompt string (`formatMessagesAsPrompt` → positional prompt arg in `buildArgs`). No session is threaded between turns.

**Why that's wrong:** the Claude Code CLI ships first-class session-continuation for exactly this, usable in print mode:
- `-c, --continue` — continue the most recent conversation
- `-r, --resume [sessionId]` — resume a specific session
- `--session-id <uuid>` — pin a session id for the run
- `--fork-session` — fork on resume

So the adapter is fighting the tool: it could capture the session id from turn 1's stream-json `init`/`system` event and `--resume <id>` (or `--session-id`) on each subsequent `call()`, giving a **real continued session**. Instead it cold-starts every turn and replays the whole transcript as a single user message.

**Consequence (fidelity gap):** an agent-under-test never experiences a real multiturn session the way a production user (or the interactive TUI) does — no preserved session state, no context caching, no resumed tool state, no session-id continuity. Any scenario that depends on "does the agent remember/behave correctly across a *real* ongoing session?" is testing prompt-replay, not session-resume. This is the same fidelity concern as #689, along a second axis.

**Decision:** fix is warranted but deferred (not this iteration). The adapter should be reworked to capture the session id on the first turn and `--resume`/`--session-id` it on subsequent turns. Tracking as a follow-up; this PR is **marked not-ready** until that design is revisited.

cc #689 (headless-vs-TUI fidelity) — related.

javascript/src/agents/__tests__/claude-code-adapter.test.ts
```diff
@@ -0,0 +1,1172 @@
...
+            scenario.judge(),
+          ],
+        });
+
+        // Judge must pass — agent recalled the token across turns.
+        expect(result.success).toBe(true);
+
+        // The real assistant output must contain the token (proves --resume
+        // carried context, not a coincidental hallucination).
+        const assistantText = result.messages
```
drewdrewthis: **[test][principles] This assertion doesn't prove `--resume` carried context — it can pass on turn-1's echo.**

`assistantText` concatenates **all** assistant messages across **both** turns (`.filter((m) => m.role === "assistant")` — no turn boundary), then checks `.toContain("KUMQUAT77")`. Turn 1's user prompt is *"Remember this exact token… **Just acknowledge it.**"* — a compliant acknowledgment ("Got it, KUMQUAT77 noted") echoes the token, so the assertion is satisfied by **turn 1 alone**, even if `--resume` carried zero context into turn 2.

The comment above it — *"proves --resume carried context, not a coincidental hallucination"* — therefore overstates the guarantee: the one test whose entire purpose is cross-turn continuation has a deterministic assertion that cannot fail on a broken `--resume`. `result.success` (the judge) is the real guard, and the mocked unit tests already pin the wiring — so the test isn't worthless, but this deterministic check should actually isolate turn 2.

**Fix** — assert against the **last** (turn-2) assistant message instead of the join:

```ts
const lastAssistant = result.messages
  .filter((m) => m.role === "assistant")
  .at(-1);
const turn2Text =
  typeof lastAssistant?.content === "string"
    ? lastAssistant.content
    : JSON.stringify(lastAssistant?.content ?? "");
expect(turn2Text).toContain("KUMQUAT77");
```

Optionally harden further: phrase turn 1 so its acknowledgment can't echo the token (e.g. *"…reply with just the word ACK"*), and lift the literal to `const TOKEN = "KUMQUAT77"` reused in all three sites (turn-1 prompt, judge criterion, assertion) so they can't drift.

drewdrewthis: Fixed in 40153ce. You're right — concatenating both turns' assistant text let turn 1's echo of the token satisfy the check, so it never proved the --resume turn carried context. Now isolates the LAST assistant message (turn 2's reply on the resumed session) and asserts the token there. Verified the framework appends no judge message to result.messages and it is chronological, so .at(-1) of role assistant is turn 2's reply. Re-ran the fixed E2E against the real CLI + judge — passes (39s).

drewdrewthis:
Follow-up issues tracked (deferred test-quality improvements, non-blocking for this PR): #699, #700.

langwatch-agent:
🐕 **PR Hound review brief** — assigned reviewer @Aryansharma28

**Review mode:** Deep Review
**Scariest prod risk:** This adapter spawns the real Claude Code CLI as a subprocess and supports `--dangerously-skip-permissions`. If that flag is not strictly opt-in (default off), a scenario test could grant unrestricted permissions to a spawned agent. Also subprocess lifecycle bugs (hung child not killed on timeout) can leak processes in CI.

- `[Must Check]` `claude-code-agent.adapter.ts` — verify `--dangerously-skip-permissions` is gated behind `skipPermissions: true` and off by default, and that the timeout path actually kills the child (no orphaned `claude` processes) and rejects rather than parsing partial output as success.
- `[Must Check]` session resume via `--resume` — confirm turn-1 `session_id` capture and later-turn delta-only sending are correct, and the self-healing eviction on "No conversation found" rebuilds from full history without dropping messages.
- `[Probably Fine]` `assertSkillWasRead` scoping — verify a different skill's `SKILL.md` no longer satisfies the assertion (claimed hardening).

- On a signal-terminated child (`exitCode === null`), does the adapter reject with the signal name rather than treating possibly-incomplete stream-json as a successful turn?

drewdrewthis:
Status reconciliation on the "⚠️ Design flaw: multiturn is stateless transcript-replay" not-ready flag: this was addressed the same day in 8b52c54de ("feat(javascript): real multiturn session continuation in claude code") and subsequent commits. At current HEAD 40153ce8, `buildArgs` threads `--resume <session_id>` on continuation turns, `call()` keys a per-threadId session map (first turn: full messages; later turns: delta-only `newMessages`), the session id is captured from the stream-json init event, and a failed resume evicts the stale id so the next turn rebuilds from full history. Test suite: 37 passed / 2 e2e-gated skipped (`src/agents/__tests__/claude-code-adapter.test.ts`). The not-ready flag no longer reflects the code.

drewdrewthis:
## Resolving the ⚠️ not-ready design-flaw flag — and correcting the earlier "already fixed" reconciliation

The [not-ready flag](https://github.com/langwatch/scenario/issues/690) said multiturn was **stateless transcript-replay**. A [later comment](#issuecomment) claimed it "no longer reflects the code." That was **half right**, and the half that was wrong was load-bearing:

- ✅ **`--resume` really was implemented** (`8b52c54`). Re-verified live rather than taken on trust: `claude -p --resume <id>` continues the session, same `session_id` across turns, turn 2 recalls a turn-1 token.
- ❌ **The "self-healing sessions" it shipped alongside never worked.** On a vanished session the adapter evicted the stale id "so the next turn rebuilds from full history" — then **rejected**. `ScenarioExecution.callAgent` rethrows whatever an adapter's `call()` throws, which **aborts the run**. There is no next turn. The eviction was unreachable.

The two tests that covered it drove `adapter.call()` a **third time by hand** and fabricated a `"(failed)"` assistant message to make turn 3 plausible — the framework never does that after a rejected turn. They tested around the bug.

**Fixed in `3b27205`:** recovery now happens *inside* the turn. A resume that fails with `No conversation found with session ID` (Claude Code `2.1.205`, exit `1`) evicts the id and re-runs the same turn from full history against a fresh session. Other failures (auth, rate limit, signal) still reject with the real stderr — no silent retry.

**Proof (real CLI, real `ScenarioExecution`, real deleted session transcript):**

```
# BEFORE (adapter @ HEAD~2)
FAIL  ... > recovers a real vanished session mid-run and still recalls turn 1
Error: [ClaudeCodeAgent] Error: Claude Code CLI failed with exit code 1: No conversation found with session ID: 118eaf15-…
 ❯ ScenarioExecution.callAgent src/execution/scenario-execution.ts:1070:13
     Tests  1 failed | 43 skipped (44)

# AFTER (adapter @ HEAD)
     Tests  1 passed | 43 skipped (44)   ← run survives, turn 2 still recalls ZEBRA42
```

Full details, screenshot, and the rest of the evidence are in the PR description. Two other things this turned up and fixed:

- The committed proof screenshot showed `Tests 14 passed (14)` and contained **no multiturn test**, while captioned as proof of the 2-turn `--resume` E2E. Replaced with a capture of the runs above.
- `rate_limit_event` is emitted by the CLI on ordinary runs, so the parser warned `unknown event type` on **every** real invocation. Added to `KNOWN_EVENT_TYPES` so the drift warning means something again.

Answering the [PR Hound](https://github.com/langwatch/scenario/pull/687) author question: yes — a signal-terminated child (`exitCode === null`, non-null `signal`) rejects with the signal name plus the stderr tail; it never parses partial output as success. That path is covered, and it is deliberately **not** treated as a recoverable stale session.

`#690` is closed by this PR.

javascript/src/agents/claude-code/claude-code-agent.adapter.ts
```diff
@@ -0,0 +1,617 @@
...
+        // empty-delta guard that never contacted the CLI). Evicting here would
+        // silently downgrade the next turn to a from-scratch prompt.
+        if (!isStaleSessionError(error)) throw error;
+        // The session is genuinely gone server-side — evict the dead id.
+        this.sessions.delete(threadId);
+        if (!this.config.replayOnLostSession) {
+          // Default: fail loudly. A replay is a real modality change (it loses
+          // server-side session state), so under an eval harness we do not do it
+          // silently; the caller opts in via `replayOnLostSession`.
+          throw new Error(
```
drewdrewthis: **[principles] The lost-session error breaks its own documented type contract.**

This path throws a plain `new Error(msg, { cause })`. Every *other* CLI failure — signal, non-zero exit, `is_error` — surfaces from `call()` as a `ClaudeCodeCliError`. And `isStaleSessionError` is module-private (`grep -c isStaleSessionError index.ts` → `0`), so it cannot be re-detected either.

So `catch (e) { if (e instanceof ClaudeCodeCliError) … }` **misses precisely the failure a caller most needs to branch on** — deciding whether to set `replayOnLostSession`, or reading `.stderr`. It is demoted to `.cause`.

Meanwhile `ClaudeCodeCliError`'s own docstring promises: *"a caller can tell a vanished session apart from a genuine failure."* That is false on exactly this path.

**Fix:** export `class LostSessionError extends ClaudeCodeCliError` and throw it here, carrying the actionable message, the `cause`, and the dead `sessionId`. `instanceof ClaudeCodeCliError` stays true; `instanceof LostSessionError` names the case. The API is unreleased, so this is free now and breaking after release.

javascript/src/agents/claude-code/claude-code-agent.adapter.ts
```diff
@@ -0,0 +1,617 @@
...
+const STALE_SESSION_ERROR = /no conversation found/i;
+
+/**
+ * A Claude Code CLI invocation that failed: it exited non-zero, died on a
+ * signal, or fielded `is_error: true` on its terminal `result` envelope. Carries
+ * the raw exit status and stderr AND the envelope's `subtype`/`errors` so a
+ * caller can tell a vanished session apart from a genuine failure (bad auth,
+ * rate limit, unknown model).
+ *
+ * Note `stderr` is unredacted CLI output and may echo sensitive detail (env, a
```
drewdrewthis: **[security] The sensitive-data caveat names only `stderr`, but two more untrusted fields were added.**

This delta gave `ClaudeCodeCliError` `readonly subtype?: string` and `readonly errors?: string[]` — both fielded from the same untrusted CLI output this comment warns about. Confirmed against the compiled bundle:

```
this.exitCode = exitCode;
this.signal = signal;
this.stderr = stderr;
this.subtype = subtype;   // ← new, enumerable
this.errors = errors;     // ← new, enumerable
```

So `Object.keys(err)` and `JSON.stringify(err)` include them. Neither this doc comment nor `claude-code.mdx`'s Error-handling field list mentions `subtype` or `errors` at all — a reader following the docs would not know they exist, let alone that `errors` carries the same risk as `stderr`.

**Fix:** extend the caveat in both places, and document what the two fields are.

Related, non-blocking: `errors[]` is now the *preferred* source embedded into `.message`, and `langwatch@0.16.1`'s `withActiveSpan` rejection handler calls `setStatus({ message: err.message })` + `recordException(err)` unconditionally, which OTel copies into span attributes. Tracked in #754.

drewdrewthis (commented):
Second review pass over the applied delta `4f33293..ade0694` — the commits that implement the first pass's findings, which no reviewer had read. Two blocking findings, filed as resolvable threads so they cannot live only in the (unresolvable) verdict comment.

rogeriochaves (approved):

Issue #648, opened 2026-06-10
ElevenLabs + WebSocket voice adapters can hang on a silent/non-audio completion (audio-gated drain, no terminal path)


drewdrewthis:
**Surfaced by the `/sweep` during [PR #647](https://github.com/langwatch/scenario/pull/647)** (which fixed the same anti-pattern in the OpenAI Realtime adapter, #646). The ElevenLabs and generic WebSocket voice adapters share an **audio-gated receive/drain loop**: it returns only on an audio-delta event, so a turn that completes **without producing audio** drains to the `response_timeout` deadline and raises, instead of returning cleanly.

This is a **pre-existing latent hang, NOT a regression** introduced by #646.

## Expected

A voice-adapter turn that completes with no audio — a silent agent turn, a socket close, or (for ElevenLabs) a `client_tool_call` event with no spoken audio — should terminate the drain cleanly (e.g. return an empty `AudioChunk` so the base `drainAgentResponse` loop exits), not hang to the timeout.

## Actual

- **ElevenLabs** (`python/scenario/voice/adapters/elevenlabs.py` `recv_audio`; `javascript/src/voice/adapters/elevenlabs.ts` `receiveAudio`): the drain returns only on an `audio` event. Documented non-audio events — including `client_tool_call`, `vad_score`, `agent_response_metadata` — are silently skipped (the `elevenlabs.py:19-21` docstring explicitly notes `client_tool_call` is skipped and "the provisioned test agent doesn't trigger them"). **Verified: ElevenLabs ConvAI does emit a `client_tool_call` event at the wire**, so a real (non-test) ElevenLabs agent emitting a tool call with no spoken audio would hit the audio-gated hang. There is no socket-close / terminal path returning an empty chunk.
- **WebSocket** generic adapter (`python/scenario/voice/adapters/websocket.py` `recv_audio`, ~lines 70-80): a `while True` loop that returns only when the protocol decodes a non-`None` audio chunk; no socket-close / end-of-stream terminal path. A clean server close with no final chunk hangs to the timeout. (This generic adapter has no tool-call support.)

For reference, adapters that already handle this correctly: **Gemini Live** (`StopAsyncIteration` + `turn_complete` → empty chunk) and **Pipecat** (socket-close handler / `stop` event). The #646 fix on the OpenAI Realtime adapter is the reference pattern: on the completion event, if there is accumulated non-audio output (or on socket close), return an empty `AudioChunk` so the drain terminates.

## Reproduction

[...]

PR #691, opened 2026-06-20
fix(voice): terminate wait=False test drain on end-of-turn, re-enable in CI

minh2416294:
## Why

The `agent(wait=False)` voice tests in `tests/voice/test_agent_wait_false.py` were skipped in CI via `@pytest.mark.skipif(CI=="true")` with the note "hangs in python-ci, can't reproduce locally." The hang isn't infinite — it's a ~90s turn. The test's `_SlowAdapter.recv_audio` returns a non-empty audio chunk on *every* call, so the base `_drain_agent_response` loop (`scenario/voice/adapter.py`) never tail-silences and runs until `accumulated_audio >= response_max_duration` (30s of audio ≈ 90s of wall-clock at this cadence). That trips CI's 60s timeout reliably under load while squeaking under it on faster local machines — a textbook timing flake. This restores the lost CI coverage of the `wait=False` concurrency primitive.

Relates to #648 (same audio-gated-drain class; this PR fixes the *test fixture* side, not the product-side adapter terminal path).

## What changed

- **Fixed the test fixture to honor the real adapter end-of-turn contract.** `_SlowAdapter` now emits one real chunk, then a terminal empty `AudioChunk(data=b"")` — the same end-of-turn signal real adapters send (OpenAI Realtime, Gemini Live, Composable) so the base drain loop exits on tail silence instead of running to `response_max_duration`. Chose fixing the unrealistic mock over adding a wall-clock cap to the base loop, because #647 establishes that the drain terminates on an audio-terminal event, deliberately keeping `response_timeout` as the only time-guard.
- **Removed the `skipif(CI)` workaround** (and the now-unused `import os`). With the fixture fixed, the tests run in CI again — closing the coverage gap the skip created.
- **Added `@pytest.mark.timeout(30)`** to `test_agent_wait_false_returns_before_turn_finishes` so any regression of the ~90s drain fails fast and loud instead of wedging the runner.

## Test plan

```bash
# from python/
CI=true uv run pytest tests/voice/test_agent_wait_false.py -p no:cacheprovider -v
# → 3 passed in ~1.2s
```

- Reproduced the original hang cross-platform before the fix: Windows + Linux (WSL2 Ubuntu), Python 3.12 — `scenario.run()` with the `wait=False` script ran ~90s (measured: `RETURNED after 90.3s success=True`) and exceeded the 60s pytest-timeout.
- After the fix, the full file passes in ~1.2s on both platforms with `CI=true` set (the condition that previously skipped it).
- The non-blocking assertion is unchanged and still meaningful: `_stamp` fires before `slow.recv_returned_at` (the `wait=False` step returns control before the slow recv completes). The sibling `test_double_wait_false_in_flight_raises` race is preserved (the fixture change only affects calls *after* the first chunk).

No playable artifact — see Test plan. This is a test-fixture/control-flow change; the demonstration is the suite going from skipped-in-CI (~90s drain on the parent commit when the skip is removed) to `3 passed in ~1.2s` on this branch with `CI=true`.

### DIFF
--- a/python/tests/voice/test_agent_wait_false.py
+++ b/python/tests/voice/test_agent_wait_false.py
@@ -4,25 +4,22 @@ Unit tests for the ``agent(wait=False)`` async primitive (§4.4 L369-382).
 Verifies the concurrency contract: control returns before the agent turn
 finishes, and subsequent blocking steps await the pending turn.
 
-These tests drive ``scenario.run`` end-to-end. They pass deterministically
-locally (~2s total) with no external services, but hang indefinitely in
-the project's python-ci workflow for reasons we haven't been able to
-reproduce outside CI. Skipped under ``CI=true`` until that's fixed;
-local development still exercises them on every run.
+These tests drive ``scenario.run`` end-to-end with no external services
+(~1s total). The earlier ``skipif(CI)`` guard masked an unrealistic fixture:
+``_SlowAdapter`` returned audio on *every* ``recv_audio`` call, so the base
+``_drain_agent_response`` loop never tail-silenced and ran to
+``response_max_duration`` (~90s of wall-clock for this cadence) — tripping
+CI's 60s timeout while squeaking under it on faster local machines. The
+fixture now emits one chunk then an end-of-turn empty chunk, the contract
+every real adapter honors, so the turn terminates promptly.
 """
 
 import asyncio
-import os
 
 import pytest
 
 import scenario
 from scenario.voice import AdapterCapabilities, AudioChunk, VoiceAgentAdapter
-
-pytestmark = pytest.mark.skipif(
-    os.environ.get("CI") == "true",
-    reason="scenario.run hangs in GitHub-Actions python-ci; runs fine locally",
-)
 
 
 class _SlowAdapter(VoiceAgentAdapter):
@@ -32,6 +29,7 @@ class _SlowAdapter(VoiceAgentAdapter):
         super().__init__()
         self.recv_delay = recv_delay
         self.recv_returned_at: float | None = None
+        self._spoke: bool = False
 
     async def connect(self):
         pass
@@ -43,6 +41,15 @@ class _SlowAdapter(VoiceAgentAdapter):
         pass
 
     async def recv_audio(self, timeout):
+        # Speak one slow chunk, then signal end-of-turn with an empty chunk so
+        # the base ``_drain_agent_response`` loop terminates on tail silence —
+        # the contract every real adapter honors. Returning audio on *every*
+        # call (the previous behavior) never tail-silences, so the drain runs
+        # to ``response_max_duration`` (~90s of wall-clock for this cadence),
+        # which is what wedged the turn in CI.
+        if self._spoke:
+            return AudioChunk(data=b"")
+        self._spoke = True
         await asyncio.sleep(self.recv_delay)
         self.recv_returned_at = asyncio.get_event_loop().time()
         return AudioChunk(data=b"\x00\x00" * 2400)
@@ -55,9 +62,15 @@ class _CannedUser(scenario.AgentAdapter):
         return "hi"
 
 
+@pytest.mark.timeout(30)
 @pytest.mark.asyncio
 async def test_agent_wait_false_returns_before_turn_finishes():
-    """A wait=False agent step must not block until the slow recv completes."""
+    """A wait=False agent step must not block until the slow recv completes.
+
+    The ``timeout`` guards the regression this fixture once caused: an adapter
+    that never tail-silences makes the drain run to ``response_max_duration``
+    (~90s), so a stuck turn fails fast and loud here instead of wedging CI.
+    """
     slow = _SlowAdapter(recv_delay=0.3)
 
     returned_at: float | None = None


minh2416294:
@coderabbitai review

minh2416294:
Thanks! Applied `self._spoke: bool = False` to match the adjacent `recv_returned_at: float | None` annotation in the same fixture.

I skipped the `-> None` on the test function: none of the 203 `async def test_*` functions under `tests/voice/` annotate a return type, so adding it here would diverge from the suite's established convention rather than match it. Happy to add it if the maintainers prefer the explicit form.

langwatch-agent:
🐕 **PR Hound review brief** — assigned reviewer @drewdrewthis

**Review mode:** Fast Skim
**Scariest prod risk:** none found (test-only fixture change, single file, CodeRabbit clean).

- `[Must Check]` `test_agent_wait_false.py` — the whole point of the test is that `scenario.run()` returns *before* the turn finishes; with the mock now tail-silencing after one chunk the turn completes in ~1.2s, so confirm the `_stamp` fires before `slow.recv_returned_at` assertion is still meaningfully non-trivial and not passing by accident.
- `[Probably Fine]` `_spoke` flag on `_SlowAdapter` — verify the sibling `test_double_wait_false_in_flight_raises` race still exercises an in-flight recv (the empty-chunk only affects calls after the first).

- Does emitting an empty `AudioChunk` on every subsequent call still faithfully model the real adapter end-of-turn contract, or could it mask a regression where the product-side drain (issue #648) fails to terminate on non-mock adapters?

minh2416294:
Thanks for the brief — went back through the three points against the code.

**1. Is `returned_at < slow.recv_returned_at` still meaningful, or passing by accident?**

Still meaningful. The tail-silence change only affects *when the drain loop exits*, not *when the first recv stamps*. `_stamp` runs synchronously right after `scenario.agent(wait=False)` returns control; `slow.recv_returned_at` is only set after the `await asyncio.sleep(0.3)` inside the first `recv_audio`. The `sleep(0.5)` step then lets that first recv actually finish, so `recv_returned_at` is populated by assert time. The ordering assertion is testing the `wait=False` primitive itself (control returns before the slow recv completes), which is orthogonal to the empty-chunk termination. Dropping the fixture's turn from ~90s to ~1.2s doesn't touch that ordering — the 0.3s recv delay is what the stamp races, and it's unchanged.

**2. Does `test_double_wait_false_in_flight_raises` still exercise an in-flight recv?**

Yes. The empty chunk only returns on the *second* `recv_audio` call (gated by `_spoke`). The first call still sleeps `recv_delay=0.5` and is in-flight when the second `agent(wait=False)` is scheduled, so the "already in flight" RuntimeError fires on a genuinely pending recv. The race is preserved.

**3. Does empty-chunk-on-every-subsequent-call faithfully model the end-of-turn contract, or could it mask an #648 regression?**

It's faithful, and it can't mask #648. The base `_drain_agent_response` loop (`scenario/voice/adapter.py`) breaks on the *first* empty chunk:

```python
while accumulated < self.response_max_duration:
    try:
        nxt = await self.recv_audio(timeout=self.response_tail_silence)
    except asyncio.TimeoutError:
        break
    if not nxt.data:
        break        # <- exits here on the first data=b""
    ...
```

So whether the mock returns empty once or on every later call is immaterial to the drain — it exits on the first `data=b""`, which is exactly the end-of-turn signal real adapters send. Returning empty on every subsequent call is just defensive (nothing calls `recv_audio` again after the break), not a behavior the loop depends on.

On #648: that's the *product-side* path — a real adapter whose recv never yields tail silence, where the drain has no wall-clock guard beyond `response_max_duration` (30s). This PR only fixes the *test fixture* to honor the contract; it doesn't touch the product-side drain, so it can't hide a regression there. If anything the mock is now *more* faithful to the real contract than the previous "audio on every call" version, which is what created the 90s CI hang in the first place.

One note on the CodeRabbit docstring-coverage warning (16.67% < 80%): I left it as-is. The flagged symbols are test functions and the `_CannedUser`/`_SlowAdapter` stub methods — none of the ~200 `async def test_*` functions under `tests/voice/` carry docstrings or return annotations, so adding them here would diverge from the suite's convention rather than match it (same reasoning as the `-> None` skip above). Happy to add them if you'd prefer the explicit form.

drewdrewthis (approved):
lgtm

minh2416294:
@drewdrewthis thank you brother

minh2416294:
﻿This one is approved by @drewdrewthis, CI is green, and it shows as mergeable. It just needs a merge. @rogeriochaves would you be able to take a look? Happy to rebase if that helps. Thank you.

drewdrewthis:
## CI regression diagnosis

**Root cause: main-moved breakage, not a problem with this PR's code.**

The 3 failing checks are caused by an infra gap in `python-ci.yml` that surfaces when a fork PR's merge commit includes upstream's `python/examples/test_testing_remote_agents_*.py` files.

### What happened

1. PR branch (`flake/diagnose`) is 11 commits behind main — the first run was `action_required` on Jun 20, so CI never ran until drewdrewthis approved on Jul 6.
2. By then, the merge commit (`95153b2`) includes upstream main's `python/examples/test_testing_remote_agents_{sse,stateful,streaming}.py`.
3. Each of those files calls `client = AsyncOpenAI()` at **module level** — pytest collection fails with `Missing credentials` before any test body runs.
4. Fork PRs get empty secrets (`OPENAI_API_KEY` = `""`), so the `Test (Examples)` step always fails for fork PRs that hit these files.

### The 3 failing checks

| Check | Cause |
|-------|-------|
| `test (3.12)` | `pytest examples/` collection error — `openai.OpenAIError: Missing credentials` on 3 files |
| `python-complete` | Aggregator fails because `test` job failed |
| `Validate PR Title` | `Resource not accessible by integration` — a GitHub permissions issue on fork PRs with the `amannn/action-semantic-pull-request` action; unrelated to the title format (title is valid conventional commit) |

### Fix needed in `python-ci.yml` (maintainer action)

Add an `if:` guard to the `Test (Examples)` step so it skips on fork PRs:

```yaml
- name: Test (Examples)
  if: github.event.pull_request.head.repo.full_name == github.repository
  run: uv run pytest examples/ -v --tb=short --timeout=300
  env:
    LANGWATCH_API_KEY: ${{ secrets.LANGWATCH_API_KEY }}
    OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
    GEMINI_API_KEY: ${{ secrets.GEMINI_API_KEY }}
  working-directory: python
```

The `python-complete` aggregator job also needs to handle this — the `if: always()` aggregation already handles `skipped` as a success, so adding the `if:` guard above is sufficient.

### This PR's code is clean

The PR changes (`tests/voice/test_agent_wait_false.py` fixture fix + skipif removal) are unaffected. Once the `python-ci.yml` guard is merged to main and a rebase is done, CI will pass cleanly. Alternatively, a maintainer can push the guard fix to main and re-trigger CI on this PR.

drewdrewthis's approval should remain valid — no code in this PR changed.

langwatch-agent (commented):
No code/security findings from static external-PR review. The fixture change matches the documented end-of-turn contract and narrows the test-only CI skip. I did not execute fork code. Residual blockers: current CI shows failing `Validate PR Title` and `test (3.12)` / `python-complete`, so those need to be fixed or explained before merge.

rogeriochaves (approved):

rogeriochaves:
@minh2416294 merged! Sorry for the delay, and thank you very much for the contribution!

minh2416294:
thank you

Issue #721, opened 2026-07-02
Extract an OpenCodeServer collaborator class for server lifecycle (serverPromise/ensureClient/close)


drewdrewthis:
## Description

Extract an `OpenCodeServer` collaborator class for the server lifecycle (`serverPromise` / `ensureClient` / `close`) — these go inert under an injected client and are cleanly extractable from the adapter. Reviewer "New Issue candidate" on **PR #701**; never filed.

## Source

> server-lifecycle as a collaborator — serverPromise/ensureClient/close go inert under an injected client; extractable to an OpenCodeServer class. New Issue candidate.

(PR #701 review verdict.)

---
*Classification: Tech-debt (refactor)*
*Filed from a PR follow-up sweep (source: PR #701). Stub — not-ready until triaged.*

langwatch-agent:
Checked against `origin/main` as a tech-debt run candidate. **Blocked on an unmerged PR, so skipping.**

The code this ticket refactors is not on main:

```
$ git grep -rln "opencode\|OpenCode" origin/main
(no output)

$ git grep -n "serverPromise\|ensureClient" origin/main -- javascript/src
(no output)
```

`serverPromise` / `ensureClient` / `close` only exist inside #701 (`feat(agents): add OpenCode agent adapter`), which is still **OPEN**. Extracting an `OpenCodeServer` collaborator now would mean either branching off an unmerged PR or reimplementing the adapter, and either way the extraction would have to be redone against whatever #701 looks like when it lands.

Worth re-picking once #701 merges: the ticket is well scoped and the seam it names (server lifecycle going inert under an injected client) is a real one.

langwatch-agent:
Checked as a candidate for today's tech-debt run and skipped: blocked on an unmerged PR.

The class this asks to be extracted from does not exist on `main`. PR #701 (`feat(agents): add OpenCode agent adapter`, branch `feat/5001-opencode-adapter`) is still OPEN, and against `origin/main` at `32abdb17` there is no OpenCode adapter at all:

```
git ls-tree -r origin/main --name-only | grep -i opencode   # no matches
git grep -l serverPromise origin/main                       # no matches
```

So `serverPromise`, `ensureClient` and `close` have no home to be extracted from yet. This wants to follow #701, and is probably best done inside it or immediately after it merges, while the shape is fresh. Leaving it open and untouched.

---
[...]

Issue #720, opened 2026-07-02
Migrate legacy OpenTelemetry 1.x subtree to 2.x to unblock flooring opentelemetry-core to >=2.8.0


drewdrewthis:
## Description

Migrate the legacy OpenTelemetry 1.x SDK subtree (pulled transitively via the langwatch SDK dependency) to 2.x, so `@opentelemetry/core` can be floored to `>=2.8.0`. Called out as "Not included" in **PR #702**; never filed.

## Source

> an older OTel 1.30.1 subtree (pulled transitively via the langwatch SDK dependency) ... Flooring core to 2.8.0 needs that 1.x subtree migrated to 2.x first, which is a separate, larger change.

(PR #702 "Not included".)

## Context

This unblocks a security/version floor that PR #702 could not land because of the transitive 1.x subtree.

---
*Classification: Tech-debt (dependency migration)*
*Filed from a PR follow-up sweep (source: PR #702). Stub — not-ready until triaged.*

Issue #516, opened 2026-05-21
chore(#372): retrofit spec-binding + process-gates on TS voice parity PRs


drewdrewthis:
## Description

Retrofit the TS voice parity slice plan (issue #372 sub-task) so PRs actually bind to `specs/voice-agents.feature` via a Gherkin runner instead of paraphrasing scenario line ranges in `describe()` names, and so each PR clears the `langwatch/*` process-gate chain (`/prove-it` → `/review` → `/gh-tag ai-reviewed`) before flipping ready.

## Context

PR #511 ([link](https://github.com/langwatch/scenario/pull/511), merged) and PR #513 ([link](https://github.com/langwatch/scenario/pull/513), draft) of the TS voice parity slice plan shipped or staged with two claims that did not hold:

- **Paraphrase masquerading as binding.** Tests claimed to be "bound to `specs/voice-agents.feature` lines XX-YY", but the binding lived only in `describe()` names — the feature file was never loaded or executed by the test runner. Test bodies were hand-written assertions that paraphrased the spec.
- **Process-gate skipped.** Neither PR went through `/prove-it` → `/review` → `/gh-tag ai-reviewed`. Neither carried the `ai-reviewed` label at merge/ready time. The gate was bypassed implicitly because there is no branch-protection enforcement — flipping `ai-reviewed` on relies on the slice lead invoking `/gh-tag` before `gh pr ready`.

Without a real binding mechanism and without the gate-chain being honored, every remaining PR in the slice plan (PR3 through PR12) inherits the same failure mode by default.

The 2026-05-21 PR #511 merge is the canonical case for the broader "claim-of-done without verification" anti-pattern (see the consuming codex's `mistakes.jsonl` and `references/common-mistakes.md`).

Related: parent slice plan is tracked in #372.

## Scope

Affected areas in `langwatch/scenario`:

- `javascript/` — no Gherkin runner is currently installed (no `vitest-cucumber`, no `@cucumber/cucumber`, no manual Gherkin step loader). All scenario "bindings" in the slice plan PRs are currently paraphrase-only.
- `javascript/tests/voice/` (or wherever the voice-parity PR #511 / PR #513 tests live) — the test files reference `specs/voice-agents.feature` by line range in `describe()` names but contain hand-written assertions that do not load the feature file.
- `specs/voice-agents.feature` — source of truth for scenarios that future PRs must bind to via the runner, not via describe-name paraphrase.
[...]

PR #728, opened 2026-07-06
test(voice): harden audio-derived STT assertion against a transient outage (#725)

### LINKED ISSUE #725: Harden audio-derived STT assertion to fail honestly under a transient STT outage
## Description

Harden the audio-derived STT assertion so it fails honestly under a transient STT outage: null out the `userSegs` transcripts before re-running STT (a ~1-line test-scope change), so a silent STT outage cannot green the test. Reviewer recommends pulling this in as a fast follow-up to **PR #713**; never filed.

## Source

> a 1-line test-scope hardening (for (const s of userSegs) s.transcript = undefined;...) would make it fail honestly under an STT outage. Recommend pulling this in ... as a fast follow-up.

(PR #713 review verdict, non-blocking #1.)

---
*Classification: Tech-debt (test hardening)*
*Filed from a PR follow-up sweep (source: PR #713). Stub — not-ready until triaged.*

langwatch-agent:
## Ticket

Closes #725. The voice example tests prove a turn is audio-derived by force-running STT over the recorded user bytes and asserting each user segment has a non-empty transcript. A transient STT outage could green that assertion on a stale transcript instead of failing honestly.

## Priority pick

Top-priority deliverable in langwatch/scenario after excluding the migration (#720) and the process-meta retrofit (#516); the other P2 refactor (#721) is blocked on the still-open PR #701 that introduces the code it targets.

## Root cause (audited)

`transcribeSegments` (javascript/src/voice/transcribe.ts) catches a per-segment provider throw, logs a warning, and leaves `transcript` untouched (documented graceful-degrade). With `{ onlyMissing: false }` it re-targets every segment, but on a throw it never overwrites. So a user segment that already carries a transcript from the live run keeps that text when STT suffers a transient outage, and the example assertion `(s.transcript ?? "").trim().length > 0` passes on stale data.

## Change

Clear the user-segment transcripts immediately before the forced re-run, in all four permutation tests that use this pattern:

- `elevenlabs-hosted.test.ts`, `openai-realtime-agent.test.ts`, `pipecat-ws.test.ts`, `gemini-live.test.ts`

After clearing, a healthy re-run fills a fresh audio-derived transcript, while an outage leaves it empty and fails the assertion honestly. Scoped to the tests; no product code changed.

## Evidence

New unit test `javascript/src/voice/__tests__/transcribe-stale-transcript.test.ts` exercises the real `transcribeSegments` path with a throwing provider and locks the mechanism:

```
Test Files  1 passed (1)
Tests  3 passed (3)   EXIT=0
```

- **stale transcript survives an outage** (documents the hole): with a live-run transcript and a throwing provider, `transcript` is unchanged and the non-empty check would still pass.
- **cleared transcript fails honestly**: after the clear, a throwing provider leaves `transcript` undefined and the non-empty check is 0.
- **happy path preserved**: after the clear, a healthy provider fills the audio-derived transcript and the check passes.

Typecheck clean in both workspaces (`javascript` and `examples/vitest`), EXIT=0. No em/en dashes; no lockfile churn.

The four edited files are env-gated real-audio E2E demos (require the STT/agent provider keys), so their full run is a reviewer/CI step; the unit test above proves the hardening mechanism deterministically without those keys.

## Advisory note

### DIFF
--- a/javascript/examples/vitest/tests/voice/elevenlabs-hosted.test.ts
+++ b/javascript/examples/vitest/tests/voice/elevenlabs-hosted.test.ts
@@ -237,6 +237,11 @@ if (hasHostedKey || hasComposableKey) {
             }
             // Audio-DERIVED transcript proof: force STT over the recorded user
             // bytes (onlyMissing:false) and require non-empty speech per turn.
+            // Clear any transcript carried over from the live run: a forced
+            // re-run leaves a stale transcript in place if STT throws (a
+            // transient outage), so without this the check below could green
+            // on old text instead of the freshly audio-derived transcript.
+            for (const s of userSegs) s.transcript = undefined;
             await voice.transcribeSegments(
               { segments: userSegs, timeline: [] },
               { onlyMissing: false },
--- a/javascript/examples/vitest/tests/voice/gemini-live.test.ts
+++ b/javascript/examples/vitest/tests/voice/gemini-live.test.ts
@@ -138,6 +138,11 @@ describeFeature(
           }
           // Audio-DERIVED transcript proof: force STT over the recorded user
           // bytes (onlyMissing:false) and require non-empty speech per user turn.
+          // Clear any transcript carried over from the live run: a forced
+          // re-run leaves a stale transcript in place if STT throws (a
+          // transient outage), so without this the check below could green
+          // on old text instead of the freshly audio-derived transcript.
+          for (const s of userSegList) s.transcript = undefined;
           await voice.transcribeSegments(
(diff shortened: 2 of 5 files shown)


langwatch-agent (commented):
No findings. The four forced STT proof sites now clear live-run transcripts before `onlyMissing:false`, so an STT outage leaves an empty transcript instead of passing on stale text. I also checked for other forced-STT assertion sites and ran the new regression test locally (3 tests passed); GitHub checks are green.

PR #748, opened 2026-07-09
fix(voice): reconcile EL audioQueue at turn boundaries (#747)

### LINKED ISSUE #747: EL voice adapter: stale audioQueue bleeds split utterances across turns (doubled greeting, transcript/audio desync)
# EL voice adapter: stale `audioQueue` bleeds split agent utterances across turns — doubled greeting, transcripts unpaired with audio, 0-audio turns

## For Humans

**The ElevenLabs voice adapter never reconciles its `audioQueue` at turn boundaries, so a split agent utterance (delivery gap > `responseTailSilence`, or the `responseMaxDuration` cap) leaks its remainder into a fake next agent turn — and every turn after that desyncs.** Customers with scripted voice scenarios (`[agent(), proceed(...)]`, hosted EL) see the judge fail with "assistant repeated its greeting" even though the live agent behaves correctly. Reproduced deterministically (2/2 + clean-shaped control).

**Ask:** fix the root cause — one agent utterance must land as one agent message, with its own transcript, regardless of TTS chunk pacing or utterance length.

- Interim workaround (masks, does not fix): `elAgent.responseTailSilence = 2.0`.
- Deterministic repro: `firstMessageOverride` greeting whose TTS exceeds 30s.

_Status: investigated (mechanism verified in code + live repro); ACs + plan + spec below. Full spec in the collapsed block._

## Expected

One agent utterance = one agent message: all of its audio and its matching transcript in a single history entry, regardless of TTS chunk pacing or utterance length.

## Actual

Split utterance produces two agent messages: part 1 with the full-utterance transcript, then (after the user's turn) part 2 — the audio remainder — carrying the SAME transcript text. Downstream turns then desync: truncated "…" transcripts that don't match their audio, transcript-only messages with 0s audio, and audio-only messages with empty transcripts. The judge reads the doubled text and fails the run.

## Reproduction

Live against a hosted EL agent, `@langwatch/scenario` 0.5.1, script `[scenario.agent(), scenario.proceed(6)]`, stock defaults:

1. Configure `firstMessageOverride` with a greeting whose TTS exceeds 30s (`responseMaxDuration`) — this is the deterministic way to split an utterance; any inter-chunk gap >0.6s splits it the same way.
2. Run. Observed 2/2:
   - msg1 assistant: 30.4s audio (capped at exactly `responseMaxDuration`)
   - msg2 user turn
   - msg3 assistant: **19.5s audio = the greeting remainder**, same greeting transcript again, arriving instantly (per-chunk instrumentation: every drained chunk gap=0ms/waited=0ms — pure queue drain; a real reply takes 4–12s)
   - subsequent turns: truncated "…" transcripts, an 11.6s-audio message with EMPTY transcript
3. Control with `responseTailSilence = 2.0`: greeting once, coherent turns, run passes (remainder still bleeds into the next turn's audio — masked, not fixed).

## Investigation (verified, file:line at `javascript/v0.5.1`)

- `audioQueue` declared `voice/adapters/elevenlabs.ts:340`; only ever push (`:554`) and shift (`:778`) — **no clear/flush at turn boundaries**.
- On-connect greeting is only buffered (`onAgentAudio` → queue, `elevenlabs.ts:536-555`); transcript stashed on `lastAgentTranscript` (`:486-503`). `scenario.agent()` does not request a turn — it drains (`voice/adapter.runtime.ts:247`, sendAudio only `if (incoming)` `:263-265`).
- `drainAgentResponse` closes a turn after `responseTailSilence` = 0.6s of no audio (`adapter.ts:79`); `responseMaxDuration` = 30s silently truncates mid-utterance leaving the remainder queued.
- Hosted EL delivers `firstMessageOverride` TTS in a single burst (measured max inter-arrival gap 257ms even with a 3s SSML break — break is silence inside the bytes, not an arrival gap). Splits in the wild come from **multi-part agent turns** (LLM generation gaps between TTS segments) or the 30s cap.
- Transcript pairing: the turn's text comes from EL response events, not re-derived from the audio actually drained — so both halves of a split utterance carry the same text, and the affected-run evidence shows both greeting entries with identical truncated text while audio #1 is the first half (transcribed: ends at a clean sentence boundary) and audio #2 is the remainder ("Or if you'd prefer…").
- History append paths checked (`scenario-execution.ts:1041,1093,1202,1804,2246`) — each appends once; this is NOT a double-append. Byte-identical repeated user audio in affected runs is the TTS cache (`voice/tts/tts.ts:40-42`, keyed sha256(text)+voice) when the sim re-states the same opener after seeing the fake re-greet.
- Platform/storage exonerated: stored message count exactly equals raw SDK snapshot events (17=17, 3=3) for affected runs.
- Barge-in interaction (verified at HEAD for the fix design): `fireUserInterrupt` pushes user audio via `adapter.sendAudio` while a non-blocking agent drain is in flight (`scenario-execution.ts`), so any boundary reconciliation hooked inside `sendAudio` would race an active drain — reconciliation must key off the runtime's own turn sequencing, not the adapter's transport entry point.

Related but distinct: #623 (scriptless agent-first / turn-1 greeting collision). This issue applies even WITH the documented `agent()`-first script. #707's `turnCommitMode` touches only the OpenAI-Realtime adapter + executor, not the EL adapter or the shared drain.

## Fix directions (input to the plan — not prescriptive)

1. Reconcile `audioQueue` at turn boundaries (naive clearing would DROP the tail — it must be attributed, not discarded).
2. Pair transcripts with the audio actually drained in the turn.
3. `responseMaxDuration` must not silently truncate mid-utterance and leave the remainder queued.

## Acceptance Criteria

Terminology: "before-run" = the deterministic repro (step 2 above) on the pre-fix code; "after-run" = the same script/config on the fixed code, run at least TWICE (both must satisfy the ACs — EL silent-flakiness #708 makes single runs unconvincing). "Whisper cross-check" = transcribing a message's own audio bytes with Whisper and comparing against that message's transcript field with normalized token Jaccard ≥ 0.7; "suffix of a text" = its last ≥8 words in order.

- [ ] **AC1 — split utterance lands as one message.** In each after-run, the `firstMessageOverride` greeting (TTS > 30s — trigger-fired guard: the greeting message's audio-seconds ≥ `responseMaxDuration` + 5, proving the run actually exercised the split trigger) is exactly ONE assistant history message: its whisper transcript contains the greeting's final sentence (the sentence that in the before-run appears only in the fake remainder turn), no later assistant message contains greeting audio or repeats the greeting text, and the run's judge verdict is PASS (the before-run's judge fails on the doubled greeting — the after-run must close that exact loop). Evidence: before/after live-run message dumps (role, audio-seconds, transcript per message) + judge verdicts embedded in the PR body.
- [ ] **AC2 — no fake instant turn / no stale bleed.** In each after-run, every assistant message's audio whisper-cross-checks against its OWN transcript field (token Jaccard ≥ 0.7), and no assistant message's audio transcribes to a suffix (≥8 trailing words) of the PREVIOUS assistant message's text (the before-run's msg3 signature). Evidence: per-assistant-message whisper table in the PR body.
- [ ] **AC3 — boundary reconciliation is attributed, not silent.** Given agent audio is still queued when the next user turn is committed (forced deterministically in a test), (a) the next assistant message contains none of that stale audio; (b) the stale audio is preserved in the run's recording attributed to the utterance that produced it (previous agent segment / manifest entry grows by exactly the stale bytes), not deleted; (c) a warning naming the reconciliation (with the reconciled duration) is logged; (d) **gap-split never becomes a fake turn** _(re-scoped post-challenge — see Design revision)._ A fake-adapter test delivering ONE utterance as two bursts (transcript with burst 1, spanning both, no user turn between): when both bursts are immediately available the un-chop drains them into ONE message; when the second burst arrives after the turn already closed it is **never emitted as a new agent turn** and is preserved in the recording attributed to the prior utterance. The literal "retroactively merge a late burst into the already-broadcast/judged history message" is NOT claimed — that message is immutable. Evidence: unit/integration tests asserting (a)+(b)+(c) and the merge-when-available / no-fake-turn-when-late guarantee.
- [ ] **AC4 — `responseMaxDuration` no longer silently truncates.** (a) An utterance longer than `responseMaxDuration` with continuous delivery lands whole in the turn's message (test with >cap continuous chunks; live AC1 run). (b) A genuinely non-terminating stream (fake adapter that never goes silent) still terminates the drain WITH an explicit warning — no infinite wedge, no silent cap — retaining at most 2× `responseMaxDuration` of audio (the test pins the implemented bound at or below this ceiling). Evidence: unit tests for (a) and (b) asserting message content, termination, bound, and warning.
- [ ] **AC5 — transcript↔audio pairing.** In each after-run: no two assistant messages carry the same EL `agent_response` text for one utterance; no assistant message carries a transcript whose audio lives in a different message (AC2's cross-check); a turn whose transcript event never arrives still falls back exactly as today (audio-only message; manifest STT back-fill) with no new fabrication. Evidence: live-run dump + whisper table; existing #734 grace-wait tests unmodified and green.

- [ ] **AC6 — clean runs unchanged (regression).** A standard hosted-EL multi-turn run (utterances < cap, defaults) still passes: one user + one assistant message per exchange, judge passes, and the run is coherent (agents demonstrably heard each other — coherence criterion, not counts). No added happy-path latency: median per-turn agent-response wall-clock ≤ pre-fix control median + 250 ms over ≥6 exchanges, and no single turn exceeds control-max + 500 ms. Evidence: post-fix live clean run dump + judge verdict + per-turn timing table (after vs control) in the PR body.
- [ ] **AC7 — shared-runtime consumers unaffected (regression).** The full JS test suite passes with no voice test deleted or weakened: OpenAI Realtime, Gemini Live, Twilio, Pipecat, composable, interruption/barge-in, tool-call terminal turn, #734 grace-wait, #705 pump-pause. Test-file assertion-count on `javascript/src/voice/**/__tests__/**` is non-decreasing vs the merge base; any decrease or weakening is enumerated line-by-line and justified in the PR body. Evidence: CI green on the PR + `git diff --stat origin/main -- 'javascript/src/voice/**/__tests__/**'` in the PR body.
- [ ] **AC8 — greeting flow preserved (regression).** The agent-first greeting (audio buffered on connect, drained by the first `agent()` with NO incoming user audio) is still delivered as turn 1 — reconciliation must not swallow or reattribute it. Evidence: unit test (greeting queued pre-call → first `agent()` returns it) + after-run msg1.
- [ ] **AC9 — barge-in safety (failure mode)** _(re-scoped post-challenge)._ A barge-in (`agent({wait:false})` + `user(...)`) delivering user audio mid-drain: leftover post-interrupt agent audio is **never emitted as the next agent turn and never corrupts the recording cursor**. Backward-attribution into the interrupted agent segment is deliberately NOT done here — on the barge-in path the interrupting user segment is already laid *after* the interrupted agent segment on the append-only byte cursor, so growing the agent segment would overlap it; the boundary reconcile therefore runs ONLY at the normal pre-user-`sendAudio` hook (cursor-safe), and the interrupt path relies on the drain no longer stranding audio. Evidence: existing interruption tests green + one new test that a post-interrupt leftover is not emitted as a new turn and the recording cursor stays monotonic (no overlapping segments).
- [ ] **AC10 — workaround stays valid (compat).** `responseTailSilence = 2.0` (the documented interim workaround) composes with the fix: a run with it set behaves per AC6 with no double-waiting or changed drain semantics beyond the longer tail window. Evidence: drain-timing unit test with a 2.0s tail.

Rollback axis skipped deliberately: the change is in-process SDK behavior — no schema, storage, or irreversible external writes; reverting the commit is a complete rollback.

## Plan

> **Design revision (post-challenge, 2026-07-09).** A devils-advocate pass on the first draft (epoch-stamped queue) surfaced two blockers: (i) an `agentResponseEpoch` counter is a **no-op at the reconcile hook** — the boundary fires before the next turn's `agent_response`, so leftover and the just-finished turn share an epoch and "stale ≡ epoch < current" reconciles nothing; (ii) "grow the prior agent segment" is **cursor-unsafe** anywhere the origin segment is no longer last on the append-only byte cursor — specifically on the barge-in recovery path, where the interrupting user segment is already laid after it. The design below drops the epoch counter for a single **position-keyed** reconcile at the one cursor-safe hook, and AC3(d)/AC9 are re-scoped to what an append-only cursor + an immutable already-judged message can actually deliver. The un-chop (step 1) is unchanged and already lands the customer fix.

**Approach.** The bleed has two triggers, fixed at the layer that owns each. (1) **Un-chop the shared drain** (`voice/adapter.runtime.ts:drainAgentResponse`): `while (accumulated < maxDuration)` exited mid-utterance and abandoned already-delivered frames in the EL `audioQueue`, which the next drain shifted out as a fake turn. `responseMaxDuration` becomes a **soft warn**; the drain ends only on tail-silence or the empty sentinel (as every other adapter already does), with an absolute `2 × responseMaxDuration` **ceiling** as the runaway backstop (bounded + warned, never a silent cap or wedge). Because hosted EL bursts a >30s greeting at ≤257ms inter-arrival (« the 600ms tail), the whole greeting now drains in ONE turn — this alone closes the deterministic customer bug. (2) **One position-keyed reconcile** for any residual (a >2× ceiling remainder, or a rare true gap-split): the EL adapter gains a duck-typed `reconcilePendingAudio()` that synchronously drains its queue; `defaultVoiceCall` calls it at **pre-user-`sendAudio`, before `recordUser`**, guarded by "a prior agent segment exists" (skips the greeting → AC8). At that instant the queue can hold ONLY prior-turn leftover — the user just spoke and EL has not begun replying — so the boundary **position is the staleness proof** (no epoch counter), and the prior agent segment is still last on the cursor so growing it is **cursor-safe**. EL exposes **no end-of-turn signal** (server events: `agent_response`/`audio`/`interruption`/`user_transcript`/`client_tool_call`/`ping`/`conversation_initiation_metadata`), which is why turn-end is inferred from silence and the reconcile is keyed on boundary position rather than a transport marker.

**Sequencing.**
1. **Un-chop the drain (AC4, AC1, AC2-live).** Rework the `drainAgentResponse` cap into wait-bound-plus-ready-drain + absolute ceiling + warning. New `RaceAdapter`-style unit tests: >cap continuous delivery lands whole; never-silent stream terminates at the ceiling with a warning; a 2.0s-tail run (AC10) still drains normally. Completion signal: new drain tests green, existing `transcript-grace-wait` / `duration-fidelity` tests unmodified and green.
2. **`reconcilePendingAudio()` on the EL adapter (AC3, AC5).** Add a duck-typed method that synchronously shifts+merges all currently-queued `audioQueue` chunks and returns them (null when empty). No epoch counter, no per-chunk stamp. Adapter-level unit test: queue holds leftover → reconcile returns it merged, queue cleared.
3. **One position-keyed reconcile in the runtime (AC2, AC3, AC8, AC9).** In `defaultVoiceCall`, at pre-user-`sendAudio` **before** `recordUser` — guarded by "a prior agent segment exists" (skips the greeting → AC8) — call `reconcilePendingAudio()` (duck-typed) and, if it returns audio, **grow the prior agent segment** (extend its `endTime` + append bytes + advance the cursor, cursor-safe at this hook) and log the reconcile warning. NOT called at drain-start and NOT on the barge-in path (cursor-unsafe there — AC9). Completion signal: AC3 fake-adapter test (leftover attributed, not a fake turn); interruption + greeting tests green.
4. **Prove live (AC1, AC2, AC5, AC6).** Standalone repro harness (hosted EL, `firstMessageOverride` >30s greeting, `[agent(), proceed(...)]`): capture a **before** run on the pre-fix build (doubled greeting, msg3 instant remainder) and **after** runs (≥2) on the fixed build (greeting once, whisper-cross-checked, judge PASS, per-turn timing table). Embed both in the PR body.

**Key decisions.**
- **Un-chop in the shared runtime, not the EL adapter** — the `responseMaxDuration` cap is a `VoiceAgentAdapter` field consumed by the shared `drainAgentResponse`; fixing it there fixes every audio transport, and the deterministic repro is the cap path. Rationale: the customer bug is provable and lives in shared code.
- **Position-keyed reconcile, not an epoch counter** — the boundary position (user replying to a just-completed agent turn ⇒ any queued agent audio is prior-turn leftover) IS the staleness proof. An `agentResponseEpoch` bumped on `agent_response` is a no-op at this hook (the next turn's epoch does not exist yet, so leftover shares the completed turn's epoch). Simpler, no undefined predicate, and it runs at the one cursor-safe point.
- **Reconcile ONLY at pre-user-`sendAudio` (before `recordUser`), synchronously in `call()`** — the prior agent segment is still last on the append-only byte cursor there, so growing it is cursor-safe. Drain-start and the barge-in path are cursor-UNSAFE (a user segment is already laid after the interrupted agent segment) — hooking there would overlap segments and corrupt `full.wav`. Never on `adapter.sendAudio`/`fireUserInterrupt` (would race the in-flight drain on the shared queue).
- **Duck-typed `reconcilePendingAudio()` seam** — mirrors the existing `lastAgentTranscript` convention (`adapter.runtime.ts:attachAgentTurnTranscript`), so Twilio/Pipecat/composable/OpenAI-Realtime/Gemini adapters that don't expose it pay zero cost and are untouched (AC7).
- **Absolute ceiling = `2 × responseMaxDuration`** — bounds a non-terminating stream (AC4b) while leaving generous headroom for a legitimately long single utterance; pinned by test.

**Risks / open questions.**
- *Barge-in leftover (AC9, re-scoped).* On the barge-in path backward-attribution is cursor-unsafe, so the guarantee is the weaker "leftover never emitted as a new turn + cursor stays monotonic," relying on the un-chop no longer stranding audio rather than on a reconcile. `[agent(), agent()]` with a single utterance > 2× the cap and NO user turn between is likewise not reconciled (no pre-user hook fires) — a contrived, non-customer pattern; raise `responseMaxDuration` if a single utterance legitimately exceeds 60s.
- *Gap-split real-time limit (AC3d, re-scoped).* A late continuation cannot be merged into an already-broadcast/judged message — that message is immutable. Guarantee: merged when the continuation is immediately available, else preserved in the recording + never a fake new turn.
- *Shared-drain latency (AC6/AC7).* Audit result: every shared adapter (gemini-live, pipecat, composable, openai-realtime) terminates a turn on an empty sentinel or a first-chunk timeout, and composable emits the whole turn as ONE chunk — none streams >30s of continuous audio without a sentinel, so the 2× ceiling never fires for them and does not double their latency. Recorded so review can re-check.
- *#708 silent-flakiness.* Live "after" proof requires ≥2 coherent runs (judge + whisper), never counts/201s alone — EL intermittently produces no audio; a flaky no-audio run is discarded and re-run, not counted as pass.

**Out of scope.**
- #623 (scriptless agent-first turn-1 greeting collision) — distinct root cause; this fix targets the split-utterance bleed even with the documented `agent()`-first script.
- `turnCommitMode` / OpenAI-Realtime turn plumbing (#707) — does not touch the EL adapter or the shared drain.
- langwatch/langwatch dashboard ingest — corrected history changes snapshot shape; a non-blocking downstream 201-ingest check is a follow-up, not this PR.
- Changing `responseTailSilence`/`responseMaxDuration` **defaults** — the fix makes the existing defaults behave correctly; the `responseTailSilence=2.0` workaround remains valid (AC10) but is no longer required.

## Evidence

- Repro logs: `dg2_long_run1.log`, `dg2_long_run2.log`, control `dg2_long_tail2_control.log`, arrival-pacing `dg2_break3_run1.log` (available on request; contain per-chunk timing instrumentation).
- Affected-run audio objects transcribed via whisper confirming the split (part 1 = first sentences, part 2 = remainder).

</details>

---
*Classification: Bug*
*Status: investigated — mechanism verified in code + live repro; ACs derived + ac-reviewer-sharpened (3 Must-Fix applied); plan pending.*

drewdrewthis:
## Reviewer cockpit

- **Scariest thing** — the primary fix is in the **shared** `drainAgentResponse` (`voice/adapter.runtime.ts`), which every voice adapter runs. A wrong change to the `responseMaxDuration` loop could alter turn segmentation for OpenAI-Realtime / Gemini / Twilio / Pipecat / composable. Mitigated: those adapters all terminate a turn on an empty sentinel (audited below), so the cap-as-terminator was never load-bearing for them; the full JS suite (906 tests) is green; the reconcile seam is EL-only (duck-typed).
- **Start here** — `voice/adapter.runtime.ts` `drainAgentResponse` (the old `while (accumulated < maxDuration)` chop → soft-warn + `2×` ceiling) and `reconcilePriorAgentAudio` (the position-keyed boundary reconcile). Those two are the whole change; `elevenlabs.ts` just adds the `reconcilePendingAudio()` seam.

## Why

Closes #747. In the hosted-ElevenLabs voice adapter, a split agent utterance bled its remainder into a **fake next agent turn**: the shared drain chopped a turn at `responseMaxDuration`=30s and abandoned already-delivered audio in the adapter's `audioQueue`, which the next `agent()` drain shifted out instantly as the start of a new turn. Downstream, transcripts paired with the wrong audio and the judge failed the run for "the assistant repeated its greeting" — even though the live agent behaved correctly.

## What changed

- **Un-chop `responseMaxDuration` in the shared drain** → alternative was raising the cap or a per-adapter flag → the cap becomes a soft **warn**; the drain now ends only on tail-silence or the empty sentinel (as every other adapter already does), with an absolute **`2 × responseMaxDuration` ceiling** as the runaway backstop. Blast radius is the shared drain, but the cap-as-terminator was dead weight for the other adapters (they end on a sentinel), so this only *stops chopping* a legitimately long single utterance.
- **Position-keyed turn-boundary reconcile** (not an epoch counter) → a devils-advocate pass showed an `agentResponseEpoch` is a no-op at the boundary (leftover shares the just-finished turn's epoch) and "grow the prior segment" is cursor-unsafe on the barge-in path → so the reconcile runs at exactly one hook: pre-user-`sendAudio`, **before** `recordUser`, where the prior agent segment is still last on the append-only byte cursor (cursor-safe) and any queued audio can only be prior-turn leftover (position = staleness proof). It grows the prior agent segment instead of letting the next drain emit the leftover as a fake turn.
- **AC3d / AC9 re-scoped** to what an append-only cursor + an already-broadcast (immutable, already-judged) message can deliver: leftover is merged when immediately available, else **never emitted as a new turn** and preserved in the recording; barge-in keeps a monotonic cursor (no backward mutation). See the issue's "Design revision" note.

## How it works

```
voice/adapter.runtime.ts
├── drainAgentResponse          responseMaxDuration → soft warn; ends on tail-silence
│                               / empty sentinel; 2x absolute ceiling backstop (+warn)
└── reconcilePriorAgentAudio    pre-user-sendAudio (before recordUser), guarded by
                                "prior agent segment is last on the cursor"; grows it
                                with reconcilePendingAudio()'s drained leftover (+warn)
voice/adapters/elevenlabs.ts
└── reconcilePendingAudio()     duck-typed seam: synchronously drain+merge the queue
```

## Test plan

- `specs/voice-el-audioqueue-turn-boundary.feature` — 12 scenarios, AC coverage map at the bottom; all 10 issue ACs mapped.
- `src/voice/__tests__/el-audioqueue-turn-boundary.test.ts` — 7 tests: AC4a (utterance > cap lands whole), AC2 (no bleed into next turn), AC4b (never-silent stream bounded at 2x + warn), AC3 (leftover attributed to prior segment, no bleed, warned, monotonic cursor), AC8 (greeting never reconciled), AC9 (cursor-unsafe shape refused), AC10 (tail=2.0 drains whole).
- Full JS suite: `906 passed | 1 skipped` (the skip is the hosted-EL live e2e that self-skips without creds). Typecheck clean.

## Human verification

Independent steps a skeptical reviewer can run to re-confirm the fix — this is *how you'd check it yourself*; the demonstration I already did is in the next section.

**See it (before = split / after = whole):**

Before/after waveform: pre-fix greeting split across two agent turns vs post-fix one continuous greeting

**Hear it** (raw hosted-EL audio, in `javascript/examples/vitest/outputs/recordings/747-audioqueue-turn-boundary/`):
- BEFORE — greeting chopped at the 30s cap: [`before-greeting-chopped-at-30s.wav`](https://github.com/langwatch/scenario/raw/c7795a087c7cdf15787f86a850b75df1ae3c4d17/javascript/examples/vitest/outputs/recordings/747-audioqueue-turn-boundary/before-greeting-chopped-at-30s.wav) — then its remainder as a **fake agent turn**: [`before-fake-remainder-turn.wav`](https://github.com/langwatch/scenario/raw/c7795a087c7cdf15787f86a850b75df1ae3c4d17/javascript/examples/vitest/outputs/recordings/747-audioqueue-turn-boundary/before-fake-remainder-turn.wav)
- AFTER — whole greeting, one segment: [`after-whole-greeting-66s.wav`](https://github.com/langwatch/scenario/raw/c7795a087c7cdf15787f86a850b75df1ae3c4d17/javascript/examples/vitest/outputs/recordings/747-audioqueue-turn-boundary/after-whole-greeting-66s.wav)

**Re-run the deterministic check (no creds needed):**
1. `cd javascript && pnpm install`
2. `./node_modules/.bin/vitest run src/voice/__tests__/el-audioqueue-turn-boundary.test.ts` → 7 passing.
3. Falsify it: `git show origin/main:javascript/src/voice/adapter.runtime.ts > src/voice/adapter.runtime.ts`, re-run the AC4a/AC2 cases → they FAIL (the bug); `git checkout src/voice/adapter.runtime.ts` to restore.

**Re-run the live sim (needs `ELEVENLABS_API_KEY` + `ELEVENLABS_AGENT_ID` + `OPENAI_API_KEY`):** point a hosted-EL agent at a `firstMessageOverride` greeting whose TTS exceeds 30s, run `[scenario.agent(), scenario.proceed(4), scenario.judge()]`, and inspect `result.audio.segments`. Pre-fix: two agent segments carrying the greeting (30.0s + 35.3s remainder), judge fails. This build: one ~66s agent segment, judge passes.

**1. Falsifiability — deterministic before/after (proven).** The regression test fails on the pre-fix runtime and passes on the fix (same test, `origin/main`'s `adapter.runtime.ts` swapped in then restored):

```
# BEFORE — origin/main's adapter.runtime.ts (pre-fix), same test:
  × AC4a: a continuous utterance longer than responseMaxDuration lands WHOLE
  × AC2:  the capped remainder does NOT bleed into the next agent turn
  Tests  2 failed | 5 skipped
# AFTER — this branch's adapter.runtime.ts, same test:
  ✓ AC4a  ✓ AC2
  Tests  2 passed | 5 skipped
```

Full #747 suite on the fix: **7 passed**. Full voice suite: **387 passed | 1 skipped**. Full JS suite: **906 passed | 1 skipped**.

**2. Live hosted-ElevenLabs voice run — before-desync / after-clean (proven).** Same harness (`firstMessageOverride` greeting whose TTS runs ~66s, well past the 30s cap; script `[agent(), proceed(4), judge()]`; independent Whisper cross-check of every agent segment; `gpt-5-mini` judge). BEFORE runs the **published pre-fix `@langwatch/scenario@0.5.1`**; AFTER runs **this branch's build**.

**BEFORE (pre-fix 0.5.1) — the split + fake-turn bleed, judge FAILS:**

| # | speaker | audio_s | transcript (first 70) | whisper jaccard |
|---|---------|--------:|------------------------|--------:|
| 0 | agent | **30.0** | Hello, and thank you so much for calling the Northwind Bank customer s… | 1.00 |
| 1 | user | 3.5 | hi i'm in a hurry can you quickly tell me my checking balance | — |
| 2 | agent | **35.3** | **If you're calling about a card, I can help with…** (the greeting's *remainder*) | 0.94 |

> Greeting **chopped at exactly 30.0s** (the cap); its second half surfaces as a **fake 35.3s agent turn** that ignores the user's balance question. Judge **success=false**: *"The agent responded by continuing/repeating a long scripted introduction… This repeats the greeting/intro and prevents the conversation from moving forward, so all three criteria fail."*

**AFTER (this branch) — greeting whole, coherent, judge PASSES (3/3 clean runs):**

| run | seg0 agent audio_s | whisper jaccard | doubled greeting? | reply on-topic? | judge |
|-----|-------------------:|----------------:|:-----------------:|:---------------:|:-----:|
| after-1 | **66.5** (one segment) | 0.97 | no | yes (6.0s, jac 1.00) | **success=true** |
| after-2 | **66.2** (one segment) | 0.96 | no | yes (9.4s, jac 1.00) | **success=true** |
| after-HEAD (`9d8bb69`, post-review-fix) | **66.6** (one segment) | 0.96 | no | yes (9.8s, jac 1.00) | **success=true** |

> after-1 judge: *"The agent did not repeat its greeting a second time and did not loop; the dialog moved forward across turns."* after-HEAD judge: *"The agent did not repeat or restate the long greeting a second time… no looping or repeated content."* The `after-HEAD` run is on the exact reviewed HEAD after the code-review fixes landed, so the demonstration reflects the current code.

**Mechanism confirmed live** (`LOG_LEVEL=warn`): `agent turn exceeded responseMaxDuration; continuing to drain in-flight audio … {responseMaxDurationS: 30, accumulatedS: 30.5}` then the whole ~66s greeting lands as one segment — the un-chop drains past the 30s cap instead of stranding the remainder. The boundary reconcile is the deterministic safety net for a residual (proven by the AC3 unit test); EL bursts the greeting fast enough that the un-chop consumes it whole, so the reconcile did not need to fire live.

## Anything surprising?

- **A high-effort self-review found and I fixed 8 defects in the first cut** (commit `9d8bb69`): the reconcile bailed before draining the queue on the **barge-in path** (the bleed survived there — now drains-first), reconciled bytes **bypassed `fireAudioChunk`** (hook/playback regression — now routed through it), the **`agent_stop_speaking` timeline event went stale** against the grown segment (now moved), the grown segment's **transcript wasn't flagged for back-fill** (now `transcriptTruncated`), a soft-cap warning edge, and the documented **`responseMaxDuration` soft-cap semantic change**. One residual limitation is documented in code: a single continuous utterance beyond the `2×` ceiling whose frames arrive after the drain returns is a point-in-time drain, not a subscription (the realistic EL-burst greeting is fully covered).
- **Python parity (sweep finding) → filed as #749.** `python/scenario/voice/adapter.py:281` has the **identical** `while accumulated < response_max_duration` chop and the same buffered-queue EL delivery — the Python voice stack shares this bug. It is **out of scope for this JS-scoped issue** (JS and Python release on separate trains; the customer hit it on JS), so it is tracked as a follow-up in #749 rather than expanding this PR.
- **AC7 (no test weakened):** the only test file touched is the new `el-audioqueue-turn-boundary.test.ts` (+422 lines); no existing voice test was modified or deleted, so assertion count strictly increased (`git diff --stat origin/main -- 'javascript/src/voice/**/__tests__/**'`).
- **AC6 (latency):** no regression observed — comparable elapsed for similar turn counts (before 26.9s vs after 26.7s), and the audit below shows the `2×` ceiling never fires for the other adapters (they end on a sentinel), so it cannot double their latency.
- **Shared-adapter audit:** gemini-live, pipecat, composable, and openai-realtime all terminate a turn on an empty sentinel (composable emits the whole turn as one chunk), so none stream >30s of continuous audio without a sentinel; the un-chop only ever *stops chopping* — it never changes where a normal turn ends.


drewdrewthis:
## Review verdict: READY

Reviewed the full PR diff (the shared drain, the EL adapter seam, tests, and the spec) plus a high-effort multi-agent pass on the pre-fix cut. Re-verdicted at HEAD `c7795a0`; the only change since the code-review pass is checked-in before/after proof artifacts (waveform PNG + .wav), no source change (`git diff 9d8bb69..HEAD` touches only `outputs/recordings/747-audioqueue-turn-boundary/`).

**What was reviewed and how it landed:**
- A 17-agent adversarial review of the first cut surfaced **8 defects**; **6 were code-fixed** in `9d8bb69` and covered by tests (the barge-in-path bleed → now drains-first; the `fireAudioChunk` bypass → routed; the stale `agent_stop_speaking` event → moved; the transcript back-fill flag; a soft-cap warning edge) and **2 are documented** (the intended `responseMaxDuration` soft-cap semantic change, and a known >`2×`-continuous residual).
- **Correctness:** the reconcile is cursor-safe (drains unconditionally to prevent bleed; attributes only when the prior agent segment is last on the append-only byte cursor). The un-chop's `2×` ceiling bounds a never-silent stream (AC4b).
- **Regression surface:** every other voice adapter terminates a turn on an empty sentinel, so the shared-drain change never chops a normal turn; the reconcile seam is EL-only (duck-typed). Full voice suite green; no existing test weakened.
- **Proof:** deterministic before/after (regression fails pre-fix, passes post-fix) + 3/3 clean live hosted-EL runs (whole 66s greeting, judge PASS), the last on the reviewed HEAD; a before/after waveform PNG + playable .wav are embedded under **## Human verification**.

**Gate state:** `ci-checks` green on HEAD; CodeRabbit review completed with 0 unresolved threads; no review-opened threads outstanding.

_Note: this verdict is an agent self-review artifact. Human review is requested from @Aryansharma28 (branch protection requires a non-author approval); @drewdrewthis (author/assignee) does the merge — never auto-merged._

javascript/src/voice/adapter.runtime.ts
```diff
@@ -411,6 +447,123 @@ function resetAgentTurnTranscript(adapter: VoiceAgentAdapter): void {
...
+  if (typeof reconcile !== "function" || !state) return;
+  // Drain the queue FIRST, unconditionally — this is what prevents the bleed. The
+  // attribution below is best-effort and never gates the drain.
+  const leftover = reconcile.call(adapter);
+  if (!leftover || leftover.data.length === 0) return;
+
+  const recording = state.voiceRecording;
+  const last =
+    recording && recording.segments.length > 0
+      ? recording.segments[recording.segments.length - 1]!
```
0xdeafcafe: safe `!`?

0xdeafcafe: can use `.at(-1)` now, i believe?

javascript/src/voice/adapter.runtime.ts
```diff
@@ -411,6 +447,123 @@ function resetAgentTurnTranscript(adapter: VoiceAgentAdapter): void {
...
+ */
+function moveAgentStopSpeaking(
+  state: VoiceExecutorState,
+  oldTime: number,
+  newTime: number,
+): void {
+  const timeline = state.voiceRecording?.timeline;
+  if (!timeline) return;
+  for (let i = timeline.length - 1; i >= 0; i--) {
+    const event = timeline[i]!;
```
0xdeafcafe: a guard here too?

0xdeafcafe (dismissed):

javascript/src/voice/adapter.runtime.ts
drewdrewthis: Fixed in 2db27ea — `!` removed, replaced with a real narrowing guard.

```ts
const segments = state.voiceRecording?.segments;
const last =
  segments && segments.length > 0
    ? segments[segments.length - 1]
    : undefined;
```

`last` is now `AudioSegment | undefined`, and the pre-existing
`if (!last || last.speaker !== "agent")` below narrows it — so the
cursor-unsafe path (no recording / greeting / barge-in) still warns and
drops rather than trusting an assertion. Behavior unchanged; AC8 + AC9
cover both branches.

javascript/src/voice/adapter.runtime.ts
drewdrewthis: Fixed in 2db27ea — guard added, `!` removed.

```ts
const event = timeline[i];
if (!event) continue;
```

Also scrubbed the third assertion in this file (`chunks[0]!` in
`mergeChunks`) and the `resolveFn!` definite-assignment assertion, so
`adapter.runtime.ts` now has zero `!` of any kind. Verified with
`eslint --rule '{"@typescript-eslint/no-non-null-assertion":"error"}'`
(0 findings) — the rule itself is left off; enabling it repo-wide is a
separate PR.

0xdeafcafe (approved):

PR #750, opened 2026-07-09
refactor(realtime): use the injectable Logger instead of console.* (#724)

### LINKED ISSUE #724: Replace console logging with the injectable Logger in realtime-event-handler.ts and realtime-agent.adapter.ts
## Description

Replace direct `console.*` calls with the injectable `Logger` in `realtime-event-handler.ts` and `realtime-agent.adapter.ts`. Listed as a still-open follow-up on **PR #669**; never filed as its own issue.

## Source

> console.*->Logger in realtime-event-handler.ts/realtime-agent.adapter.ts

(PR #669 review verdict, "Follow-ups".)

---
*Classification: Tech-debt*
*Filed from a PR follow-up sweep (source: PR #669). Stub — not-ready until triaged.*

langwatch-agent:
## Ticket

Closes #724. The two realtime files still used raw `console.*` calls instead of the SDK's shared, level-aware `Logger` (a leftover follow-up from PR #669).

## Change

Route the four `console.*` calls through the shared `Logger` (`new Logger("<ClassName>")`), matching the convention already used by `user-simulator-agent`, `judge-agent`, and `event-bus`:

- `realtime-agent.adapter.ts`: two diagnostic `console.log` -> `logger.debug` (silent by default, gated behind `LOG_LEVEL` like the rest of the SDK, so the framework no longer prints unconditionally on every realtime call).
- `realtime-event-handler.ts`: two `console.error` -> `logger.error`.

Dropped the leftover emoji prefixes; the `Logger` already tags each line with its class context. No behavior change beyond routing through the shared logger.

## Evidence

- **No `console.*` remains** in either file (`grep` for `console.` returns nothing).
- **Behavior preserved:** `vitest run` on the realtime unit suites (`realtime-adapter-frames`, `realtime-response-formatter`, `connected-state`) -> `Test Files 3 passed (3)`, `Tests 9 passed (9)`.
- **Type-clean:** `pnpm typecheck` (tsc --noEmit) exits 0.
- **Scoped:** only the two files named in the ticket; the `Logger` import and one field per class plus the four call-site rewrites.

## Advisory note

### DIFF
--- a/javascript/src/agents/realtime/realtime-agent.adapter.ts
+++ b/javascript/src/agents/realtime/realtime-agent.adapter.ts
@@ -19,6 +19,7 @@ import {
 import { ResponseFormatter } from "./response-formatter";
 import type { AgentInput, AgentReturnTypes, AgentRole } from "../../domain";
 import { AgentAdapter } from "../../domain/agents";
+import { Logger } from "../../utils/logger";
 
 /**
  * Configuration for RealtimeAgentAdapter
@@ -96,6 +97,7 @@ export class RealtimeAgentAdapter extends AgentAdapter {
   private messageProcessor = new MessageProcessor();
   private responseFormatter = new ResponseFormatter();
   private audioEvents = new EventEmitter();
+  private readonly logger = new Logger("RealtimeAgentAdapter");
 
   /**
    * Creates a new RealtimeAgentAdapter instance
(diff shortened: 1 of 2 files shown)


langwatch-agent (commented):
No findings. The realtime changes are localized console.* replacements, match the shared Logger API, preserve transport error propagation, and move diagnostics behind the project logging level consistently with existing Logger usage. I did not find functional, security, or regression concerns.

Issue #565, opened 2026-05-27
refactor(typescript-sdk): enable + clean root-package eslint debt (~269 problems; CI does not lint root)


drewdrewthis:
## For Humans

**The TypeScript lint gate has holes, but not the ones this issue was filed about — and the shipped library is already clean.** PR #755 (merged 2026-07-09) added a `lint:lib` CI step that gates `javascript/src/**` excluding tests. That surface is now **0 problems**. The remaining ungated debt is **230 problems**, and it lives in test files (209), one example package that CI silently skips (19), and two root-level files (2).

**Ask:** ratify the revised acceptance criteria below (two of the original five are now wrong) before implementation starts.
- The "~269 problems" figure re-measures to **253** today, and **0** of them are in the shipped library.
- The "18 `import/no-unresolved`" are **not** root-package debt and **not** a resolver bug — they are a nested React example that lints green under its own config.
- New: the lint gate **silently depends on `pnpm build` having run**, and `pnpm -r run lint` **silently skips** any package without a `lint` script.

_Status: `refactor`, `P3 - low`, `tech-debt`, `investigating`. Awaiting AC ratification. Full spec below._

<details><summary>For Agents — full spec</summary>

## Investigation

**Root finding: the issue's premise was correct when filed and is now half-obsolete. The shipped library is already gated and already clean; the real remaining debt is test files, one silently-skipped example package, and two structural gaps in how the gate is composed.**

### Measurement environment

All numbers below were measured this turn, not inherited:

| | |
|---|---|
| Repo state | `langwatch/scenario` @ `88aec40` (main, 2026-07-30) |
| Command | `pnpm install --frozen-lockfile` → `pnpm build` → `eslint . --format json`, run in `javascript/` |
| ESLint | 9.39.4 (with nested-config lookup active — it picks up `examples/openai-realtime-demo/realtime-client/eslint.config.js`) |
| Node / pnpm | v24.13.0 / 10.22.0 |

### Headline numbers (post-build)

`pnpm lint` (= `eslint .` from `javascript/`) → **253 problems (241 errors, 12 warnings)**. 106 are autofixable; **145 remain after `eslint . --fix`**.

### Coverage map — what CI actually gates today

| Scope | Problems | Gated in CI today | After `--fix` |
|---|---:|---|---:|
| `src/**` non-test — the shipped library | **0** | ✅ `lint:lib`, added by #755 | 0 |
[...]

PR #755, opened 2026-07-09
chore(lint): forbid non-null assertions in library src and gate it in CI (#751)

### LINKED ISSUE #751: eslint: no TS rules applied — enable @typescript-eslint/no-non-null-assertion (13 `!`s slipped through #748)
## Summary

The scenario `javascript/eslint.config.mjs` applies **no TypeScript-specific lint rules** — `typescript-eslint` is imported but never wired into the config — so non-null assertions (`!`) are never flagged and pass CI.

## Investigation

- **Root cause:** `eslint.config.mjs` imports `typescript-eslint` (`import tseslint from "typescript-eslint"`) but never spreads its config or registers any `@typescript-eslint/*` rule, so there is **zero TS-specific linting**.
- **Evidence:** PR #748 introduced 13 non-null assertions across the voice adapter and its test; all passed `pnpm lint` in CI because no rule flags `!`.
- CI **does** run lint (`javascript-ci.yml` → `pnpm lint:all`) — the gap is the missing rule, not missing lint execution.

## Plan

- Wire the imported `typescript-eslint` into the flat config's rules block.
- Enable **`@typescript-eslint/no-non-null-assertion: "error"`**.
- Consider adopting the broader `@typescript-eslint` recommended set as a follow-up; scope this issue to `no-non-null-assertion`.
- Own PR (separate from #748), through the normal review cycle.

## Acceptance Criteria

- `eslint.config.mjs` applies `@typescript-eslint` with `no-non-null-assertion: error`.
- `pnpm lint` fails on a `!` non-null assertion (demonstrate before/after).
- Existing code passes — fix, or justify any remaining `!` with an inline disable + reason.

---
*Classification: Feature*
*Note: filed via raw `gh issue create` out of procedure 2026-07-09, then reconciled to the `create-issue` structure + labels.*

langwatch-agent:
## Ticket

Closes #751. Progresses #565.

The root `@langwatch/scenario` package is not linted in CI (`lint:all` runs `pnpm -r` over the example packages only, not the root), and `@typescript-eslint/no-non-null-assertion` was not enabled, so non-null assertions (`!`) in library code passed review unchecked. This enables the rule for the shipped library, fixes the existing violations, and gates the library in CI.

Note: the issue's original root-cause line ("typescript-eslint imported but never wired") is stale on current `main` (the config already spreads `tseslint.configs.recommended`); the real gaps were the missing `no-non-null-assertion` rule and the fact that CI never lints the root package.

## Change

- **Enable the rule (scoped):** `@typescript-eslint/no-non-null-assertion: "error"` for non-test `src/**`. Tests and examples legitimately assert known-present fixtures, so they are excluded; the broader test/example cleanup stays with #565. Scoping to `src/` also means the existing example lint (`lint:all`) is unaffected.
- **Fix all 59 library violations (no behavior change):**
  - Redundant `!` on array / typed-array indexing dropped. `noUncheckedIndexedAccess` is off, so `arr[i]` is already the non-undefined element type and the assertion was a no-op.
  - `Map.get()`, optional-field, and env cases replaced with real guards / local captures that narrow the type and throw a clear error if the connection invariant is ever violated (judge span helpers, vitest reporter, scenario-execution, and the pipecat / twilio / playback adapters).
  - One justified `prefer-const` inline-disable where a `cleanup()` closure captures the timer handle before it is assigned (declaration and assignment cannot be merged).
- **Bring the library fully clean under the config:** auto-fix import ordering and type one stray `any` (`as ModelMessage[]`) so `src/` passes the whole config, not just the new rule.
- **Gate it in CI:** add a `lint:lib` script (`eslint 'src/**/*.ts'` excluding tests) and a "Lint (library)" step in `javascript-ci.yml`.

## Evidence

- **Rule enforced + fails on new `!` (AC2, before/after):** injecting a `!` into a library file makes `pnpm lint:lib` exit 1 flagging `no-non-null-assertion`; reverting returns it to exit 0.
- **Existing library passes (AC3):** `pnpm lint:lib` -> exit 0 (0 errors).
- **Existing CI unaffected:** `pnpm lint:all` -> exit 0 (rule scoped to `src/`, examples untouched).
- **No behavior change:** `vitest run src/` -> `Test Files 79 passed | 1 skipped`, `Tests 906 passed | 1 skipped`; `pnpm typecheck` exits 0.
- No lockfile churn; `actionlint` clean on the workflow change.

## Advisory note

## Summary by CodeRabbit

* **New Features**
  * Added a new library lint check to CI for JavaScript code.

* **Bug Fixes**
  * Improved runtime safety by adding clearer checks for missing connections, IDs, and stream state in voice and realtime flows.
  * Made several data-processing paths more defensive when handling missing values.
  * Reduced reliance on unsafe assumptions in agent, scenario, and reporting logic.

* **Chores**
  * Tightened linting rules and cleaned up formatting to support safer, more consistent code.

### DIFF
--- a/.github/workflows/javascript-ci.yml
+++ b/.github/workflows/javascript-ci.yml
@@ -72,6 +72,14 @@ jobs:
         run: pnpm lint:all
         working-directory: javascript
 
+      # Lints the shipped library source (src/, excluding tests) with the root
+      # package config, which the workspace-recursive lint:all does not cover.
+      # Enforces no-non-null-assertion (#751) and the rest of the config on the
+      # library; extending the gate to tests/examples is tracked in #565.
+      - name: Lint (library)
+        run: pnpm lint:lib
+        working-directory: javascript
+
       - name: Type check
         run: pnpm typecheck:all
         working-directory: javascript
--- a/javascript/eslint.config.mjs
+++ b/javascript/eslint.config.mjs
@@ -75,4 +75,15 @@ export default defineConfig([
       ],
     },
   },
+  {
+    // Forbid non-null assertions (`!`) in shipped library source. Tests and
+    // examples legitimately assert known-present fixtures, so the rule is
+    // scoped to non-test `src/` only; extending the lint gate to the rest of
+    // the package is tracked in #565.
+    files: ["src/**/*.ts"],
+    ignores: ["src/**/*.test.ts", "src/**/__tests__/**"],
+    rules: {
+      "@typescript-eslint/no-non-null-assertion": "error",
+    },
+  },
 ]);
--- a/javascript/package.json
+++ b/javascript/package.json
@@ -18,6 +18,7 @@
     "watch": "pnpm run build -- --watch",
     "typecheck": "tsc --noEmit",
     "lint": "eslint .",
+    "lint:lib": "eslint 'src/**/*.ts' --ignore-pattern '**/*.test.ts' --ignore-pattern '**/__tests__/**'",
     "format": "eslint . --fix",
     "test": "vitest run",
     "test:watch": "vitest",
--- a/javascript/src/agents/judge/judge-agent.ts
+++ b/javascript/src/agents/judge/judge-agent.ts
@@ -125,16 +125,19 @@ function collapseDiscoveryHistory(
   return out;
 }
 
-import { JudgeUtils } from "./judge-utils";
 import { estimateTokens, DEFAULT_TOKEN_THRESHOLD } from "./estimate-tokens";
+import { JudgeResult } from "./interfaces";
+import { judgeSpanCollector, JudgeSpanCollector } from "./judge-span-collector";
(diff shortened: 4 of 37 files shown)


langwatch-agent (commented):
No findings.

I focused on the pipecat/twilio/playback connection guard rewrites and the lint/config scoping. The adapter changes preserve the previous lifecycle assertions and then capture narrowed locals; I did not find a behavior change in the connected/stream-live paths. The new lint rule is scoped to shipped `src/**/*.ts` while excluding tests, and CI adds the root library lint gate without changing the existing example lint behavior.

Local validation on the PR head:
- `pnpm lint:lib` passed
- `pnpm typecheck` passed
- `pnpm vitest run src/` passed: 79 files passed, 906 tests passed, 1 skipped

Residual note: GitHub checks were still queued/running when I reviewed, so wait for CI before merge.

0xdeafcafe (approved):

PR #586, opened 2026-05-29
chore: main-side cleanup — docs + spec + python/TS parity

drewdrewthis:
## Summary

Consolidated `main`-side cleanup PR: docs housekeeping + spec housekeeping + python/TS parity fixes. Surfaced by a directory-structure audit on PR [#561](https://github.com/langwatch/scenario/pull/561). Lands the changes for the follow-up issues listed under **Closes** below (most already closed; see that section).

## File-structure changes

What moved, what was deleted, what was added (the recordings tree moved wholesale — 68 files — shown collapsed):

```
docs/
  voice-bug-bash.md                          ✖ deleted (orphan QA guide, zero refs)
  voice-twilio.md                            ✖ deleted (superseded by voice/adapters/twilio.mdx)
  voice/
    capability-matrix.md                     ✖ deleted ─┐ folded into the published mdx
    happy-path-elevenlabs.md                 ──► docs/docs/pages/voice/happy-path-elevenlabs.mdx   (published)
    happy-path-openai-realtime.md            ──► docs/docs/pages/voice/happy-path-openai-realtime.mdx (published)
  LOW_RISK_PULL_REQUESTS.md                  ──► .github/LOW_RISK_PULL_REQUESTS.md   (GitHub-process doc)
  docs/pages/voice/capability-matrix.mdx     ✎ edited (now the single source-of-truth; sidebar in docs/vocs.config.tsx)

javascript/
  docs/voice/capability-matrix.md            ✖ deleted ─┘ (was a divergent copy)

python/
  recordings/                                ──► python/outputs/recordings/   (whole tree: 18 demos, 68 files)
  outputs/README.md                          ✚ added (documents the new outputs/ parent)
  scenario/voice/assets/noise/*.wav          ✎ refreshed (cafe/street/office/airport + babble)
  scenario/voice/capabilities.py             ✎ edited (error msg → published capability-matrix URL)

javascript/src/voice/capabilities.ts         ✎ edited (error msg → published capability-matrix URL)
.github/workflows/pr-auto-approve.yml        ✎ edited (HTTP-406 oversized-diff handling; new LOW_RISK path)
javascript/tsconfig.json                     ✎ edited (removed duplicate "target" key)
specs/langwatch-pr-gate-pattern.feature      ✎ edited (dropped a stale, no-longer-meaningful scenario)
```

### Docs housekeeping
[...]

PR #799, opened 2026-07-16
chore: repoint the python recordings gitignore policy at python/outputs/ (#624)

### LINKED ISSUE #624: refactor: .gitignore recordings policy orphaned at old python/recordings/* path — outputs/recordings/ uncovered, dirties tree on every e2e run
# refactor: `.gitignore` recordings policy is orphaned at the old `python/recordings/*` path — voice-demo artifacts under `python/outputs/recordings/` are uncovered and dirty the tree on every e2e run

**Repo:** langwatch/scenario
**Type:** refactor / repo-hygiene
**Labels (suggest):** chore, repo-hygiene

## Background

`python/outputs/` is the intentional, designed home for voice-demo/test-run artifacts (it has its own `python/outputs/README.md` explaining the parent-dir rationale; `python/outputs/recordings/` holds the "prove it" PR-evidence recordings documented in `python/outputs/recordings/README.md`). The recordings that are committed there are deliberate evidence, not an accidental commit.

The `.gitignore` block that governs these recordings was authored for the demos' **original** location, `python/recordings/`. Its job: ignore the recordings that get regenerated on every local run, while keeping only a hand-picked committed-evidence subset (each demo's `full.wav` + `manifest.json`, plus `segments/` for the seven core-provider demos).

PR #586 (`371f94cd` — "chore: main-side cleanup — docs + spec + python/TS parity") moved the demo helper, both READMEs, and all the tracked recordings from `python/recordings/` to `python/outputs/recordings/` (the new path is hard-coded in `python/examples/voice/_recording_helper.py:13`). **The `.gitignore` rules were not updated in that move.** They still point at `python/recordings/*`, which now has zero tracked files — the rule is a dead no-op, and the new `python/outputs/recordings/` path is matched by no ignore rule at all.

## Current-state violations (evidence, gathered on branch `feat/621-agent-first-and-echo-safe-transcript`, today)

1. **The ignore policy points at an empty path.**
   ```
   $ git ls-files python/recordings/ | wc -l
   0
   $ git check-ignore -v python/recordings/foo/full.wav
   .gitignore:202:python/recordings/*	python/recordings/foo/full.wav      # matches the OLD, now-empty path
   ```

2. **The real artifact path is covered by nothing.**
   ```
   $ git check-ignore -v python/outputs/recordings/basic_greeting/full.wav \
                          python/outputs/recordings/basic_greeting/segments/
   (no output; exit 1)   # NOT ignored
   ```

3. **Committed-evidence recordings get rewritten on every e2e run, dirtying the tree.** A normal e2e/demo run rewrites `full.wav` + `manifest.json` for the demos it touches (new `generated_at` timestamp, new durations, new per-segment filenames), and drops untracked `segments/` dirs:
   ```
   $ git status --short
    M python/outputs/recordings/angry_customer/full.wav
    M python/outputs/recordings/angry_customer/manifest.json
    M python/outputs/recordings/background_handoff/full.wav
    M python/outputs/recordings/background_handoff/manifest.json
    M python/outputs/recordings/basic_greeting/full.wav
    M python/outputs/recordings/basic_greeting/manifest.json
   ?? python/outputs/recordings/angry_customer/segments/
   ?? python/outputs/recordings/background_handoff/segments/
   ?? python/outputs/recordings/basic_greeting/segments/

   $ git diff --stat python/outputs/recordings/*/full.wav
    angry_customer/full.wav     | Bin 852092 -> 1392800 bytes
    background_handoff/full.wav | Bin 616956 -> 617408 bytes
    basic_greeting/full.wav     | Bin 681192 -> 664452 bytes
   ```
   The untracked `segments/` dirs are exactly what the old gitignore's `…/segments/` lines were meant to suppress for the non-core demos — but those lines, too, point at the dead `python/recordings/` path.

4. **The exception allowlist is stale relative to the actual committed inventory** — so this is NOT a blind path-swap. The old block's `!`-exceptions name `elevenlabs_*`, `twilio_*`, `gemini_live*`, `openai_realtime_agent`, `pipecat_*`, `basic_greeting`, `voice_text_parity`, `recording_playback`. But the tree at `python/outputs/recordings/` actually tracks several demos that were never in that allowlist (`angry_customer`, `background_handoff`, `interruption_recovery`, `random_interruptions`, `stt_swap`), and the segments-exclusion lines don't match the committed reality either (e.g. `gemini_live/` has 4 committed `segments/` files; the old block lists `gemini_live` only as a top-level un-ignore with no segments rule). Tracked inventory today:
   ```
   angry_customer:       2 files (segments: 0)
   background_handoff:   2 files (segments: 0)
   basic_greeting:       2 files (segments: 0)
   elevenlabs_branded:   6 files (segments: 4)
   elevenlabs_hosted:    7 files (segments: 5)
   elevenlabs_interruption: 8 files (segments: 6)
   gemini_live:          6 files (segments: 4)
   gemini_live_interruption: 8 files (segments: 6)
   interruption_recovery: 2 files (segments: 0)
   openai_realtime_agent: 2 files (segments: 0)
   pipecat_scenario:     2 files (segments: 0)
   pipecat_ws:           2 files (segments: 0)
   random_interruptions: 2 files (segments: 0)
   recording_playback:   2 files (segments: 0)
   stt_swap:             2 files (segments: 0)
   twilio_inbound:       6 files (segments: 4)
   twilio_outbound:      4 files (segments: 2)
   voice_text_parity:    2 files (segments: 0)
   ```

5. **The comment + README pointer reference the dead path too.** The `.gitignore` comment (line ~201) says "See python/recordings/README.md" — that file no longer exists at that path (it's now `python/outputs/recordings/README.md`). Dead cross-reference.

## Why now

Every e2e/voice-demo run leaves a dirty working tree (modified committed `full.wav`/`manifest.json` + untracked `segments/`), which:
- masks real changes in `git status`,
- invites accidental `git add .` of multi-hundred-KB regenerated binaries (the `full.wav` deltas above are ~0.5–1.4 MB each), and
- defeats the original, deliberate design intent of the ignore block (keep only the curated PR-evidence subset tracked; ignore everything regenerated).

The fix is to **port the recordings ignore policy from the dead `python/recordings/` path to the live `python/outputs/recordings/` path**, reconciled against the *actual* committed inventory above (not a blind string-replace), and update the stale comment/README pointer. This restores the intended behavior — a clean tree after a demo run, with only the curated evidence subset tracked.

## Scope notes (read before defining ACs)

- The committed-evidence recordings are **intentional and must stay tracked** — this is a `.gitignore`-correction + selective-untrack problem, NOT a blanket `git rm -r` of `outputs/recordings/`. Investigation must decide, per demo, which files stay tracked (curated evidence) vs. which the new ignore rules should exclude, using the inventory in violation 4 as the source of truth.
- A secondary open question for investigation: should committed binary evidence (`full.wav`) live in-tree at all, or move to Git LFS / a release asset? Out of scope for the minimal hygiene fix, but worth a one-line note + possible follow-up issue.
- The dead `python/recordings/` directory still exists on disk locally (untracked) but has 0 tracked files; investigation should confirm whether any tooling still writes there.

---

*Refactor stub — no acceptance criteria yet. ACs come from the `/create-issue` investigation step, reconciled against the live tracked inventory and the two READMEs (`python/outputs/README.md`, `python/outputs/recordings/README.md`).*

langwatch-agent:
## Ticket

Closes #624. The recordings `.gitignore` block still governs `python/recordings/`, the path #586 emptied when it moved the demos to `python/outputs/recordings/`. **Nothing is tracked under the old path any more (0 files), so all 38 lines of the block match nothing**, and the real artifact directory (68 tracked files) is governed by no rule at all.

## Change

One file, `.gitignore`. Repoint the block and adopt the same `outputs/` parent shape the **TypeScript block directly below it already uses**, so `python/outputs/` is treated like `javascript/examples/vitest/outputs/` and future artifact siblings (`traces/`, `logs/`) are ignored by default rather than by omission. Also fixes the two dead cross-references to `python/recordings/README.md`.

**One correction to the ticket, worth a reviewer's eye.** It warns that "the exception allowlist is stale relative to the actual committed inventory, so this is NOT a blind path-swap". That is no longer true: the block has been kept current since filing. I derived the allowlist from the tracked tree rather than trusting either list, and they match exactly:

| | Demos |
| --- | --- |
| Track `segments/` (7) | `elevenlabs_hosted`, `elevenlabs_branded`, `elevenlabs_interruption`, `gemini_live`, `gemini_live_interruption`, `twilio_inbound`, `twilio_outbound` |
| Track only `full.wav` + `manifest.json` (11) | `openai_realtime_agent`, `pipecat_scenario`, `pipecat_ws`, `basic_greeting`, `voice_text_parity`, `recording_playback`, `interruption_recovery`, `random_interruptions`, `angry_customer`, `background_handoff`, `stt_swap` |

Those are precisely the existing core / additional exception groups. So the allowlist is carried over unchanged and only the path moved, which is what makes this reviewable.

## Evidence

Simulated a demo run dropping artifacts, then diffed `git status` between the two `.gitignore` versions:

**On `origin/main` today, the tree is dirtied:**
```
?? python/outputs/recordings/angry_customer/segments/
?? python/outputs/recordings/background_handoff/segments/
?? python/outputs/recordings/basic_greeting/segments/
?? python/outputs/recordings/brand_new_demo/
?? python/outputs/recordings/stt_swap/segments/
?? python/outputs/traces/
```

**With this change, identical simulation, clean:**
```
(nothing)
```

Three invariants checked mechanically:

| Invariant | Result |
| --- | --- |
| No **tracked** file becomes ignored (would break a later `git add`) | all **69** tracked files under `python/outputs/` remain un-ignored |
| Regenerated / new artifacts **are** ignored | additional-demo `segments/`, a brand-new demo's `full.wav`, and `python/outputs/traces/` all ignored |
| Core-demo `segments/` stay **committable** | `elevenlabs_*`, `gemini_live`, `twilio_*` segments plus both READMEs un-ignored |

**Scope caveat, stated plainly:** the ticket also lists the tracked `full.wav` / `manifest.json` showing as ` M` after a run. `.gitignore` cannot fix that, since it only affects untracked files; a tracked file always reports modifications. This PR removes the `??` noise and the dead path. The ` M` churn is inherent to committing regenerated evidence and needs a separate decision (leave it, or stop tracking the volatile ones).

## Advisory note

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -195,47 +195,56 @@ docs/docs/public/reference/python/**/*
 .claude/settings.local.json
 .worktrees
 
-# Voice demo recordings — most are regenerated per local run, NOT
-# committed. But the subset listed below is checked in as PR
-# evidence: each demo's `full.wav`, `manifest.json`, and per-turn
-# `segments/` directory are the canonical "prove it" artifacts for
-# reviewers. See python/recordings/README.md.
-python/recordings/*
-!python/recordings/README.md
-!python/recordings/elevenlabs_hosted/
-!python/recordings/elevenlabs_branded/
-!python/recordings/elevenlabs_interruption/
-!python/recordings/twilio_inbound/
-!python/recordings/twilio_outbound/
-!python/recordings/gemini_live/
-!python/recordings/gemini_live_interruption/
+# Python voice demo outputs (recordings + manifests). Most are regenerated
+# per local run, NOT committed. But the subset listed below is checked in as
+# PR evidence: each demo's `full.wav`, `manifest.json`, and for the core
+# provider demos the per-turn `segments/` directory are the canonical
+# "prove it" artifacts for reviewers. See python/outputs/README.md (parent
+# index) and python/outputs/recordings/README.md (audio policy).
+#
+# Shape: outputs/ is the artifact parent (room for traces/, logs/,
+# screenshots/ later). recordings/ is the audio subdir, and only it is
+# whitelisted today; new artifact siblings will get their own block when added.
+python/outputs/*
+!python/outputs/README.md
+!python/outputs/recordings/
+python/outputs/recordings/*
+!python/outputs/recordings/README.md
+# Core provider demos: commit full.wav + manifest.json + per-turn segments/.
+!python/outputs/recordings/elevenlabs_hosted/
+!python/outputs/recordings/elevenlabs_branded/
+!python/outputs/recordings/elevenlabs_interruption/
+!python/outputs/recordings/twilio_inbound/
+!python/outputs/recordings/twilio_outbound/
+!python/outputs/recordings/gemini_live/
+!python/outputs/recordings/gemini_live_interruption/
 # Additional demos — commit only full.wav + manifest.json (skip segments/
 # to avoid hundreds of tiny per-turn wavs in the tree).
-!python/recordings/openai_realtime_agent/
-python/recordings/openai_realtime_agent/segments/
-!python/recordings/pipecat_scenario/
-python/recordings/pipecat_scenario/segments/
-!python/recordings/pipecat_ws/
-python/recordings/pipecat_ws/segments/
-!python/recordings/basic_greeting/
-python/recordings/basic_greeting/segments/
-!python/recordings/voice_text_parity/
-python/recordings/voice_text_parity/segments/
-!python/recordings/recording_playback/
-python/recordings/recording_playback/segments/
-!python/recordings/interruption_recovery/
-python/recordings/interruption_recovery/segments/
-!python/recordings/random_interruptions/
-python/recordings/random_interruptions/segments/
-!python/recordings/angry_customer/
-python/recordings/angry_customer/segments/
-!python/recordings/background_handoff/
-python/recordings/background_handoff/segments/
-!python/recordings/stt_swap/
-python/recordings/stt_swap/segments/
+!python/outputs/recordings/openai_realtime_agent/
+python/outputs/recordings/openai_realtime_agent/segments/
+!python/outputs/recordings/pipecat_scenario/
+python/outputs/recordings/pipecat_scenario/segments/
+!python/outputs/recordings/pipecat_ws/
+python/outputs/recordings/pipecat_ws/segments/
+!python/outputs/recordings/basic_greeting/
+python/outputs/recordings/basic_greeting/segments/
+!python/outputs/recordings/voice_text_parity/
+python/outputs/recordings/voice_text_parity/segments/
+!python/outputs/recordings/recording_playback/
+python/outputs/recordings/recording_playback/segments/
+!python/outputs/recordings/interruption_recovery/
+python/outputs/recordings/interruption_recovery/segments/
+!python/outputs/recordings/random_interruptions/
+python/outputs/recordings/random_interruptions/segments/
+!python/outputs/recordings/angry_customer/
+python/outputs/recordings/angry_customer/segments/
+!python/outputs/recordings/background_handoff/
+python/outputs/recordings/background_handoff/segments/
+!python/outputs/recordings/stt_swap/
+python/outputs/recordings/stt_swap/segments/
 
 # TypeScript voice demo outputs (recordings + manifests) — same policy as
-# python/recordings/ above. Regenerated per local run by the @e2e demo tests
+# python/outputs/recordings/ above. Regenerated per local run by the @e2e demo tests
 # under javascript/examples/vitest/tests/voice/; the committed subset below
 # is the canonical "prove it" audio evidence. See
 # javascript/examples/vitest/outputs/README.md (parent index) and


langwatch-agent (commented):
No P0-P2 findings.

Validated this against Git's tracked index and ignore matcher at `b1691b5b34d864cba1402c949d43cbca85c4a7d1`:

- `python/recordings/` has 0 tracked files; `python/outputs/recordings/` has 68, plus the `python/outputs/README.md` parent file (69 files under `python/outputs/` total).
- The carried-forward allowlist is exact: the 7 core demos are the only demos with tracked `segments/`; all 11 additional demos have only their tracked non-segment artifacts.
- No tracked output file matches an ignore rule.
- Virtual-path checks confirm new demo artifacts, additional-demo `segments/`, and unrelated `python/outputs/` artifacts are ignored, while core-demo `segments/` and both README files remain committable.

The noted modifications to already tracked `full.wav` and `manifest.json` files are correctly outside `.gitignore`'s scope.

PR #610, opened 2026-06-04
docs(voice/#606): document STT/TTS model choices as deliberate current-gen

drewdrewthis:
## Problem
The voice SDK's default STT (\`gpt-4o-transcribe\`) and TTS (\`gpt-4o-mini-tts\`) constants carried minimal doc comments, leaving it unclear whether the \`gpt-4o-*\` naming was intentional or an oversight. Separately, the \`multimodal-audio-to-text\` example had three judge criteria that tested incidental phrasing rather than audio-processing capability, causing false failures against \`gpt-audio-mini\` (which was swapped in for the deleted \`gpt-4o-audio-preview\` in PR #607).

## Changes

**\`javascript/src/voice/voice-models.ts\`** — expanded \`OPENAI_STT_MODEL\` and \`OPENAI_TTS_MODEL\` doc comments to explain the deliberate current-gen choice and name the revisit trigger. Added trailing comment documenting the Python-only \`OPENAI_BOT_STT_MODEL\` constant.

**\`python/scenario/config/voice_models.py\`** — equivalent "Deliberate:" rationale appended to both constants' existing comment blocks.

**\`javascript/examples/vitest/tests/multimodal-audio-to-text.test.ts\`** — three brittle judge criteria replaced with capability-focused, model-agnostic equivalents. Skip comment updated to reflect current state (model fixed in PR #607; criteria relaxed here).

**\`python/tests/voice/test_feature_file_contract.py\`** — fixed pre-existing CI breakage inherited from \`main\`: updated hardcoded scenario count (108→127) and tag split (75/8/25→79/13/35). Root cause: PR #561 added voice scenarios without updating the contract test; \`main\` python-ci has been red since \`5847c4b4\`. Tracked in #609.

No model values changed. No skip markers removed (that's #486's scope).

## ACs met (from #606)
- **AC1** — STT/TTS doc comments in both TS and Python ✅
- **AC2** — \`OPENAI_BOT_STT_MODEL\` parity gap documented (Python-only, comment explains why) ✅
- **AC3** — \`multimodal-audio-to-text.test.ts\` judge criteria are model-agnostic capability checks ✅

## Bonus fix
- Pre-existing \`python-ci\` red on \`main\` (issue #609): contract test counts updated to match actual feature file ✅

Closes #606
Closes #609

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #595, opened 2026-06-01
chore: scrub bare §-section "source proposal" provenance refs (unresolvable internal-doc refs)


drewdrewthis:
## Background

PR #594 scrubbed ~130 bare acceptance-criteria tags (`AC-1.1`, `AC-14`, …) that pointed at an external GitHub-issue AC list a code reader cannot open. While doing that, a sibling pattern surfaced: bare `§x.y` / `Source §N L###` "source proposal" provenance annotations scattered across the code, specs, and docs.

The governing principle (from the PR-description policy added alongside #594, and stated directly): **referencing valid links is fine; referencing internal content that might change is not.** A bare `§N.M L###` pointer at an out-of-repo proposal document is an unresolvable reference to mutable internal content — and the embedded line numbers rot the moment that document changes.

## Current-state violations

- **160 `§`-reference lines across 61 tracked files** (`python/examples/voice/*.py`, `python/scenario/*`, `specs/*.feature`, `docs/**/*.mdx`). Examples: `# Source §8 L1243-1257`, `Source §6.3, L931-967 and §8 emotional escalation`.
- **~159 of the 160 are bare/non-linkable** — only ~1 is an actual hyperlink to an openable in-repo target.
- **The "source proposal" document is not in the repository** — no proposal / design / RFC file exists (grep finds none; only the feature specs themselves). So these refs point at something a reader cannot reach, and which has no stable anchor.

## Why now

The hygiene rationale and the safe transform are already proven by PR #594 (drop the tag, keep the human-readable substance; comments/docstrings only; behavioral-logic and Gherkin-step/scenario-count lines untouched, guarded by the feature-file contract test). Doing the `§` sweep while that pattern is fresh — but as a **separate** change so it does not block #594 — closes the whole class of unresolvable-internal-reference smell in one follow-up.

## Scope notes (for investigation to refine)

- Remove bare, non-linkable `§x.y` / `Source §N L###` tags; preserve the description each annotates.
- **Keep** any `§`-ref already expressed as a working hyperlink to an in-repo file.
- No behavioral / logic / Gherkin-step / scenario-count changes — verify with the same guards as #594 (feature-file contract test stays green: 108 scenarios, tag split 75/8/25).
- ~61 files, likely one sweep PR. Sequence **after** #594 (overlapping files, no logical dependency).

---
*Classification: Refactor*
*Status: stub — filed deliberately as standalone follow-up to PR #594. Not auto-investigated per request; run /investigate when prioritized.*

[...]

Issue #350, opened 2026-04-15
Voice Agents: implement per research proposal


drewdrewthis:
> [!IMPORTANT]
> **Demo verification checklist (added 2026-04-28)** — local listen-test status. Each demo runs via `cd python && uv run python examples/voice_<name>.py`; output lands at `python/recordings/<name>/`.
>
> **Stack used per demo (creds required):**
> - **OpenAI-only path** (`OPENAI_API_KEY`): `example_6_1_basic_greeting`, `example_6_2_interruption_recovery`, `example_6_3_angry_customer`, `example_6_4_dtmf_ivr`, `example_6_5_tool_verification`, `example_6_6_prerecorded_audio`, `example_6_7_random_interruptions`, `example_6_8_silence_handling`, `demo_observability`, `demo_recording_playback`, `demo_stt_swap`, `demo_voice_text_parity`, `pain_accent_loop`, `pain_background_handoff`, `pain_emotional_escalation`, `pain_long_hold`, `pain_multi_intent`, `pipecat_scenario`
> - **OpenAI Realtime** (`OPENAI_API_KEY`): `demo_openai_realtime_agent`, `demo_openai_realtime_user`, `demo_pipecat_ws`
> - **ElevenLabs** (`ELEVENLABS_API_KEY` + `ELEVENLABS_AGENT_ID`): `demo_elevenlabs_hosted`, `demo_elevenlabs_branded`
> - **Gemini Live** (`GEMINI_API_KEY` with Live API access): `demo_gemini_live` — *blocked: current key has no `bidiGenerateContent` models*
> - **Twilio** (`TWILIO_*`): `demo_twilio_inbound`, `demo_twilio_outbound` — *blocked: `TWILIO_AUTH_TOKEN` returns 401, see ND-15*
>
> **Verified locally (Drew listened):**
> - [x] `example_6_1_basic_greeting` — works, recording good. Surfaced **ND-17 turn-taking overlap bug** (see below).
>
> **Pending Drew listen-test:** every other demo above. Cred-blocked demos remain blocked until ND-15 cred rotation lands.
>
[...]

PR #823, opened 2026-07-17
docs(voice): drop references to a docs/proposals tree that never landed (#613)

### LINKED ISSUE #613: refactor(voice): reconcile dangling docs/proposals/ references — the directory is absent from the repo
## Background

The voice/issue-350 work referenced a `docs/proposals/` directory that **never landed** — the entire `docs/proposals/` tree is absent from `origin/main` and from all `--all` history. Several source, test, and spec files still point at proposal docs that don't exist, leaving dangling "see this doc" / "regenerate this report" references for any reader who follows them.

Surfaced during a `/sweep` while fixing #609 (stale feature-file contract counts).

## Current-state violations

Three live references to absent `docs/proposals/` files:

| Location | References | Exists? |
|----------|-----------|---------|
| `python/scenario/voice/adapters/twilio.py:17` | `docs/proposals/issue-350-ralph-real-transports.md` | ❌ absent |
| `specs/voice-docs-surface.feature:146` | `docs/proposals/issue-350-voice-agents-source.md` (scenario asserts a page links to it) | ❌ absent |
| `python/tests/voice/test_feature_file_contract.py` | `docs/proposals/issue-350-prove-it-report.md` | ❌ absent — being reworded via #609's fix in #610/#612 |

Plus a related smell in the same test file: the function `test_tag_split_matches_prove_it_report` is **named** after the phantom prove-it report. Even once #610/#612 remove the in-body path references, the function name still points a `grep` at a report that never existed.

## Why now

The dangling references are actively accreting (this is the same `docs/proposals/` pattern hit from three different files), and a reader who follows any of them wastes time hunting for a doc that isn't there. The #609 fix cleans up only the prove-it-report path reference in one file — the other two references and the stale function name remain.

Not blocking CI — these are prose/comment/spec-text references, not failing assertions — which is why they belong in a dedicated cleanup rather than the #609 hotfix.

_Investigation pending — ACs and proof produced by /investigate._

---
*Classification: Refactor*
*Status: stub — investigation pending.*

langwatch-agent:
## Ticket

Closes #613. `docs/proposals/` is absent from `origin/main` **and from all history** (`git log --all -- 'docs/proposals/*'` is empty), so every pointer at it sends a reader hunting for a file that was never committed.

## Premise correction

The ticket lists **three** live path references. Only **two** survive on `main`: the third, in `test_feature_file_contract.py`, was already cleaned up by #610/#612 as the ticket anticipated. What remains there is the *wording* and the function name, which the ticket also flags.

The ticket is marked "investigation pending", so here is the current state, verified rather than assumed:

| Location | Reference | State |
| --- | --- | --- |
| `python/scenario/voice/adapters/twilio.py:17` | `docs/proposals/issue-350-ralph-real-transports.md` | live, removed here |
| `specs/voice-docs-surface.feature:146` | `docs/proposals/issue-350-voice-agents-source.md` | live, removed here |
| `python/tests/voice/test_feature_file_contract.py` | path ref | **already gone** (#610/#612) |
| same file | `prove-it report` wording + function name | live, fixed here |

## Change

Three files, net -3 lines.

**`twilio.py`** loses `See source §5.3 and docs/proposals/issue-350-ralph-real-transports.md.` Both halves are unresolvable (the `§5.3` is the sibling pattern tracked in #595), and the paragraphs above it already describe the connect/place_call/wait_for_call flow it pointed at, so nothing is lost.

**`voice-docs-surface.feature`** required the published docs to link the proposal "for design context". I checked: **nothing under `docs/` links it, and nothing can**, so the scenario asserted a requirement that was impossible to satisfy. It now covers only the cross-link that does exist (`specs/voice-agents.feature`), and the scenario title drops "the proposal" to match. Safe because the feature has **no pytest-bdd binding** (nothing in the repo references `voice-docs-surface`), so this changes no test.

**`test_feature_file_contract.py`** is the one worth a look. Its docstring claimed it catches "an AC deleted without updating the prove-it report", and `test_tag_split_matches_prove_it_report` was named after it. **The report never existed, and the test never read one**: it asserts the tag split against counts pinned in its own body. The name pointed a `grep` at a phantom and described behaviour the code does not implement. Both now say what the module actually does (`test_tag_split_matches_the_pinned_counts`).

## Evidence

| Check | Result |
| --- | --- |
| No dangling refs remain | `git grep 'docs/proposals\|issue-350-\|prove-it report'` returns nothing |
| Rename is safe | old function name referenced nowhere in the repo |
| Contract test still runs | **5 passed**, including the renamed `test_tag_split_matches_the_pinned_counts` |
| Twilio + contract suites | **76 passed, 2 skipped** |
| Adapter still imports | `import twilio` OK, docstring intact at 14 lines, no `proposals` mention |
| Spec still valid Gherkin | parses OK, 18 scenarios, edited one renamed as intended |

**Not verified:** the full `pytest tests/voice/` sweep times out locally on tests that reach the network. I ran every suite that touches the changed files instead; CI is the check that matters.

## Advisory note

### DIFF
--- a/specs/voice-docs-surface.feature
+++ b/specs/voice-docs-surface.feature
@@ -140,11 +140,10 @@ Feature: User-facing docs surface for the voice-agent adapter system
     And the legacy Multimodal -> Voice Agents group is removed from the sidebar config
 
   @integration
-  Scenario: The voice docs cross-link to the proposal and the behavioral contract
+  Scenario: The voice docs cross-link to the behavioral contract
     Given the Voice Agents docs section is published
     When a reader opens Getting Started or Choosing an Adapter
-    Then the page links to docs/proposals/issue-350-voice-agents-source.md for design context
-    And the page links to specs/voice-agents.feature for the 99-scenario behavioral contract
+    Then the page links to specs/voice-agents.feature for the 99-scenario behavioral contract
 
   # ============================================================
   # Group: Legacy gpt-4o-audio-preview voice surface retirement
--- a/python/scenario/voice/adapters/twilio.py
+++ b/python/scenario/voice/adapters/twilio.py
@@ -13,8 +13,6 @@ One adapter class serves both directions a Twilio number can participate in:
 FastAPI webhook + Media Streams WS server, open a public URL (user-supplied
 or via ``CloudflareTunnel``). After ``connect()``, call either
 ``place_call()`` or ``wait_for_call()``.
-
-See source §5.3 and docs/proposals/issue-350-ralph-real-transports.md.
 """
 
 from __future__ import annotations
--- a/python/tests/voice/test_feature_file_contract.py
+++ b/python/tests/voice/test_feature_file_contract.py
@@ -3,8 +3,8 @@ Structural guard for specs/voice-agents.feature — the behavioral contract.
 
 Parses the feature file with the same Gherkin parser pytest-bdd uses and
 asserts it is well-formed. Catches regressions where a scenario gets
-dropped, a tag gets misspelled, or an AC is deleted without updating the
-prove-it report.
+dropped, a tag gets misspelled, or the tag split drifts from the counts
+this module pins.
 
 This is the minimum-viable half of "BDD wiring." Full pytest-bdd binding
 (one executable test per Scenario) is tracked as follow-up; local
@@ -102,7 +102,7 @@ def test_every_scenario_is_tagged_unit_integration_or_e2e(parsed_feature):
     )
 
 
-def test_tag_split_matches_prove_it_report(parsed_feature):
+def test_tag_split_matches_the_pinned_counts(parsed_feature):
     """
     The post-#561/#604 tag split is 79 @unit / 13 @integration / 35 @e2e.
     The end-to-end examples and pain-pattern scenarios are @e2e (happy paths


specs/voice-docs-surface.feature
```diff
@@ -140,11 +140,10 @@ Feature: User-facing docs surface for the voice-agent adapter system
     And the legacy Multimodal -> Voice Agents group is removed from the sidebar config
 
   @integration
-  Scenario: The voice docs cross-link to the proposal and the behavioral contract
+  Scenario: The voice docs cross-link to the behavioral contract
```
langwatch-agent, line 143: **P2 — Do not delete the design-context acceptance criterion.**

The AC coverage map below still says AC 14 requires links to both the proposal and the feature file, while this change removes the proposal half from its only scenario. More importantly, the published choosing-an-adapter page already has a non-phantom design-context link to GitHub issue #350. Deleting the requirement means that link can disappear later without a spec change, even though the docs still describe it as the design rationale.

Replace the nonexistent repository-path link with the existing #350 issue link (scope it to Choosing an Adapter if necessary), retain the behavioral-contract link where it actually exists, and update AC 14's coverage-map wording to match. The function rename and deletion of the nonexistent docs/proposals path are otherwise correct.

langwatch-agent (commented):
1 P2 — the spec cleanup removes a still-satisfiable design-context requirement; see the inline comment. All other reviewed changes are correct.

rogeriochaves (approved):

PR #837, opened 2026-07-22
fix(judge): protect unbounded judge transcript for non-litellm agents

### LINKED ISSUE #836: JudgeAgent trace-size protection only covers litellm spans — raw transcript block is unbounded for non-litellm agent adapters
### Environment
- `langwatch-scenario==0.7.26`
- `litellm==1.80.0`
- Judge model tested: `openai/gpt-4o-mini` (128K context, confirmed via litellm's own `model_prices_and_context_window_backup.json`)

### Summary
`JudgeAgent.call()` builds two separate blocks for the judge prompt:
1. `digest` — from `self._build_trace_digest(spans)`, gated by `is_large_trace = estimate_tokens(full_digest) > DEFAULT_TOKEN_THRESHOLD` (8192 tokens, `scenario/_judge/estimate_tokens.py:10`). When large, this switches to a structure-only view plus a bounded `expand_trace`/`grep_trace` discovery loop (`TOOL_RESULT_TOKEN_BUDGET=4096`, `max_discovery_steps=10`).
2. `transcript` — from `JudgeUtils.build_transcript_from_messages(input.messages)`. This is JSON-serialized and concatenated **with no token or character cap of any kind** (the only truncation in that function is `_truncate_base64_media`, which only strips embedded image/base64 data).

The `spans` that drive `is_large_trace` come exclusively from `JudgeSpanCollector`, which only ever contains spans created by `self._trace.autotrack_litellm_calls(litellm)` — i.e., completions made through `litellm` in the harness's own process. **Any agent under test that calls its own backend directly (REST/SSE/gRPC/etc., not via `litellm`) produces tool-call data that never becomes a span.** For such agents, `spans` stays small, `is_large_trace` never trips, and the entire size-management system is inert — while the actual tool-call content (which can be arbitrarily large: SQL result sets, long documents, etc.) flows unbounded through the unprotected `transcript` block instead.

### Repro
We wrap a proprietary agent API. Its tool calls execute server-side and come back to us as a JSON `tool_call_history` blob over SSE (not `litellm`), which we convert to standard OpenAI-shaped `tool`/`assistant` messages and append to scenario state — the same shape `input.messages` would have for a `litellm`-based agent.

For one real scenario, the agent made 27 tool calls (several returning large SQL result sets). Judged with `gpt-4o-mini`:
- The judge's own `finish_test` call succeeded (structured, well-formed output — not a context-length error from the provider).
- It reported `faithfulness: 0.0`, reasoning: *"No tool calls were made to verify the claims."* — false; all 27 calls and their results are present in `input.messages` / the `<transcript>` block we can independently confirm was built from them.

We believe the transcript (uncapped) grew large enough that the small/cheap judge model's attention effectively dropped or ignored large portions of it ("lost in the middle"), while `is_large_trace` never triggered because `spans` (litellm-only) stayed small — so the discovery-loop protection that exists specifically for this situation never activated.

### Expected behavior
Trace-size protection should be based on the actual prompt content sent to the judge (i.e., transcript size, not just span/digest size) — or at minimum, `build_transcript_from_messages` should have some bounded fallback (character/token cap, or its own discovery-style access pattern) so a large `transcript` block can't silently bypass the only size-management system the library has. Any agent adapter that doesn't route tool calls through `litellm` is currently unprotected regardless of how it reports tool-call history.

### Suggested directions (not prescriptive)
- Compute `is_large_trace` (or an equivalent decision) from the *actual* assembled prompt size (transcript + digest), not from `spans` alone.
- Apply the same discovery-loop/expand-trace pattern to `transcript` content when it exceeds a threshold, rather than only to the span digest.
- At minimum, document that `AgentAdapter` implementations not using `litellm` for their own calls get no trace-size protection, so users know to build their own guard.

### Why this matters
This fails silently and confidently — the judge returns a normal, structured, plausible-looking verdict with specific (wrong) reasoning, not an error. For a framework whose output is trusted for pass/fail decisions, a silent false-negative like this is worse than a crash.

Happy to share a minimal repro harness if useful (a fake `AgentAdapter` that returns N large tool-call messages without touching `litellm`, judged against a small model).

Aryansharma28:
Fixes #836

## Problem

`JudgeAgent`'s trace-size protection (structure-only digest + `expand_trace`/`grep_trace` discovery loop) is gated entirely on `is_large_trace`, which is computed **only** from spans collected by `JudgeSpanCollector` — i.e. spans produced by `autotrack_litellm_calls`. Any `AgentAdapter` that talks to its own backend directly (REST/SSE/gRPC/etc.) and never routes calls through `litellm` produces tool-call messages that never become spans.

For such agents, `spans` stays empty/small forever, so `is_large_trace` never trips — while the actual `transcript` block (built from `input.messages` via `JudgeUtils.build_transcript_from_messages`) has **no size cap of any kind** beyond stripping base64 media. Large tool-call content (SQL result sets, long documents, etc.) flows into the judge prompt completely unbounded.

The reporter hit this with a proprietary agent whose 27 tool calls (with large SQL result payloads) never became spans; the judge (`gpt-4o-mini`) silently dropped/ignored large portions of the unbounded transcript ("lost in the middle") and returned a confident but wrong `faithfulness: 0.0` verdict claiming "no tool calls were made" — a false negative with no error signal at all.

## Fix

- `is_large_trace` is now `is_large_span_trace OR is_large_transcript`, where `is_large_transcript` is computed from the actual transcript's own estimated token count — independent of spans.
- When the transcript itself is large, it's rendered as a structure-only **skeleton** (one line per message: index, role, tool-call name(s), estimated size — no content), the same pattern already used for large span digests.
- Two new discovery tools, `expand_transcript(indices)` and `grep_transcript(pattern)`, mirror `expand_trace`/`grep_trace` but operate on message position instead of span ID, so the judge can still reach full content on demand instead of it being force-fed (or silently dropped) on every call.
- `JudgeUtils.build_transcript_from_messages` is refactored to expose `render_transcript_lines`, a per-message renderer shared by the full transcript builder and the new discovery tools, so `expand_transcript` shows byte-identical content to what the full transcript would have shown.
- Small transcripts (the common case) are completely unaffected — same full inline rendering as before.

## Testing

- New `tests/test_judge_transcript_size_protection.py` reproduces the exact bug: empty `spans` (simulating a non-litellm agent) + a large transcript (27 tool calls with large payloads, mirroring the reporter's repro). Verified this test **fails against unmodified `main`** (4/5 cases) and passes after the fix.
- Full existing judge test suite (`test_judge_agent*.py`, `test_judge_discovery_*.py`, `test_judge_utilities.py`, `test_judge_force_verdict_hardening.py`) passes unchanged — the span-based path is untouched.
- Full `pytest tests/` run: 1139 passed (the only failures are pre-existing e2e tests requiring live API keys, unrelated to this change).
- `pyright` clean on all touched files.

## Test plan
- [x] Reproduced the bug with a failing test against `main`
- [x] Fix makes the new test pass
- [x] Existing judge/discovery test suite passes unchanged
- [x] Full python test suite passes (excluding pre-existing env-gated e2e failures)
- [x] `pyright` clean

### DIFF
--- a/python/examples/test_judge_transcript_size_protection_e2e.py
+++ b/python/examples/test_judge_transcript_size_protection_e2e.py
@@ -60,13 +60,17 @@ class NonLitellmSalesAgent(scenario.AgentAdapter):
                 f"row_{i}_{j}=value_{j}" for j in range(100)
             )
             messages.append({"role": "tool", "tool_call_id": call_id, "content": f"[{row}]"})
+        # Deliberately generic -- NEEDLE_FACT must not be restated here. If it
+        # were, the judge could pass by reading this summary alone (e.g. via
+        # a cheap expand_transcript on just the last message), without ever
+        # searching the buried tool result, which would defeat the point of
+        # this test: proving the judge can dig a fact out of a large,
+        # non-litellm tool-call history rather than relying on it being
+        # conveniently restated in the final assistant turn.
         messages.append(
             {
                 "role": "assistant",
-                "content": (
-                    "I checked 27 sales records. Order ORD-88421 was refunded "
-                    "for $204.50."
-                ),
+                "content": "I checked 27 sales records.",
             }
         )
         return messages
--- a/python/examples/test_judge_transcript_size_protection_e2e.py
+++ b/python/examples/test_judge_transcript_size_protection_e2e.py
@@ -1,0 +1,101 @@
+"""
+End-to-end example: a non-litellm agent adapter with a large tool-call
+transcript, judged by a real JudgeAgent.
+
+Reproduces the exact shape of issue #836: an AgentAdapter that never routes
+its own tool calls through litellm (so JudgeSpanCollector never sees them)
+returns many large tool-call/tool-result messages in one turn. Before the
+fix, this content flowed into the judge prompt completely unbounded and the
+judge's size-management system (gated only on litellm spans) never
+activated.
+
+This test buries a single fact deep inside 27 large tool-call results and
+asks the judge to verify it -- proving the judge can still find and use
+that fact via the transcript's own discovery tools (expand_transcript /
+grep_transcript) instead of the content silently exceeding the judge
+model's effective attention ("lost in the middle").
+"""
+
+import json
+
+import pytest
+
+import scenario
+from scenario.types import AgentInput, AgentReturnTypes
+
+scenario.configure(default_model="openai/gpt-4.1-mini")
+
+NEEDLE_QUERY_INDEX = 13
+NEEDLE_FACT = "order_id=ORD-88421 status=REFUNDED refund_amount=204.50"
+
+
+class NonLitellmSalesAgent(scenario.AgentAdapter):
+    """Simulates an agent that talks to its own backend directly (e.g. a
+    proprietary REST/SSE API) and never calls litellm itself for tool
+    execution -- so none of its tool calls ever become an OpenTelemetry
+    span. Returns 27 large tool-call/tool-result message pairs in one turn,
+    matching the reporter's real repro shape (issue #836)."""
+
+    async def call(self, input: AgentInput) -> AgentReturnTypes:
+        messages = []
+        for i in range(27):
+            call_id = f"call_{i}"
+            messages.append(
+                {
+                    "role": "assistant",
+                    "content": None,
+                    "tool_calls": [
+                        {
+                            "id": call_id,
+                            "type": "function",
+                            "function": {
+                                "name": "run_sql",
+                                "arguments": json.dumps({"query": f"SELECT * FROM sales_{i}"}),
+                            },
+                        }
+                    ],
+                }
+            )
+            row = NEEDLE_FACT if i == NEEDLE_QUERY_INDEX else ",".join(
+                f"row_{i}_{j}=value_{j}" for j in range(100)
+            )
+            messages.append({"role": "tool", "tool_call_id": call_id, "content": f"[{row}]"})
+        messages.append(
+            {
+                "role": "assistant",
+                "content": (
+                    "I checked 27 sales records. Order ORD-88421 was refunded "
+                    "for $204.50."
+                ),
+            }
+        )
+        return messages
+
+
+@pytest.mark.agent_test
+@pytest.mark.asyncio
+async def test_judge_finds_needle_in_large_non_litellm_transcript():
+    """A judge criterion that depends on a fact buried inside a large,
+    non-litellm tool-call transcript must still be verifiable -- proving
+    #836's transcript-size protection activates for agents that never
+    produce spans."""
+    result = await scenario.run(
+        name="non-litellm large transcript judge",
+        description="User asks the agent to check on the status of order ORD-88421.",
+        agents=[
+            NonLitellmSalesAgent(),
+            scenario.UserSimulatorAgent(),
+            scenario.JudgeAgent(
+                criteria=[
+                    "Agent's tool call results confirm order ORD-88421 was refunded for $204.50",
+                ],
+            ),
+        ],
+        script=[
+            scenario.user("Can you check the status of order ORD-88421?"),
+            scenario.agent(),
+            scenario.judge(),
+        ],
+    )
+
+    assert result.success is True, result.reasoning
--- a/python/scenario/_judge/__init__.py
+++ b/python/scenario/_judge/__init__.py
@@ -13,6 +13,7 @@ from .judge_span_digest_formatter import (
 )
 from .judge_utils import JudgeUtils
 from .trace_tools import expand_trace, grep_trace
+from .transcript_tools import expand_transcript, grep_transcript
 
 __all__ = [
     "DEFAULT_TOKEN_THRESHOLD",
@@ -20,6 +21,8 @@ __all__ = [
     "JudgeUtils",
     "estimate_tokens",
     "expand_trace",
+    "expand_transcript",
     "grep_trace",
+    "grep_transcript",
     "judge_span_digest_formatter",
 ]
--- a/python/scenario/_judge/judge_utils.py
+++ b/python/scenario/_judge/judge_utils.py
@@ -177,8 +177,69 @@ def _render_tool_call(call: Any) -> Optional[str]:
     return f"{name}({rendered_args})"
 
 
+def _render_message_line(msg: Any, id_to_name: Dict[str, str]) -> str:
+    """
+    Renders a single message as one transcript line. Shared by
+    ``build_transcript_from_messages`` (the full transcript) and the
+    transcript discovery tools (``expand_transcript``/``grep_transcript``),
+    which need the exact same per-message rendering to index into.
+    """
+    role = msg.get("role", "unknown")
+    content = msg.get("content", "")
+    truncated_content = _truncate_base64_media(content)
+
+    tool_calls = msg.get("tool_calls")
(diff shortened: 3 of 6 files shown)


Aryansharma28 (commented):
## Deep self-review of #837

Since I authored this fix, I ran it adversarially against itself: exercised every guard clause directly, tried to break `expand_transcript`/`grep_transcript` with malformed LLM tool-call input, checked the combined-large-span-and-large-transcript path, and triaged every CodeRabbit finding rather than rubber-stamping or dismissing them. Three real issues surfaced and are fixed in follow-up commits (`45bd5bd`, `76c56de`). No open findings remain.

### Verdict: no blockers. Ready to merge once CI's unrelated infra issue (see below) is resolved or waived.

---

### [P1] `expand_transcript` crashed on non-integer indices from a lax LLM provider — FIXED (45bd5bd)

**Why:** `expand_transcript`'s `indices` argument is reachable from an LLM tool call. The tool schema declares `"items": {"type": "integer"}`, but not every litellm-routed provider enforces function-calling argument types as strictly as OpenAI. I reproduced the crash directly:
```python
expand_transcript(messages, indices=["1", 5])
# TypeError: '<' not supported between instances of 'str' and 'int'
```
`sorted(set(indices) - set(valid_range))` raised the moment a non-integer entered the list — an unhandled exception that would have propagated up through `_execute_discovery_tool` → `_run_discovery_loop` → `call()`, crashing the entire judge call instead of degrading gracefully.

**Fix:** coerce/filter `indices` defensively (int() with try/except, malformed entries reported back to the LLM instead of raising) — matching the "never raises" contract this file already applies to tool-call argument rendering (`_render_tool_arguments`'s docstring). Covered by `test_non_integer_indices_do_not_raise` / `test_unparseable_indices_are_reported_not_raised`.

### [P2] `build_transcript_skeleton` had no cap on its own total size (CodeRabbit) — FIXED (76c56de)

**Why:** CodeRabbit correctly flagged that the skeleton renders one line per message with no total budget. A transcript of thousands of small messages (a streaming/high-frequency tool-call agent) has a tiny per-message cost but an unbounded total — undercutting the entire point of this fix, since the skeleton is the thing that's supposed to stay under `token_threshold`. I verified: 5000 short messages produced a **32852-token skeleton against an 8192 default threshold** — 4x over.

**Fix:** bounded to `TOOL_RESULT_TOKEN_BUDGET` (~4096 tokens) with a head+tail window (start frames the scenario, end is what the judge is deciding on) and the elided middle still reachable via `expand_transcript`/`grep_transcript`. Verified the fix brings the same 5000-message case down to 4099 tokens while preserving `[0]` and `[4999]`. Covered by `TestSkeletonItselfIsBounded`.

### [P3] `expand_transcript` index matching was O(N×M) (CodeRabbit, partial) — FIXED (76c56de)

**Why:** membership-checked `entry.index in parsed_indices` against a `List[int]` rather than a set. Cheap, real, no downside to fix.

**What I did NOT apply from CodeRabbit's suggestion:** the "incrementally render + stop early" rewrite and a `maxItems` schema cap. The total-transcript-string cost this would guard against is already unconditionally paid every judge call (`JudgeUtils.build_transcript_from_messages` runs regardless of size to decide `is_large_transcript`), so the marginal protection is small relative to the added complexity, and `expand_trace`'s existing `span_ids` schema has no equivalent cap either — adding one only to the new tool would be an asymmetry without a matching real-world risk.

### Triaged and intentionally not applied

- **CodeRabbit: tighten `Any`/`Dict`/`List` to concrete generics** across `judge_utils.py`/`transcript_tools.py`/`judge_agent.py`. These match the pre-existing convention in this exact package — `_execute_discovery_tool(self, tool_call: Any, spans: Sequence[Any])` used `Any` for both parameters *before* this PR touched it. Tightening only the new code would make it inconsistent with the file it extends.
- **CodeRabbit: "use jsonify instead of json.dumps"** (ast-grep). False positive — this is a Flask-specific lint rule firing on a non-Flask codebase; there's no Flask response object anywhere near this code.
- **CodeRabbit: add concrete type annotations to test helpers.** The existing sibling test file (`test_judge_agent_progressive.py`) already uses the same loose-annotation style I matched (`def create_large_trace() -> list:`, `def mock_litellm_response(tool_name: str, args: dict):` with no return type). Diverging only in the new test file would reduce consistency, not improve it.
- **CodeRabbit: restore `ScenarioConfig.default_config` to its prior value instead of `None` in the test fixture teardown.** Valid observation, but it's copy-pasted verbatim from the same `setup_config` fixture already duplicated across every judge test file in this suite. Fixing it in isolation in one file doesn't actually improve test isolation (the other files still reset to `None`), and a suite-wide fix is out of scope for #836.

### Verified directly (adversarial, not just read)

- Reproduced the original bug against unmodified `main`: 4/5 new tests fail without the fix, pass with it.
- Combined-large-trigger case: built a scenario where both span digest *and* transcript exceed threshold simultaneously — confirmed all four discovery tools (`expand_trace`, `grep_trace`, `expand_transcript`, `grep_transcript`) are offered together with no crash.
- Empty-messages, malformed `tool_calls`, out-of-range/unparseable indices, no-grep-match, >20-grep-match, and truncation-at-budget paths all exercised directly.
- `scenario/_judge/transcript_tools.py` (the new module) is at **100% line coverage**. Remaining gaps in `judge_utils.py`/`judge_agent.py` are pre-existing, in code this PR didn't touch (base64/audio media truncation edge cases, voice modality resolution, deep verdict-parsing branches) — out of scope for #836.
- `pyright` clean on every touched file.
- Full `pytest tests/` (1139 tests): passes; only pre-existing e2e failures requiring live API keys, unrelated to this change.

### Things I couldn't verify from the diff alone

- **CI's `Test (Examples)` step is failing on this PR with `AuthenticationError: Incorrect API key provided`** for every example test. This is unambiguous: it fails before any of this PR's code runs, on every single example regardless of content, and main's most recent successful run (2026-07-20) passed this exact step — this reads as the `OPENAI_API_KEY` repo secret being invalid/rotated as of today, not something introduced by this change. I don't have access to repo secrets to confirm or fix it. Flagging so a maintainer with secret access can check, rather than blocking this PR on it.

Aryansharma28:
All 7 CodeRabbit review threads are now resolved:

- **"Cap the skeleton itself"** — fixed in `76c56de`
- **"Bound expansion before rendering"** — partially fixed in `76c56de` (O(1) set membership); the incremental-render rewrite wasn't applied (reasoning in my earlier review comment)
- **"Ensure test requires transcript discovery"** — fixed in `903b626`
- **"Tighten typings"** (judge_utils.py/transcript_tools.py/judge_agent.py) — fixed in `f3e4eb7` for the code this PR actually introduced; pre-existing untouched code and the builtin-generics style switch were deliberately left alone (would make new code inconsistent with the rest of the package)
- **"Add concrete types to test helpers"** — fixed in `f3e4eb7`
- **"Restore previous default configuration"** in the test fixture — fixed in `f3e4eb7`
- **"Replace cast(Any, messages) with a concrete type"** — fixed in `f3e4eb7` everywhere the underlying data is well-formed; the `indices` casts in the two deliberately-malformed-input tests stay `Any` since those tests exist specifically to pass data that violates the type contract

Full test suite (113 tests across the judge suite) + pyright: clean after every change. `python-ci`'s only remaining red check is `Test (Examples)`, unrelated to this PR (pre-existing broken `OPENAI_API_KEY` repo secret, flagged in my earlier review).

rogeriochaves (approved):

PR #841, opened 2026-07-23
fix(voice): point the Python STT seam at set_stt_provider

### LINKED ISSUE #743: fix(voice): scenario.configure(stt=...) is exported but wired to nothing
`scenario.configure(stt=...)` is exported (`python/scenario/__init__.py:160`) and `stt.py`'s docstring claims it installs the provider, but **no config path actually calls `set_stt_provider`** — verified: the only callers of `set_stt_provider` are `tests/voice/test_stt.py` and `test_adapter_base.py`. So the per-run/user-facing STT provider has no working entry point via `configure()`; the provider is currently reachable only via internal threading / the adapter `stt` field.

**Fix:** either wire `configure(stt=...)` to the resolution path (so a user can actually set the STT provider), or remove the stale export + docstring claim. Surfaced during the sc#740 Slice C (per-run STT isolation) planning — that slice introduces `resolve_stt_provider`; this issue is the missing user-facing wiring, tracked separately.

Refs: `stt.py:10,142-156`, `__init__.py:160`, sc#740.

C0deRatoR:
## Why

`scenario.configure(stt=...)` is what `python/scenario/voice/stt.py` told people to
call, what the no-provider warning in `_transcribe.py` told people to call, and what
`examples/voice/stt_swap.py` said it was demonstrating. The argument never existed, so
every one of those readers got `TypeError: configure() got an unexpected keyword
argument 'stt'`.

Fixes #743.

## What changed, and why not the other way

The issue offered two fixes: wire `configure(stt=...)` up, or drop the claim. This PR
drops the claim, because the repo has already decided against the argument twice:

- `docs/adr/003-voice-internal-design.md` (**Accepted**) lists "an invented
  `scenario.configure({ stt })` API" as a defect and specifies its removal.
- `docs/adr/002-voice-provider-state.md` removed it from the TypeScript surface and
  asks Python to move to **per-run** voice config, calling the module global a
  concurrency bug. Adding `stt=` to `configure()` would build a second process-wide
  provider knob, which is the "parity on the bug" that ADR says not to do.

The entry point that works today is the exported `scenario.set_stt_provider()`, which
`stt_swap.py` already calls in its own code. So the fix is to make everything users read
name that call.

- `stt.py`: module docstring names `set_stt_provider`, and records that the provider is
  process-wide and that per-run config is the target design.
- `set_stt_provider()`: rejects anything with no callable `transcribe`, at the call the
  user wrote, instead of surfacing as a missing-attribute error deep inside a
  transcription pass. This is the validation Copilot asked for on the first revision,
  moved to the real entry point. The check is structural rather than an `isinstance`,
  matching `isSttProvider` in `javascript/src/voice/config.ts`, so the same object is
  not valid in one runtime and rejected in the other.
- `transcribe()`: reads through `get_stt_provider()` so the seam has a single read
  point for the per-run migration ADR-002 describes.
- `_transcribe.py`: the no-provider warning names the call that works.
- `stt_swap.py`: docstring matches what the example does.

Per-run Python voice config is deliberately **not** in scope here. That is the larger
ADR-002 follow-up and it touches the judge, the user simulator and the executor.

## Tests

`python/tests/voice/test_stt.py`:

- `test_set_stt_provider_is_exported_from_the_package_root`
- `test_configure_does_not_accept_an_stt_argument` - pins the ADR decision so the knob
  cannot come back as a global.
- `test_no_shipped_source_advertises_configure_stt` - walks the whole shipped package
  and the examples, so a new file cannot reintroduce the false claim.
- `test_set_stt_provider_rejects_a_non_provider`
- `test_set_stt_provider_accepts_a_structural_provider` - a duck-typed provider works,
  so the Python check cannot drift from the TypeScript one.

`test_configure_does_not_accept_an_stt_argument`,
`test_no_shipped_source_advertises_configure_stt` and
`test_set_stt_provider_rejects_a_non_provider` all fail on `main` and pass here.

The `stt_swap` demo's own proof was broken and is fixed here. It asserts
`get_stt_provider() is stt`, transcribes through the lookup, and snapshots the
transcribe count as `scenario.run()` returns. Its post-run loop shares the same
accumulator, so `swap_verified` could previously be satisfied by that loop alone, with
no STT during the run at all. Both findings are CodeRabbit's.

`specs/voice-agents.feature` gains four `@unit` scenarios covering the same contract,
and `test_feature_file_contract.py` moves its pinned counts to 129 scenarios and
83 `@unit`.

## Verification

- `uv run --frozen pytest tests/voice/ -m "not integration"` - 585 passed
- `uv run --frozen pytest tests/ -m "not integration"` - the only failures are three
  live-LLM tests (`test_arun_real_llm_integration`, two in `test_red_team_agent`) that
  time out against the gateway from a laptop. They touch no STT code and pass in CI.
- `uv run --frozen pyright .` - 0 errors
- `black` and `isort` report no new findings. The remaining ones in
  `_transcribe.py`, `stt.py`, `stt_swap.py` and `test_feature_file_contract.py` are
  present on `main` unchanged, and neither tool gates `python-ci`.

There is no UI here. The proof is that the three new regression tests fail on `main`
for the three separate reasons the issue names, and that the source scan makes the
docs claim unfalsifiable going forward rather than fixed once.

### DIFF
--- a/python/examples/voice/stt_swap.py
+++ b/python/examples/voice/stt_swap.py
@@ -113,11 +113,15 @@ async def main() -> scenario.ScenarioResult:
             max_turns=4,
         )
 
+    # Snapshot before any post-run work, so the success criterion below counts
+    # only the calls the run itself made.
+    run_transcribe_count = len(_transcribe_calls)
+
     # Demonstrate the swapped provider by transcribing each audio segment
     # through the process-wide lookup.  Going via ``get_stt_provider()`` rather
-    # than the local ``stt`` handle is what makes the swap observable: a
-    # transcribe count above zero then proves the installed provider ran, not
-    # merely that this script holds a reference to one.
+    # than the local ``stt`` handle is what makes the swap observable: the
+    # identity check proves the install took effect rather than the script
+    # merely holding a reference.
     installed = get_stt_provider()
     assert installed is stt, (
         "set_stt_provider did not install the demo provider; "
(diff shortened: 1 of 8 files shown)


C0deRatoR:
@coderabbitai review

Issue #768, opened 2026-07-10
Twilio wrapper tests never exercise the real TwilioAgentAdapter.connect()


drewdrewthis:
## Background

PR #767 added wrapper-level smoke tests that drive the REAL `connect()` → `call()` → `disconnect()` flow for **OpenAI Realtime** and **Gemini Live**, faking only at the network-client boundary. Twilio did not get the same treatment.

`make_connected_twilio_adapter` (`python/scenario/voice/testing/wrapper_harness.py`) hand-reconstructs the connected state rather than calling the real `TwilioAgentAdapter.connect()`:

```python
adapter._rest = object()              # sentinel; never called
adapter._inbound_queue = asyncio.Queue()
adapter._stream_connected = asyncio.Event()
adapter._server_shutdown = asyncio.Event()
adapter._webhook_server = TwilioWebhookServer(adapter)
```

This is justified — the real `connect()` needs live Twilio REST credentials to resolve the phone-number SID — and everything the media-stream path touches downstream is the real object. But it means the hand-built state can **drift** from what a real `connect()` actually produces, which is precisely the "test green / production red" failure mode that #697's P0 demonstrated and that the wrapper harness exists to prevent.

Pre-existing (the pattern was extracted, not introduced, by #767), but now cheaper to close because the harness exists.

## What "done" looks like

Fake Twilio at the **network-client boundary** — the REST client construction — the same way #767 fakes `websockets.connect` and `genai.Client`. Then drive the real `TwilioAgentAdapter.connect()` and assert the state it builds, so `make_connected_twilio_adapter`'s hand-built equivalent is pinned to the real thing.

## Acceptance criteria

- [ ] At least one Twilio test constructs the real `TwilioAgentAdapter` and calls the real `connect()`, with the Twilio REST client faked at its construction boundary (no live credentials).
- [ ] That test asserts the connected state (`_rest`, `_inbound_queue`, `_stream_connected`, `_server_shutdown`, `_webhook_server`) matches what `make_connected_twilio_adapter` hand-builds — so drift between the two fails a test rather than hiding.
- [ ] Falsifiability: changing what real `connect()` sets up turns the new test RED.
- [ ] Test is hermetic — no API keys, no network, not skipped on a normal PR run.

## Related

- #766 — the wrapper-harness extraction (this issue is its Twilio-shaped remainder)
- #764 — the CI seam-gate that would flag new seam-only tests
- #697 — the P0 that motivated all of this

---
[...]

PR #849, opened 2026-07-25
fix(voice): wake blocked Twilio receive on disconnect

### LINKED ISSUE #757: voice/twilio: PY disconnect() leaves a blocked recv_audio parked until timeout; JS reset() rejects it
### What

A Python/TypeScript behavioural divergence in `TwilioAgentAdapter`'s teardown:

- **TS** — `InboundQueue.reset()` (`javascript/src/voice/adapters/twilio.ts`) clears buffered chunks **and rejects every parked waiter** with `TwilioAgentAdapter: stream reset before audio arrived.` A consumer blocked in `receiveAudio` fails fast.
- **PY** — `disconnect()` (`python/scenario/voice/adapters/twilio.py`) sets `self._inbound_queue = None`. It never wakes a consumer already blocked in `await self._inbound_queue.get()`; that task holds a reference to the old queue object and stays parked until its own `asyncio.wait_for` timeout elapses (up to `response_timeout`, default 60s).

### Why it matters

A `disconnect()` racing an in-flight drain stalls the drain for the full receive budget instead of failing immediately. Same adapter, same call, two very different teardown latencies depending on language.

### Suggested direction

Give the Python side an explicit wake: either put a sentinel before nulling the queue, or hold waiters explicitly (as the TS `InboundQueue` does) so `disconnect()` can cancel them.

### Context

Pre-existing; surfaced by self-review while driving #697, which touched `recv_audio` but not `disconnect()`. #697 adds an explicit `RuntimeError("inbound queue is gone…")` guard for the *post*-nulling read, which is a different (and currently unreachable) path.

---

_Planned 2026-07-31 (planner pass, grounded on `main` @ `88aec40`)._

## Acceptance Criteria

Node ids below name the file the implementer creates: `python/tests/voice/test_twilio_disconnect_wake.py`. All Python ACs build the adapter with the existing seam `make_connected_twilio_adapter()` (`python/scenario/voice/testing/wrapper_harness.py:121`) — the same seam the #695 wrapper-layer tests use. **Honest-evidence caveat (#768):** that harness hand-reconstructs connected state and does NOT call the real `connect()`, so these ACs pin `disconnect()`'s wake behaviour against hand-built state, not against production `connect()` output. No AC below may be ticked with a claim that it exercises the real `connect()`.

Every timing AC asserts a **wall-clock bound**, never "faster than `response_timeout`". The nominal `recv_audio(timeout=...)` in each test MUST be >= 30.0s so that the pre-fix adapter's failure mode is a visible 30s park, and the asserted bound is a separate, much smaller number. Default `response_timeout` is `60.0` (`python/scenario/voice/adapter.py:112`).

- [ ] **AC1 — the bug: a consumer parked BEFORE `disconnect()` is woken within 250 ms.** A task is started on `adapter.recv_audio(timeout=30.0)` and confirmed parked (queue empty, task not done after a yield). `await adapter.disconnect()` is then called. The parked task completes — by raising — with `monotonic()` elapsed from the `disconnect()` call to task completion **< 0.25 s**. **Fails if** the task is still pending 0.25 s after `disconnect()` returns, or if it only completes at the 30 s `wait_for` deadline. **Evidence:** pytest stdout for `python/tests/voice/test_twilio_disconnect_wake.py::test_parked_recv_audio_wakes_within_250ms`, with the measured elapsed seconds printed in the assertion message so the tick shows the real number, not just PASSED.

- [ ] **AC2 — the raised error is attributable and parity-matched to TS.** The exception from AC1 is a `RuntimeError` whose `str()` contains the substring `stream reset before audio arrived`. This is the TS wording verbatim from `javascript/src/voice/adapters/twilio.ts:741`: `"TwilioAgentAdapter: stream reset before audio arrived."` It is NOT `asyncio.TimeoutError`, NOT `asyncio.CancelledError`, and NOT a bare `AttributeError`. **Decision (not an implementer choice): the Python error is a Python-native `RuntimeError`, not a new class and not a `CancelledError`.** Rationale from this codebase: every other Twilio teardown failure already raises `RuntimeError` (`python/scenario/voice/adapters/twilio.py:644` "inbound queue is gone; adapter disconnected."), and `CancelledError` is a `BaseException` that would silently unwind the drain task instead of surfacing. Parity is on the *message*, not the exception class — TS has no exception taxonomy to match. **Evidence:** `pytest.raises(RuntimeError, match="stream reset before audio arrived")` in the AC1 node; pytest stdout.

- [ ] **AC3 — `disconnect()` with NO parked consumer raises nothing.** On a connected adapter with an empty inbound queue and zero waiters, `await adapter.disconnect()` returns normally, and afterwards `adapter._inbound_queue is None`. **Fails if** `disconnect()` raises, or leaves `_inbound_queue` non-`None`. **Evidence:** pytest stdout for `...::test_disconnect_with_no_parked_consumer_is_silent`.

- [ ] **AC4 — `disconnect()` is idempotent under a double call.** `await adapter.disconnect()` called twice: the second call returns normally within 1.0 s wall clock and raises nothing. (Today the second call short-circuits at `if self._rest is None: return`, `twilio.py:272-273`; this AC pins that the new wake mechanism does not break that short-circuit — e.g. by dereferencing a now-`None` `_inbound_queue`.) **Evidence:** pytest stdout for `...::test_double_disconnect_is_idempotent`.

- [ ] **AC5 — REGRESSION: `recv_audio` started AFTER `disconnect()` still hits the #697 guard.** The existing node `python/tests/voice/test_twilio_silent_stop_drain.py::test_recv_audio_with_nulled_queue_raises_connection_error` still passes unmodified — `recv_audio` on an adapter whose `_inbound_queue` is `None` raises `RuntimeError` matching `inbound queue is gone` (`twilio.py:636-646`). This is a *different* message from AC2 by design: AC2 is "you were parked and we tore down under you", AC5 is "you arrived after teardown". **Fails if** the fix rewrites, weakens, or deletes that guard, or collapses the two messages into one. **Evidence:** pytest stdout showing that exact node id PASSED, and `git diff` on `test_twilio_silent_stop_drain.py` showing that test unmodified.

- [ ] **AC6 — REGRESSION: the whole #695 silent-stop suite is still green.** `python/tests/voice/test_twilio_silent_stop_drain.py` passes in full, including the `TestRealRoute` class — the end-of-call sentinel path (loop `finally` enqueues one empty chunk, `python/scenario/voice/adapters/_twilio_server.py:405-406`) must not be disturbed by the new disconnect-time wake. In particular a normal end-of-call must still deliver an **empty `AudioChunk`**, not the AC2 `RuntimeError`. **Evidence:** pytest stdout, full-file summary line (counts visible, `0 failed`, not `skipped`).

- [ ] **AC7 — TS side is untouched and still rejects parked waiters.** `javascript/src/voice/adapters/__tests__/twilio-silent-stop-drain.test.ts` and `twilio.test.ts` pass, and `git diff --stat` shows zero lines changed under `javascript/src/voice/adapters/`. #757 is a Python-side convergence onto existing TS behaviour; any TS edit is out of scope and must be justified separately. **Evidence:** vitest summary line + `git diff --stat javascript/` output (empty).

## Design constraint the reviewer should see explicitly

The issue body offered two directions ("put a sentinel before nulling the queue, **or** hold waiters explicitly"). **AC2 rules out the sentinel**: a sentinel `put(AudioChunk(data=b""))` before nulling would wake the consumer with an *empty chunk*, satisfying AC1's timing but failing AC2's attributable error — and it would collide with the end-of-call sentinel AC6 protects. The AC set therefore steers the implementer toward explicit waiter-holding (TS `InboundQueue` parity, `twilio.ts:714-720`, `:737-744`). Deliberate; reversible by the owner with a one-line AC change.

_Also flagged: the 250 ms bound in AC1 is a chosen number, not codebase-derived (nearest precedent is `OUTER_CEILING = 2.0` at `test_twilio_silent_stop_drain.py:85`). The **shape** — a wall-clock bound, not "< response_timeout" — is the non-negotiable part._

## Proof

Run and paste, in one turn: `cd python && uv run pytest tests/voice/test_twilio_disconnect_wake.py tests/voice/test_twilio_silent_stop_drain.py -v` (full node-id list + summary line for AC1/AC3-AC6, with AC1's measured wake-elapsed seconds visible in the captured output), then `cd javascript && npx vitest run src/voice/adapters/__tests__/twilio-silent-stop-drain.test.ts src/voice/adapters/__tests__/twilio.test.ts` plus `git diff --stat javascript/` (must print nothing) for AC7. A tick without the elapsed-seconds number for AC1 is not a tick — "PASSED" alone cannot distinguish a 10 ms wake from a 29 s one under a 30 s nominal timeout.

allin2:
## Why

When the Python Twilio adapter disconnects while `recv_audio()` is already
blocked on the inbound queue, clearing the adapter's queue reference does not
wake that consumer. It remains parked until its receive timeout expires.

Closes #757

## What changed

- Reuse the adapter's existing empty `AudioChunk` end-of-stream sentinel to
  wake an in-flight receiver before the inbound queue is discarded.
- Keep the public API and normal stream shutdown behavior unchanged.
- Cover the disconnect/receive race with an observable queue double, avoiding
  fixed sleeps and external services.

## Test plan

- Confirmed the new regression test fails before the source fix with
  `TimeoutError`.
- `pytest tests/voice/test_twilio_adapter.py -q` — 24 passed
- `pytest tests/test_scenario.py -q` — 17 passed
- `pyright scenario/voice/adapters/twilio.py tests/voice/test_twilio_adapter.py`
  — 0 errors
- Capability matrix byte-identity test — passed

No playable artifact is applicable for this library-only concurrency fix.
The regression test deterministically waits until `recv_audio()` has entered
the queue's `get()` before disconnecting, then verifies it returns the existing
empty end-of-stream chunk within a bounded observation window.

### DIFF
--- a/python/scenario/voice/adapters/twilio.py
+++ b/python/scenario/voice/adapters/twilio.py
@@ -354,6 +354,13 @@ class TwilioAgentAdapter(VoiceAgentAdapter):
         self._stream_sid = None
         self._stream_connected = None
         self._stream_ws = None
+        if self._inbound_queue is not None:
+            # A recv_audio() task may already be blocked in get() on this queue.
+            # Wake that in-flight consumer before dropping our reference; setting
+            # _inbound_queue to None alone leaves it parked until its receive
+            # timeout expires (#757). An empty chunk is the adapter's existing
+            # end-of-stream signal and lets the shared drain finish cleanly.
+            self._inbound_queue.put_nowait(AudioChunk(data=b""))
         self._inbound_queue = None
         self._stream_ended = False
         self._frames_received = 0
--- a/python/tests/voice/test_twilio_adapter.py
+++ b/python/tests/voice/test_twilio_adapter.py
@@ -9,7 +9,7 @@ from typing import Any, Optional
 
 import pytest
 
-from scenario.voice import TwilioAgentAdapter
+from scenario.voice import AudioChunk, TwilioAgentAdapter
 
 
 def _make_adapter(**overrides: Any) -> TwilioAgentAdapter:
@@ -112,11 +112,43 @@ def _install_fake_rest(monkeypatch: Any) -> list[FakeREST]:
 
 # ---------------------------------------------------------------- connect/disconnect
 
+
+class _ObservableAudioQueue(asyncio.Queue[AudioChunk]):
+    """Queue double that exposes when a consumer has entered ``get()``."""
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.consumer_waiting = asyncio.Event()
+
+    async def get(self) -> AudioChunk:
+        self.consumer_waiting.set()
+        return await super().get()
+
+
 @pytest.mark.asyncio
 async def test_connect_requires_public_base_url():
     a = _make_adapter(public_base_url=None)
     with pytest.raises(RuntimeError, match="public_base_url"):
         await a.connect()
+
+
+@pytest.mark.asyncio
+async def test_disconnect_wakes_recv_audio_waiting_on_inbound_queue():
+    """A disconnect racing an in-flight receive must not wait for its 60s budget."""
+    a = _make_adapter()
+    queue = _ObservableAudioQueue()
+    a._rest = object()  # type: ignore[assignment]  # connected-state sentinel
+    a._inbound_queue = queue
+    a._stream_ws = object()
+    a._stream_sid = "MZ757"
+
+    receive = asyncio.create_task(a.recv_audio(timeout=60.0))
+    await asyncio.wait_for(queue.consumer_waiting.wait(), timeout=1.0)
+
+    await a.disconnect()
+
+    chunk = await asyncio.wait_for(receive, timeout=1.0)
+    assert chunk.data == b""
 
 
 @pytest.mark.asyncio


langwatch-agent (commented):
No blocking findings from static review of this external contribution (Twilio disconnect fix). Required checks are currently passing. The contributor branch was not checked out or executed.

langwatch-agent:
**Human Review Brief** (attention map, not a review)

**Mode:** Targeted Review (outside contributor, voice adapter lifecycle)
**Linked issue:** none declared.

**What this touches:** Twilio receive on disconnect. These are teardown paths, which is where voice adapters usually leak.

**Top inspection targets**
1. `Must Check` - the fix does not swap a hang for a leak. Waking a blocked receive or terminating a process cleanly must still close the session, end any open spans, and release the transport, or the symptom moves rather than resolving.
2. `Must Check` - idempotent teardown. Disconnect handling commonly runs twice (explicit close plus a transport event); confirm a second call is harmless.
3. `Ask Author` - whether this is reproducible in a test, since the existing test surface for abnormal termination is usually thin.

**Production risk:** leaked processes or unclosed sessions accumulating under repeated disconnects.

**Questions for the author (max 3)**
1. Is teardown safe to call twice?
2. Is there a regression test for the abnormal-disconnect path?

langwatch-agent (commented):
Static review of external contribution at `d69df0e0e092ef2eb9e5d4753b604fef0a941824` (no PR-branch code or dependencies executed). No merge-blocking issue found. Enqueuing an empty chunk before clearing the inbound queue wakes a receiver already blocked in `recv_audio`; because the queue is unbounded, the non-blocking enqueue cannot fail due to capacity. The change preserves the existing disconnected behavior for subsequent receives. CI evidence was considered but not rerun under the external-contribution trust boundary.

rogeriochaves (approved):

langwatch-agent:
Reviewed on request. **Short answer: yes, the fix makes sense for #757.** The mechanism is right, it is minimal, and I verified its central claim rather than taking the test plan's word for it. Three things a reviewer should weigh before merging, one of which is a concrete ask.

## Verified

Checked out the branch and ran it both ways:

```
tests/voice/test_twilio_adapter.py                      24 passed
same file, with ONLY the disconnect() hunk reverted      1 failed
  FAILED test_disconnect_wakes_recv_audio_waiting_on_inbound_queue
```

So the regression test is bound to the source change and is not vacuous.

Two things I checked that could have made this wrong, and neither does:

- `_inbound_queue` is built as a bare `asyncio.Queue()` (twilio.py:247) with no `maxsize`, so `put_nowait` in teardown cannot raise `QueueFull`.
- A second `disconnect()` is a no-op, since `_inbound_queue` is `None` by then.

The approach is also not novel here, which is a point in its favour. `recv_audio` already documents (twilio.py:629-635) that the media-stream loop's `finally` enqueues one empty chunk precisely because "flipping `_stream_ended` alone wakes nobody", and that further empty chunks are synthesized after drain. This hunk makes `disconnect()` do what the loop's `finally` already does.

## 1. It closes the latency half of the divergence, not the outcome half

#757 is titled on a PY/TS divergence. TS rejects the parked waiter with `new Error("TwilioAgentAdapter: stream reset before audio arrived.")` (twilio.ts:741). This PR wakes the PY consumer with a successful empty `AudioChunk`.

After this, the two sides agree on *how fast* teardown returns and still disagree on *what it returns*: TS raises, PY hands back a normal end-of-stream value. A caller blocked in `recv_audio` when the adapter is torn down under it now gets the same value as a clean hangup.

That may well be the intended call, since it is what the rest of this adapter does. But if `Closes #757` stands, it is worth saying in the PR that the outcome divergence is deliberately left open, rather than letting the issue close implying full parity.

## 2. It adds a third producer of the ambiguous empty chunk, which #868 is open about

#868 is specifically that this adapter's empty-`AudioChunk` sentinel makes a transport crash indistinguishable from a clean hangup, and that the sibling Pipecat adapter raises `AgentStreamEndedError` instead. This PR adds disconnect-during-receive as a third source of that same value.

Not a blocker, and not a reason to redesign here: consistency with the adapter's current contract is the right default for a fix this size. But #868 and this now interact, and whoever picks up #868 should know the surface grew.

## 3. The ACs name a seam and a file, and the PR uses neither (the concrete ask)

#757's acceptance criteria say the test goes in `python/tests/voice/test_twilio_disconnect_wake.py` and builds the adapter through the existing seam `make_connected_twilio_adapter()`. That function does exist, at `python/scenario/voice/testing/wrapper_harness.py:121`. This PR instead adds the test to `test_twilio_adapter.py` and hand-assigns four privates:

```python
a._rest = object()
a._inbound_queue = queue
a._stream_ws = object()
a._stream_sid = "MZ757"
```

The ACs anticipated this exact weakness, in their "honest-evidence caveat (#768)": even the harness only hand-reconstructs connected state. Hand-poking privates at the call site is a step further from production than the seam the ACs chose, and it will drift silently if the connected-state fields change.

Switching to `make_connected_twilio_adapter()` looks cheap and would put this on the same footing as the #695 wrapper-layer tests. Worth doing before merge, or worth a line saying why the seam did not fit.

## What is good and should not be lost in the above

The `_ObservableAudioQueue` double is the right shape: it waits until the consumer has actually entered `get()` instead of sleeping, so the test is deterministic rather than timing-dependent. That is what the ACs asked for and it is the part of this PR I would copy elsewhere.

---
<sub>🤖 **Filed by `tech-debt-fixer` via the create-issue procedure, on behalf of the fleet.** Authored under the owner's GitHub token for API access — the content is the fleet's, not his. Owner-authored and fleet-authored issues are otherwise indistinguishable in this repo; this footer is the only signal.</sub>

PR #850, opened 2026-07-25
fix(voice): ignore empty first chunk for speaking event

### LINKED ISSUE #569: fix(voice/ts): AgentSpeakingEvent.set() fires on an empty first audio chunk (adapter.runtime.ts)
Flagged in the #561 /review (non-blocking — no observed failure). The drain in `javascript/src/voice/adapter.runtime.ts` (~:318) calls `speakingEvent.set()` on the first `receiveAudio()` even when `first.data.length === 0` → can falsely signal "agent speaking" on a no-audio turn, which the interrupt-truncation path keys off. Guard: only `set()` when the first chunk has audio. Adjacent to the barge-in/truncation logic fixed in #561. Epic #370.

allin2:
## Why

An empty first audio chunk does not itself prove that the agent spoke, but the TypeScript voice runtime resolved `AgentSpeakingEvent` immediately. That could make interruption logic treat a silent turn as active speech, while simply gating on the first chunk could miss real audio that follows an empty prefix.

Closes #569

## What changed

- Resolve `AgentSpeakingEvent` and start playback timing only when the drain observes its first playable audio chunk, whether that is the initial chunk or a later one.
- Keep the event unset for a response containing only empty terminal chunks, and preserve the public API.
- Cover both empty-only and empty-prefix-then-audio paths with deterministic in-memory regression tests.

## Test plan

- Confirmed `does not set the speaking event when the first chunk is empty` failed before the runtime fix (`expected false`, received `true`).
- Confirmed `sets the speaking event when audio follows a leading empty chunk` failed before the review fix (`expected true`, received `false`).
- `pnpm exec vitest run src/voice/__tests__/interrupt-truncation.test.ts` — 12 passed.
- `pnpm exec vitest run src/voice/__tests__ src/voice/adapters/__tests__` — 45 files passed, 1 skipped; 436 tests passed, 1 skipped.
- `pnpm test:ci` — 89 files passed, 1 skipped; 1036 tests passed, 4 skipped.
- `pnpm lint:all`
- `pnpm typecheck:all`
- `pnpm build:all`
- `pnpm smoke:dist` — CJS and ESM distributions load successfully.

No playable artifact — this is an internal voice-runtime state fix. Run the named regression tests in the Test plan; they verify the event stays unset for an empty-only response and becomes set when playable audio follows an empty prefix.

### DIFF
--- a/javascript/src/voice/adapter.runtime.ts
+++ b/javascript/src/voice/adapter.runtime.ts
@@ -721,8 +721,8 @@ async function drainInner(
   );
   if (first.data.length > 0) {
     onFirstChunk();
+    speakingEvent.set();
   }
-  speakingEvent.set();
 
   const chunks: AudioChunk[] = [first];
   let accumulated = first.durationSeconds;
--- a/javascript/src/voice/adapter.runtime.ts
+++ b/javascript/src/voice/adapter.runtime.ts
@@ -719,10 +719,16 @@ async function drainInner(
     "voice.audio.first_chunk_latency_ms",
     Math.round(performance.now() - receiveStart),
   );
-  if (first.data.length > 0) {
+  // The initial receive can be an empty prefix while the drain still yields
+  // audio. Publish speech on the first playable chunk wherever it appears.
+  let sawPlayableChunk = false;
+  const markFirstPlayableChunk = (chunk: AudioChunk): void => {
+    if (sawPlayableChunk || chunk.data.length === 0) return;
+    sawPlayableChunk = true;
     onFirstChunk();
     speakingEvent.set();
-  }
+  };
+  markFirstPlayableChunk(first);
 
   const chunks: AudioChunk[] = [first];
   let accumulated = first.durationSeconds;
@@ -761,6 +767,7 @@ async function drainInner(
       terminatedReason = "terminal_chunk";
       break;
     }
+    markFirstPlayableChunk(next);
     chunks.push(next);
     accumulated += next.durationSeconds;
     maybeWarnSoftCap();
--- a/javascript/src/voice/__tests__/interrupt-truncation.test.ts
+++ b/javascript/src/voice/__tests__/interrupt-truncation.test.ts
@@ -135,6 +135,19 @@ describe("defaultVoiceCall publishes the agent speaking event", () => {
     expect(adapter.agentSpeakingEvent!.isSet()).toBe(false);
   });
 
+  it("sets the speaking event when audio follows a leading empty chunk", async () => {
+    const adapter = new SpeakingAdapter([
+      new AudioChunk({ data: new Uint8Array(0) }),
+      tone(0.1, "after empty"),
+    ]);
+
+    await defaultVoiceCall(adapter, bareInput);
+
+    expect(adapter.agentSpeakingEvent).toBeDefined();
+    expect(adapter.agentSpeakingEvent!.isSet()).toBe(true);
+    await expect(adapter.agentSpeakingEvent!.wait()).resolves.toBeUndefined();
+  });
+
   it("clears the event at the start of the next turn (re-armed)", async () => {
     // Turn-2's first chunk is delayed so the cleared-but-not-yet-set window is
     // observable across a macrotask yield. Without it the immediate chunk would
--- a/javascript/src/voice/__tests__/interrupt-truncation.test.ts
+++ b/javascript/src/voice/__tests__/interrupt-truncation.test.ts
@@ -124,6 +124,17 @@ describe("defaultVoiceCall publishes the agent speaking event", () => {
     await expect(adapter.agentSpeakingEvent!.wait()).resolves.toBeUndefined();
   });
 
+  it("does not set the speaking event when the first chunk is empty", async () => {
+    const adapter = new SpeakingAdapter([
+      new AudioChunk({ data: new Uint8Array(0) }),
+    ]);
+
+    await defaultVoiceCall(adapter, bareInput);
+
+    expect(adapter.agentSpeakingEvent).toBeDefined();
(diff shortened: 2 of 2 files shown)


javascript/src/voice/adapter.runtime.ts
```diff
@@ -721,8 +721,8 @@ async function drainInner(
   );
   if (first.data.length > 0) {
     onFirstChunk();
+    speakingEvent.set();
```
langwatch-agent, line 729: **P1 — A leading empty chunk does not mean the response contains no playable audio.** `drainInner` continues reading after this first chunk, so an input sequence such as `[empty, non-empty, end]` emits audio but never calls `onFirstChunk()` or sets `speakingEvent`. That leaves the speaking/interrupt state inconsistent for a real response. Keep discarding zero-length prefixes until the first playable chunk (or terminal end), then set the event at that point; add a regression test for an empty chunk followed by audio.

langwatch-agent (changes requested):
Requesting changes: 1 P1 correctness issue. See the inline comment.

javascript/src/voice/adapter.runtime.ts
allin2: Addressed in `3380f97`.

- `drainInner` now marks the first playable chunk wherever it appears, so a leading empty chunk no longer suppresses `onFirstChunk()` or `speakingEvent.set()`.
- Added the requested regression for `[empty, non-empty, end]`, while retaining the empty-only assertion.
- Verified `interrupt-truncation.test.ts` (12 passed), the Voice suites (436 passed, 1 skipped), and the full JavaScript suite (1036 passed, 4 skipped). Build, lint, typecheck, and CJS/ESM smoke checks also pass.

Could you please re-review?

langwatch-agent (commented):
Static review found no actionable voice-runtime or liveness issue in the current fork change. Empty initial chunks no longer signal speech, while the later playable chunk still starts the recorder and speaking event; interruption remains time-bounded elsewhere. I did not execute the untrusted head or its scripts.

langwatch-agent (commented):
Static review of external contribution at `3380f97992847641f75449b012c4eb74f9f1da03` (no PR-branch code or dependencies executed). No merge-blocking issue found. The speaking event is now set only after the first playable audio chunk, while an initially empty chunk no longer produces a false speaking transition; the expected empty-first-then-audio and empty-only behavior is covered. CI evidence was considered but not rerun under the external-contribution trust boundary.

langwatch-agent (commented):

LangWatch-Review: verdict=clean sha=3380f97992847641f75449b012c4eb74f9f1da03 p0=0 p1=0 p2=0 p3=0

langwatch-agent (commented):

LangWatch-Review: verdict=clean sha=3380f97992847641f75449b012c4eb74f9f1da03 p0=0 p1=0 p2=0 p3=0

langwatch-agent:
## Human Review Brief

> [!CAUTION]
> **This changes when `AgentSpeakingEvent` fires, and interruption logic reads that event.**
>
> Previously the TypeScript voice runtime resolved the event on the first audio chunk, empty or not. An empty first chunk does not prove the agent spoke, so a silent turn could be treated as active speech by interruption logic. The fix waits for the first *playable* chunk, whether that is the initial one or a later one, and leaves the event unset for a response that contains only empty terminal chunks.
>
> Both halves matter and the PR gets both. Gating naively on the first chunk would have missed real audio arriving after an empty prefix, which is the obvious wrong fix. The consequence to think about is the other direction: a turn that produces no playable audio now never resolves the event at all. Anything awaiting it needs to handle that, and the timing of playback start moves later by however long the empty prefix lasts.

| | |
| --- | --- |
| **Mode** | `Fast Skim`. One runtime file plus tests, in the JavaScript SDK. |
| **Issue** | `Closes #569`. Matches. |
| **State** | +34 / -3 across 2 files. CI green, mergeable. Requested reviewer `sergioestebance`. Open since 25 July, so a month. Outside contributor. |
| **Evidence** | Tests cover both the empty-only case and the empty-prefix-then-audio case, which are exactly the two the fix has to separate. Public API preserved. |
| **Where to look** | What awaits `AgentSpeakingEvent`, and what it does when the event never resolves |

<details>
<summary><b>What you are looking at, if the voice subsystem is not your area</b></summary>

</details>

<details>
<summary><b>The one useful check</b></summary>

Find the callers. If anything `await`s `AgentSpeakingEvent` without a timeout, a silent turn now hangs it where before it resolved immediately and wrongly. That is a better failure than the one being fixed and it is still a behaviour change worth being deliberate about.

Playback timing also now starts later, at the first playable chunk rather than the first chunk. If any metric measures time-to-first-audio off that clock, its numbers move.

</details>

> [!NOTE]
> **The event should mean the agent actually spoke. It now does, and the case where it never fires is the one to trace.**

langwatch-agent:
> [!NOTE]

<details open>
<summary><b>What you are looking at, if the voice path is not your directory</b></summary>

Scenario's voice support drives a realtime model over a streaming connection: audio frames go up, audio and events come back, and the library turns that stream into the turn structure the rest of scenario reasons about. A **turn** is one participant speaking until they stop.

The hard part is that the stream does not announce turns cleanly. Frames arrive continuously, silence is not a message, and an interruption means the model was mid-turn when the user started talking again. So the adapters carry a small state machine that decides where one turn ends and the next begins, driven by events like speaking-started and speaking-stopped plus the frames themselves.

That state machine is where nearly every voice defect lives, and it fails in a characteristic way: not a crash, but a turn boundary in the wrong place, which surfaces much later as a judge reading a transcript that does not match what was said. When reading a voice fix, the question is almost always "what does this do to the boundary", not "does this code work".
</details>

Issue #627, opened 2026-06-09
bug: scenario.__version__ hardcoded "0.1.0" — never tracks released distribution version


drewdrewthis:
`scenario.__version__` is a hardcoded `"0.1.0"` and never tracks the released distribution version. Anyone reading the SDK's runtime version (logging, telemetry, bug reports, conditional feature gates) gets a stale, misleading value.

## Expected

`import scenario; scenario.__version__` returns the version of the installed distribution — e.g. `0.7.29` after the latest release. The runtime version attribute is a single source of truth, kept in sync with the published package version (e.g. derived from `importlib.metadata.version("langwatch-scenario")`, or written by the release/publish workflow).

## Actual

`scenario.__version__` is a frozen literal `"0.1.0"`. The release/publish workflow tags `python/vX.Y.Z` and builds with the `pyproject.toml` version, but does not update the in-code attribute — so it has stayed `"0.1.0"` across every release while the distribution version moved on.

## Reproduction

Observed while verifying the `0.7.29` release in a clean virtualenv:

```
$ pip install 'langwatch-scenario==0.7.29'
$ pip show langwatch-scenario | grep Version
Version: 0.7.29
$ python -c "import scenario; print(scenario.__version__)"
0.1.0
```

`pip show` reports the correct distribution version; `scenario.__version__` does not match it.

## Scope note

This report is Python-side. The TypeScript/JS port may have the same drift (a hardcoded version constant that the publish flow does not update) — worth confirming during investigation.

---
*Classification: Bug*
*Status: stub — investigation pending. Issue is not-ready until the investigation step completes.*

Issue #597, opened 2026-06-01
UserSimulatorAgent default prompt encodes text-chat style — breaks voice scenarios


rogeriochaves:
## Summary

`UserSimulatorAgent`'s default system prompt instructs the LLM to talk like a text-chat user:

> "Approach this naturally, as a human user would, with very short inputs, few words, all lowercase, imperative, not periods, like when they google or talk to chatgpt."

(see `python/scenario/user_simulator_agent.py`, similar in JS)

That's exactly right for TEXT scenarios but **actively wrong for voice scenarios**. When the simulator's text is then run through TTS (ElevenLabs / OpenAI / etc.), the synthesised audio sounds robotic and telegraphic — nothing like how a real caller speaks on a phone.

## Repro

Run any of the voice examples (e.g. `python/examples/voice/angry_customer.py`) and listen to the user-sim audio segments. Sample transcripts we saw on a customer-support demo:

- `[stressed] locked out after three failed logins from lisbon, need password reset right now`
- `[stressed] name jamie wilson email jamie.wilson@example.com last 4 4242 send reset there now`
- `[stressed] jamie wilson, jamie.wilson@example.com, last 4 4242. actually just send the reset link to my new email`

Compare to what a real caller would actually say:

> "Hi, I'm Jamie Wilson — I've been locked out of my account after three failed login attempts from a hotel in Lisbon, and I really need a password reset right now. My email is jamie.wilson@example.com and the last four of my card are 4242."

The telegraphic style also makes the **agent under test** perform worse — bots are trained on natural language and stumble when the user sounds like a search query.

## Proposed fix

When `UserSimulatorAgent` is being used with a TTS voice (i.e. `voice=...` is set, or the agent is in a voice script), the prompt should swap the "talk like google" instruction for something like:

> "You are SPEAKING on a phone, not typing. Talk in natural spoken sentences — full clauses with subjects and verbs, not telegraphic phrases. Real callers don't speak like google queries."

Either:
1. Detect voice mode automatically (presence of `voice=` or audio script step), and switch the default prompt block; or
2. Expose a `voice_mode: bool` / `style: Literal["text", "voice"]` param so users can opt in.

## Workaround

We've added the spoken-sentence nudge inline to every persona block in our demo project as a stopgap. Works fine but feels like a thing the framework should own, since every voice user runs into the same friction.

Happy to send a PR if a maintainer points at the preferred prompt-injection seam.

PR #594, opened 2026-06-01
chore: scrub vestigial AC-reference tags from code + specs

drewdrewthis:
## Why

The codebase carried ~130 bare acceptance-criteria tags (`AC-1.1`, `AC-14`, `AC-X4`, and companion proposal-section refs like `§4.3`) scattered across specs, scripts, and tests. These point at **external working artifacts** — a GitHub-issue AC list and a proposal doc's sections — that a code reader cannot open or resolve. They're vestigial scaffolding from the original implementation, now pure noise to anyone reading the code.

This applies the same hygiene the repo's PR-description policy expects (no non-linkable working-artifact references) to the code itself.

_No backing issue — this is direct repo-hygiene cleanup surfaced during the PR-description-policy work. The broader follow-up (bare `§`-section provenance refs) is tracked separately in #595; this PR does not `Close` an issue._

## What changed

- **Removed every `AC-N.M` / `§x` tag across 10 files**, preserving the human-readable substance each one annotated — comment prose, `echo`/`pass`/`fail` message strings, docstrings, and spec section banners. Example: `# AC-14 — Voice demo recordings` → `# Voice demo recordings`; `echo "--- AC-1.1: pull_request_target trigger ---"` → `echo "--- pull_request_target trigger ---"`.
- **No behavioral change.** Only comments, string literals, and docstrings were touched — verified that zero `Scenario:` / `Given` / `When` / `Then` / `@tag` lines changed in the feature files, and zero logic lines (grep/conditionals/patterns) changed in `validate-pr-auto-approve.sh`.

## Test plan

- `python/tests/voice/test_feature_file_contract.py` passes — the structural contract (`len(scenarios) == 108`, tag split `(75, 8, 25)`) is unchanged, proving no scenario/tag lines were disturbed.
- `scripts/validate-pr-auto-approve.sh` behaves identically (its check logic is byte-for-byte unchanged; only message-string labels differ).

## How I can prove I was successful

- `git grep -rI "AC-[0-9X]"` on this branch returns **zero** matches (down from 130).
- The diff is comments/strings/docstrings only: `git diff origin/main...HEAD` shows 10 files, +100/−145, with no Gherkin step or shell-logic lines among the changes.

## Anything surprising?

[...]

PR #857, opened 2026-07-28
fix(security): close every open alert on scenario python/uv.lock

langwatch-agent:
## What

Closes **all 38** open alerts on `python/uv.lock`. With #873 (the 117 on the JavaScript manifests) that is the entire scenario inbox, 155 of 155.

| package | from | to | alerts |
|---|---|---|---|
| pillow | 12.2.0 | 12.3.0 | #485, #503-#514 |
| gitpython | 3.1.50 | 3.1.59 | #536-#539, #554-#557, #563, #586, #587, #590, #622-#627 |
| pyasn1 | 0.6.3 | 0.6.4 | #534, #535, #568 |
| aiohttp | 3.14.1 | 3.14.3 | #594, #595, #596 |
| cryptography | 48.0.1 | 50.0.0 | #597 |

Every floor is the highest fix across all the advisories covering that package, not the first patched version of any single one. gitpython alone carries sixteen alerts across seven advisories; landing on 3.1.59 clears all of them in one move rather than climbing them one release at a time.

## The wrapt cap comes off in the same pass

`wrapt` was pinned `<2` with a comment that named its own exit condition:

> Pin <2 — wrapt 2.x is RC-only as of 2026-05 and its type stubs break pyright on `@scenario.cache()`-decorated example agents (decorated callables surface as accepting 0 positional args). Revert when wrapt 2.x stabilises and the stubs are correct.

Half of that is now true and half is not. wrapt 2.x is stable (2.4.0). The stubs still do exactly what the comment describes: lifting the cap alone leaves four pyright errors, all of the form `Expected 1 positional argument`, in `examples/test_travel_agent.py`, `examples/test_vegetarian_recipe_agent.py`, `examples/test_weather_agent.py` and `examples/test_weather_agent__gemini.py`.

So the fix is on this side rather than waiting on upstream. `scenario_cache` returns a `wrapt.decorator` proxy, which at runtime forwards the wrapped callable's signature; it is only the static type that is wrong. Typing the return as `Callable[[WrappedCallable], WrappedCallable]` states what the proxy actually is, and pyright is clean on wrapt 2.4.0 with it. No runtime behaviour changes.

### Why lift it at all

`langwatch-scenario` capping `wrapt<2` is what makes `unstructured>=0.24.0` unresolvable in the LangWatch python SDK. Every `unstructured` release from 0.23.2 up requires `wrapt>=2.1.1,<3`, and uv refuses the pair outright:

```
And because unstructured==0.25.2 depends on wrapt>=2.1.1,<3.0.0 and
langwatch-scenario==1.3.0 depends on wrapt>=1.17.2,<2, we can conclude
that langwatch:examples and langwatch-scenario==1.3.0 are incompatible.
```

That is a CRITICAL advisory with a published fix that cannot be taken while the two are jointly unsatisfiable. A release carrying this change unblocks it.

## Verified

- `pyright .`: **0 errors** on wrapt 2.4.0 (4 errors with the cap lifted and no typing fix, 0 errors on the pinned baseline, so the change is a true equivalent rather than a suppression)
- `pytest tests/ -m "not integration"`: **1494 passed**, 10 skipped, 36 deselected, 0 failed
- version-set diff of the whole lock: seven packages moved, none downgraded, none removed
- closure re-checked against each advisory's `vulnerableVersionRange` rather than its `firstPatchedVersion`: **0 of 38 left open**

https://claude.ai/code/session_017SsqT8ApEch8zUo7Aqc3y7


langwatch-agent:
CI green: 12 checks passing, 0 failing, no failed workflow runs. `python-ci` (`test (3.12)`) passed, which is the job that exercises the changed lock. Ready for human review.

langwatch-agent:
Rebased and extended to close **every** open alert on `python/uv.lock` (38 of 38, up from 28) now that gitpython 3.1.58 has aged past the release gate. Five packages move, nothing moves down. Baselined against main: identical 39-test failure set, all credential-dependent. Ready for human review.

langwatch-agent:
Rebased onto current main. The only lockfile movement was `langwatch-scenario` itself going 1.1.1 to 1.2.0, which is main's own release commit; no dependency changed version.

Verified after the rebase: `uv lock` exit 0 and `uv run --with pandas pytest --collect-only` collects 1391 tests with no import errors, which covers main's new judge module on the bumped stack.

Still closes all 38 alerts on `python/uv.lock`. Together with #873 that is the entire scenario inbox, 148 of 148. Ready for human review.

langwatch-agent:
Rebased onto current main. `uv lock --check` clean, closure still 38 of 38.

Baselined the suite against a fresh `origin/main` checkout rather than trusting yesterday's numbers, and the two things that actually matter both match exactly:

- **failing-test id set**: identical, 41 ids either side
- **error-type profile**: identical, category for category (31 `Required env var(s) missing: OPENAI_API_KEY`, 4 `openai.OpenAIError`, 2 `aiohttp.TransferEncodingError`, and so on)

Counts matching is not enough on its own, so those are the checks worth reporting: nothing in this branch introduces a new failure or moves an existing one to a different error type.

**On the red `python-complete`:** it is not from this branch. `python-ci` has failed on every `main` run since 2026-08-20T06:30, and the set of failing example tests on main swings run to run:

| main run | example tests failed |
|---|---|
| `3dd5150e` | 1 |
| `beabe5ef` | 26 |
| `a721af95` | 20 |
| `a3f40b4d` | 1 |

This PR's run failed exactly one, `examples/test_running_in_parallel.py::test_user_is_hungry`, on an LLM judge deciding the assistant never handed over the recipe. That is inside main's own variance band and better than most of it. The `PydanticSerializationUnexpectedValue` lines above it are warnings, not the failure.

Same root area as langwatch/scenario#882, where the javascript side turned out to be an outright rejected `OPENAI_API_KEY`. Worth checking whether the python job draws on the same secret, because a rotated key would explain both the flakiness and the size of those 20-to-26-test runs.

The five packages this PR moves are `aiohttp`, `cryptography`, `gitpython`, `pillow` and `pyasn1`. None of them is on the path between a prompt and a judge's verdict.

langwatch-agent:
Rebased onto current main and re-verified the closure claim.

All 38 `python/uv.lock` alerts close against this branch's lock, checked by interval-testing each resolved version against the alert's own `vulnerableVersionRange`.

The lockfile diff is large but the actual version movement is five floors:

```
aiohttp:       3.14.1  -> 3.14.3
cryptography:  48.0.1  -> 50.0.0
gitpython:     3.1.50  -> 3.1.59
pillow:        12.2.0  -> 12.3.0
pyasn1:        0.6.3   -> 0.6.4
```

No package is added, removed or downgraded; the package name set is identical to main at 175 entries. `pillow` is the heavy lifter, as it usually is on these locks.

Worth noting for anyone re-running this check: uv normalises package names in the lock, so a naive lookup of the names Dependabot reports (`Pillow`, `GitPython`) finds nothing and reads as "package removed" when the package is simply present under its normalised name. The numbers above are after normalising.

Together with #873, which covers the 90 JS alerts, this accounts for all 128 open alerts on this repo.

Ready for human review.

langwatch-agent:
Green now: 3/3 required gates, no reds.

The earlier `python-complete / test (3.12)` red is worth recording rather than just re-running away, because it looked like it could have been real. The unit suite passed clean (1364 passed, 0 failed); the failure was the later `Test (Examples)` step, where `examples/test_lovable_clone.py::test_lovable_clone` hit the 300s `--timeout=300` cap. On the last green run of that job it took about 61 seconds, so a 5x blowup is not the kind of thing to wave through as noise.

What ruled it out as a real break, before re-running:

- `LovableAgent.clone_template()` does `shutil.copytree` and then deletes `node_modules`. There is no install and no build, so the template's `package.json` and lockfiles are inert for this test. The agent only reads and edits `src/index.css` and `Index.tsx`.
- The identical failure hit #857, which touches only `python/uv.lock` and no template file at all.
- The test is a live multi-agent loop (agent, user simulator, judge, `max_turns=5`) and is already marked `@pytest.mark.flaky(reruns=2)`.
- The unrelated `release-please` python branch failed the same `Test (Examples)` step on 2026-08-22, on a different test, with an LLM-judge `AssertionError: assert False`. So the step flakes across branches, not just here.

Re-run confirms it: 30 passed, 1 skipped, with `test_lovable_clone` passing in 274s and 290s respectively, both just under the cap. One of the two needed an internal rerun, which is the flaky marker doing its job. The cap is tight enough for this test that a slow provider day tips it over, which is a CI robustness point rather than anything about this PR.

No code changed between the failing and passing runs.

langwatch-agent:
## Human Review Brief

> [!CAUTION]
> Merging this raises five transitive floors in `constraint-dependencies` and relocks. `cryptography>=50.0.0` is the one to look at: it is a major line and it sits underneath anything doing TLS, signing or key handling in this package's tree. The security case is strong and the risk is not the advisories, it is a major bump landing as a transitive constraint where nobody is testing it directly.

| | |
|---|---|
| **Issue** | None linked. Clears 38 of 38 open alerts on `python/uv.lock` |
| **Size** | +283/-280 across 2 files, one of them the lock |
| **Mode** | Fast skim |
| **Blast radius** | Every transitive dependency resolution in the Python package |

<details>
<summary><b>What you are looking at, if this repo's Python dependency surface is not your area</b></summary>

The Python SDK resolves with `uv`, and `python/uv.lock` is the resolved graph. GitHub raises advisories against that lockfile, and almost none of them are packages `scenario` depends on directly: `pillow`, `gitpython`, `pyasn1`, `aiohttp` and `cryptography` arrive transitively through the model clients, the observability stack and the example dependencies.

That is why the fix goes in `constraint-dependencies` rather than in the dependency list. A constraint says "whatever resolves this package, it must land at or above this version" without adding it as a dependency of the project. It is the Python equivalent of an npm `overrides` block, and this repo already carries a run of them (`python-liquid`, `filelock`, `virtualenv`) for the same reason.

Two properties of that mechanism shape the review.

- **A constraint is a floor, not a pin.** It closes the advisory window and keeps letting the resolver move forward, which is what you want for a security bump and is why `>=` is the right operator here rather than `==`.
- **It is enforced only at resolve time.** A constraint that no dependency graph reaches quietly does nothing, and nothing reports that. So the evidence that a constraint worked is the lockfile diff, not the constraint line.

Each added line carries the advisory numbers it closes as a trailing comment, which is what makes the block auditable later: a constraint with no stated reason is one nobody will ever dare remove.

`cryptography>=50.0.0` is the entry to look at hardest. It is the only one crossing a major version, and it sits under anything doing TLS or signing.

</details>

<details>
<summary><b>What it clears</b></summary>

| floor | alerts | what they cover |
|---|---|---|
| `pillow>=12.3.0` | 13 | heap overflows and out-of-bounds reads in the DDS, BLP, ICNS, PSD and JPEG2000 decoders |
| `gitpython>=3.1.58` | 15 | the GitPython advisory cluster |
| `aiohttp>=3.14.3` | 3 | |
| `pyasn1>=0.6.4` | 2 | |
| `cryptography>=50.0.0` | 1 | |

28 HIGH and 10 MODERATE, and the description says 0 remain.
</details>

<details>
<summary><b>The verification method is the part worth trusting</b></summary>

Closure was checked by interval-testing every resolved version in the lock against each advisory's `vulnerableVersionRange`, rather than comparing against `firstPatchedVersion`. That distinction matters: `firstPatchedVersion` names one line, so a fix backported to a parallel major gets miscounted by the naive check, in both directions. Doing it by range is the correct method and it is why the 38 of 38 claim is worth believing without re-deriving it.

Also noted: the `gitpython` floor was previously held at 3.1.55 because 3.1.58 was inside the release-age gate, and it has since aged past. That is the gate working rather than being worked around.
</details>

<details>
<summary><b>What to check</b></summary>

Whether the test suite actually exercises anything under `cryptography` and `aiohttp` at their new versions, or whether they are present-but-untouched. If untouched, the risk of the bump is close to zero and this is a straightforward merge. If a code path does use them, run that path once.

`pillow` at 12.x is worth the same question for anything doing image handling in the examples.
</details>

> [!NOTE]
> Advisory only. The hound does not review code line by line, does not approve and does not merge. Pairs with #873; together they take this repo's alert inbox to zero.

langwatch-agent:
This PR's own verification is complete and green; the red check is the shared gateway outage that is hitting every PR in this repo today.

**What is red.** `test (3.12)` fails in the examples suite, 27 of 30 tests, all with the same refusal:

```
RuntimeError: [UserSimulatorAgent] litellm.RateLimitError: RateLimitError:
  OpenAIException - Your organization's AI gateway access is exhausted.
  Contact your LangWatch admin.
```

The gate names five of them as though they were real defects. They are not. `classify-examples-failures.py` only recognises HTTP 402 `budget_exceeded`, and a call that leaves through litellm carries the gateway's sentence instead, so the marker never matches. #873 adds that marker; on its run, these same five reclassified as budget refusals, which is the direct evidence that none of them is a defect.

**With that fix in place this check would be green here.** The gate's rule is that an outage is only tolerated for a PR that changes nothing the suite covers, and this one changes three files, none under `python/examples/`:

```
python/pyproject.toml
python/scenario/cache.py
python/uv.lock
```

So `EXAMPLES_TOUCHED` is false and the outage branch would warn and pass. It stays red only because the classifier fix lives on #873, which has not merged. Either that merging or the daily budget window resetting will clear it.

**What is verified.** Re-ran everything today on the rebased branch:

- `uv sync --all-extras` clean
- `uv run pyright .` clean, which is the check the `wrapt<2` lift actually needed: wrapt 2.x's stubs make every `@scenario.cache()`-decorated callable look like it takes no positional arguments, and typing `scenario_cache`'s return as the identity on the callable it decorates is what keeps that quiet
- `uv run pytest tests/ -m "not integration"`: **1494 passed**, 10 skipped
- `uv run pytest examples/ --collect-only`: 32 tests collect, so nothing is broken at import
- all 7 resolved versions this change moves audited against every published advisory for their package: zero vulnerable resolutions
- closure re-checked: **38 of 38** alerts on `python/uv.lock` closed

PR #859, opened 2026-07-29
fix(python): support public tracer providers

telyai:
Scenario currently recognizes only the SDK `TracerProvider` (or one delegated SDK provider). OpenTelemetry’s public provider API allows other implementations to expose `add_span_processor`; Temporal 1.30 uses that shape for its replay-safe provider. Detect the public processor hook before proxy unwrapping so Scenario attaches its collector/exporters to the existing provider instead of attempting to replace the global provider. Includes a focused regression test.

### DIFF
--- a/python/scenario/_tracing/setup.py
+++ b/python/scenario/_tracing/setup.py
@@ -10,7 +10,7 @@ Supports:
 
 import logging
 import os
-from typing import List, Optional, Sequence
+from typing import List, Optional, Protocol, Sequence, cast
 
 from opentelemetry import trace
 from opentelemetry.sdk.trace import TracerProvider, SpanProcessor
@@ -23,6 +23,10 @@ from .filtering_exporter import FilteringSpanExporter
 logger = logging.getLogger("scenario.tracing")
 
 _initialized = False
+
+
+class _SpanProcessorProvider(Protocol):
+    def add_span_processor(self, span_processor: SpanProcessor) -> None: ...
 
 
 def setup_scenario_tracing(
@@ -88,7 +92,7 @@ def ensure_tracing_initialized(observability: Optional[dict] = None) -> None:
     _initialized = True
 
 
-def _get_concrete_provider(provider) -> Optional[TracerProvider]:
+def _get_concrete_provider(provider) -> Optional[_SpanProcessorProvider]:
     """Returns the concrete TracerProvider if one exists.
 
     Checks the provider itself and one level of delegation
@@ -96,6 +100,12 @@ def _get_concrete_provider(provider) -> Optional[TracerProvider]:
     """
     if isinstance(provider, TracerProvider):
         return provider
+
+    # OpenTelemetry's public provider interface does not require an SDK
+    # TracerProvider subclass. Providers such as Temporal's replay-safe wrapper
+    # expose the processor hook directly and are safe to configure in place.
+    if callable(getattr(provider, "add_span_processor", None)):
+        return cast(_SpanProcessorProvider, provider)
 
     # Check delegation pattern
     delegate = None
@@ -122,16 +132,17 @@ def _do_setup(
     concrete = _get_concrete_provider(existing_provider)
 
     if concrete is not None:
-        _attach_to_existing(concrete, span_filter, span_processors, trace_exporter)
+        _attach_to_existing(concrete, span_filter, span_processors, trace_exporter, instrumentors)
     else:
         _full_setup(span_filter, span_processors, trace_exporter, instrumentors)
 
 
 def _attach_to_existing(
-    provider: TracerProvider,
+    provider: _SpanProcessorProvider,
     span_filter: Optional[SpanFilter],
     span_processors: Optional[List[SpanProcessor]],
     trace_exporter: Optional[SpanExporter],
+    instrumentors: Optional[Sequence],
 ) -> None:
     """Attach processors to an existing TracerProvider."""
     provider.add_span_processor(judge_span_collector)
@@ -149,6 +160,10 @@ def _attach_to_existing(
 
     # Add LangWatch exporter
     _add_langwatch_exporter(provider, span_filter)
+
+    if instrumentors:
+        for instrumentor in instrumentors:
+            instrumentor.instrument(tracer_provider=provider)
 
 
 def _full_setup(
--- a/python/tests/test_tracing_setup.py
+++ b/python/tests/test_tracing_setup.py
@@ -8,6 +8,7 @@ from scenario._tracing.setup import (
     setup_scenario_tracing,
     ensure_tracing_initialized,
     _get_concrete_provider,
+    _do_setup,
     _reset_tracing_for_tests,
 )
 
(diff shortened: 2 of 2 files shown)


langwatch-agent (commented):
Static review found no actionable correctness, security, or tracing-ownership issue in the current fork change. The public-provider capability check remains restricted to attaching a span processor and instrumentors receive the selected provider. I did not execute the untrusted head or its scripts.

langwatch-agent (commented):
Static review of external contribution at `5ca69c7dd9c445e87b6d4ab299320f11c7aad3bc` (no PR-branch code or dependencies executed). No merge-blocking issue found. The structural provider check accepts public tracer providers only when they expose the needed processor-registration method, preserves the no-provider path, and passes the selected provider through to available instrumentors. CI evidence was considered but not rerun under the external-contribution trust boundary.

langwatch-agent (commented):
Clean static review. The duck-typed provider boundary is constrained to the processor operation used here, and the added instrumentation path is covered. Residual risk: the fork is stale and has no current full CI result.\n\nLangWatch-Review: verdict=clean sha=5ca69c7dd9c445e87b6d4ab299320f11c7aad3bc p0=0 p1=0 p2=0 p3=0

langwatch-agent (commented):

LangWatch-Review: verdict=clean sha=5ca69c7dd9c445e87b6d4ab299320f11c7aad3bc p0=0 p1=0 p2=0 p3=0

langwatch-agent:
## Human Review Brief

> [!CAUTION]
> **This attaches Scenario's exporter to a provider Scenario does not own and cannot flush.**
>
> The new `_get_concrete_provider` returns any object with a callable `add_span_processor`, and the setup then adds Scenario's processors to it. That is the right idea and it hands control of the span lifecycle to someone else's provider. Whether the run report ever reaches LangWatch now depends on that provider being shut down or force-flushed at exit, which Scenario no longer does and cannot check.
>
> The motivating case makes this concrete rather than theoretical. Temporal's replay-safe provider is replay-safe precisely because it suppresses span emission during replay. Attaching to it is correct, and it means a scenario run inside a replay produces no spans, silently and by design. Someone will hit that and read it as Scenario being broken.
>
> The other thing to check is ordering. The duck-type check runs before proxy unwrapping, and an SDK `TracerProvider` also has `add_span_processor`, so it now matches on the first branch and whatever unwrapping or delegation logic came after is skipped for every existing user. That is very likely fine, since attaching in place is what the old path was trying to reach anyway, and it is a behaviour change for everybody rather than only for the new case.

| | |
| --- | --- |
| **Mode** | `Targeted Review`. Two small functions and a test file. Read the detection order and the lifecycle question. |
| **Issue** | No linked issue. The motivating case is named in the description: Temporal 1.30's replay-safe provider. |
| **State** | +45 / -4 across 2 files. CI green, `mergeable`. Requested reviewer `rogeriochaves`. Open since 29 July. Outside contributor. |
| **Evidence** | A focused regression test with a hand-rolled `PublicProvider` exposing only `add_span_processor`, covering both the detection and the instrumentor wiring. It does not cover an SDK provider going down the new branch, which is the case that changed for everyone. |
| **Where to look** | `_get_concrete_provider` and its ordering against the old proxy path · whether anything flushes on exit · what an instrumentor does when handed a non-SDK provider |

<details>
<summary><b>What you are looking at, if scenario's tracing setup is not your area</b></summary>

Scenario emits OpenTelemetry spans, and LangWatch is one consumer of them. `python/scenario/_tracing/setup.py` is the module that gets a provider in place, and its job is more delicate than it looks because **scenario is a library running inside somebody else's application**, which may already have configured OpenTelemetry for its own reasons.

That produces the shape you see in the file:

- Initialisation is lazy by default, deferred to the first `run()`, but a user can call `setup_scenario_tracing(...)` first and take full control, in which case `run()` skips its own setup. `_initialized` is the latch.
- It detects a pre-existing provider rather than stamping over one, and `_get_concrete_provider` follows **one level of delegation** to find the real thing behind a proxy, because `opentelemetry.trace` hands out a proxy provider before a real one is set.
- Whatever else the user configures, `judge_span_collector` is always registered as a span processor. It is not optional: the judge's spans are how a run's verdict reaches LangWatch.

The type detail this PR turns on: the code has been checking `isinstance(provider, TracerProvider)` against the **SDK's concrete class**, `opentelemetry.sdk.trace.TracerProvider`. That is a narrower question than the one it actually needs to ask. All it does with the result is call `add_span_processor`, and plenty of legitimate providers offer that method without being instances of that particular class, including wrappers and vendor providers that applications install. A concrete-class check on something a host application supplies is a compatibility problem waiting to happen, and the fix direction is to ask for the capability instead.

</details>

<details>
<summary><b>What is being fixed</b></summary>

Scenario needs somewhere to send its spans. Previously it recognised only the OpenTelemetry SDK `TracerProvider`, or one delegated SDK provider behind a proxy. Anything else and it would try to replace the global provider, which stomps on whatever the host application had set up.

But OpenTelemetry's public provider API does not require an SDK subclass. Other implementations expose `add_span_processor` and are perfectly safe to configure in place. Temporal's is one. So the fix duck-types on the hook instead of the type, and attaches rather than replaces.

That is the right instinct: attaching to what is there beats replacing what someone else configured.

</details>

<details>
<summary><b>Worth asking the author</b></summary>

**Who flushes at the end of a run?** Scenario's report is only useful if the spans arrive. With an SDK provider there is a known shutdown path. With a foreign provider there is not, and a run that ends by exiting the process may lose everything buffered. If Scenario keeps a handle on its own `BatchSpanProcessor`, force-flushing that at the end of a run is cheap and closes it regardless of whose provider it lives on.

**Does an instrumentor accept a non-SDK provider?** `instrumentor.instrument(tracer_provider=provider)` is called with whatever came back. Instrumentors vary in how much they assume about the object they are given.

**Should the SDK check come first?** Putting the `isinstance` check ahead of the duck-type check would keep the existing path byte-for-byte identical for existing users and confine the change to the case it is for. If the new ordering is deliberate because attaching in place is better even for SDK providers, that is worth saying, since it is the part of the change that touches everybody.

**Is the Temporal replay case documented anywhere?** No spans during replay is correct behaviour that will read as a bug. One line in the docs saves someone a day.

</details>

> [!NOTE]
> **Attaching to an existing provider is the right move; owning nothing about its lifecycle is the cost.** The review question is what guarantees the spans get out.

langwatch-agent:
> [!NOTE]

<details open>
<summary><b>What you are looking at, if the Python SDK is not your directory</b></summary>

The Python SDK instruments a user's application and ships spans to LangWatch over OpenTelemetry. Two facts about that shape explain most of its bugs.

**It runs inside somebody else's process.** The SDK does not own the OTel setup: the host application may already have a tracer provider, may configure its own exporters, may be a framework that sets things up before our code runs. So the SDK constantly has to decide between using what is there and installing its own, and getting that wrong either drops our spans or hijacks the host's telemetry.

**It caches aggressively**, because tracing sits on a hot path and re-deriving the same values per call is not affordable. Cache keys are therefore derived from call arguments, which is exactly where subtle correctness bugs come from: two calls that differ only in a way the key ignores collapse into one entry.

Both patterns produce silent wrongness rather than errors. A dropped span or a wrong cache hit looks like nothing at all until someone notices missing data days later.
</details>

Issue #209, opened 2026-01-12
Refactor: Replace magic boolean parameters with named options


drewdrewthis:
## Problem

In `src/execution/scenario-execution.ts`, there are boolean parameters that make code hard to read at call sites:

```typescript
// Line 431
private async _step(
  goToNextTurn: boolean = true,
  onTurn?: (state: ScenarioExecutionStateLike) => void | Promise<void>
): Promise<void>

// Line 508
private async callAgent(
  idx: number,
  role: AgentRole,
  judgmentRequest: boolean = false
): Promise<void>
```

Boolean parameters force readers to remember what `true` or `false` means at each call site.

## Proposed Solution

Use named options objects or discriminated function names:

```typescript
interface StepOptions {
  advanceTurnIfNeeded: boolean;
  onTurn?: (state: ScenarioExecutionStateLike) => void | Promise<void>;
}

private async _step(options: StepOptions): Promise<void>

// Or split into explicit methods:
private async stepWithTurnAdvance(): Promise<void>
private async stepWithoutTurnAdvance(): Promise<void>
```

## Files Affected
- `src/execution/scenario-execution.ts`

langwatch-agent:
Premise re-checked in today's tech-debt run. **The harm this ticket describes cannot occur any more**, because neither parameter is passed as a boolean literal at any call site.

The ticket's argument is:

> Boolean parameters force readers to remember what `true` or `false` means at each call site.

Every call site on current `main`, in `javascript/src/execution/scenario-execution.ts`:

```
908:    await this._step();
949:      return this._step(goToNextTurn, onTurn);
1608:      await this._step(goToNextTurn, onTurn);

961:    await this.callAgent(index, currentRole);
1835:      promise: this.callAgent(idx, AgentRole.AGENT)
2352:    await this.callAgent(index, role, judgmentRequest);
```

Not one passes `true` or `false`. `_step` is either called with no arguments, taking the default, or with a named `goToNextTurn` variable that says what it is at the point of the call.

**And half the ticket is already fixed.** `judgmentRequest` is no longer a boolean at all:

```ts
judgmentRequest?: JudgmentRequest
```

It carries the criteria now (`judgmentRequest?.criteria`), so the flag became a value with a type, which is the stronger version of what this ticket asks for.

What is left is the signature reading `goToNextTurn: boolean = true` in isolation. That is a smaller complaint than the one filed, and changing it now would rewrite three call sites that already read clearly. Skipping rather than churning them.

PR #860, opened 2026-07-30
refactor(execution): consolidate the two next-agent-for-role lookups (#210)

### LINKED ISSUE #210: Refactor: Consolidate duplicate nextAgentForRole methods
## Problem

In `src/execution/scenario-execution.ts`, there are two nearly identical methods:

```typescript
// Lines 1074-1089
private nextAgentForRole(role: AgentRole): { idx: number; agent: AgentAdapter | null; } {
  for (const agent of this.agents) {
    if (agent.role === role && this.pendingAgentsOnTurn.has(agent) && this.pendingRolesOnTurn.includes(role)) {
      return { idx: this.agents.indexOf(agent), agent };
    }
  }
  return { idx: -1, agent: null };
}

// Lines 1151-1161
private getNextAgentForRole(role: AgentRole): { index: number; agent: AgentAdapter } | null {
  for (let i = 0; i < this.agents.length; i++) {
    const agent = this.agents[i];
    if (agent.role === role && this.pendingAgentsOnTurn.has(agent)) {
      return { index: i, agent };
    }
  }
  return null;
}
```

These do almost the same thing with slightly different return types. This violates DRY.

## Proposed Solution

Consolidate into one method with consistent naming:

```typescript
private findNextAgentForRole(role: AgentRole): { index: number; agent: AgentAdapter } | null {
  for (let i = 0; i < this.agents.length; i++) {
    const agent = this.agents[i];
    if (agent.role === role && this.pendingAgentsOnTurn.has(agent)) {
      return { index: i, agent };
    }
  }
  return null;
}
```

## Files Affected
- `src/execution/scenario-execution.ts`

langwatch-agent:
## Ticket

Closes #210. `nextAgentForRole` and `getNextAgentForRole` in `scenario-execution.ts` answered the same question with different return shapes. Now one `findNextAgentForRole` returning `{ index, agent } | null`, used by all five call sites.

## The ticket's premise needed one correction

The issue calls them "nearly identical" and proposes consolidating on a body that keeps `pendingRolesOnTurn.includes(role)`. Consolidating that way would have changed behaviour, because the condition is exactly what the two did not share, and its absence on the scripted side is load-bearing.

**Dropping it is the behaviour-preserving direction.** The condition cannot be false at any of the three call sites that had it:

- `_step` (`:921`) passes `pendingRolesOnTurn[0]`, so `includes` asks whether an array contains its own head. On an empty queue the role is `undefined` and no agent matches on `agent.role` either way.
- The inline-barge lookup (`:1751`) iterates `for (const r of pendingRolesOnTurn)`, so `r` is in the array by construction.
- Its second call (`:1759`) asks for `AGENT` only after `nextRole` was assigned from that same iteration, and nothing mutates the array in between.

**Keeping it would have changed the scripted path.** `scriptCallAgent` runs `consumeUntilRole(role)` first, which drains `pendingRolesOnTurn` entirely when the role is not in it; the agent has to remain reachable after that. Adding the condition there makes the first lookup miss and takes the `newTurn()` retry branch.

## A parity finding, pinned rather than fixed

Python's `_next_agent_for_role` (`python/scenario/scenario_executor.py:538`) is a single method that **does** keep `_pending_roles_on_turn`, and its `script_call_agent` (`:1673`) uses it. So Python takes the `newTurn()` that TypeScript skips, and the two languages differ on turn counting in this case.

I have not aligned them. It is a behaviour change with `maxTurns` consequences, and which side is right is a decision for whoever owns the parity question, not something to smuggle into a DRY refactor. This PR pins today's TypeScript value so the choice becomes a visible failing test rather than a silent drift.

## Evidence

**Before:** 92 files, 1079 tests, green. **After:** 93 files, 1081 tests, green. `pnpm typecheck` and `pnpm lint:lib` clean.

**Nothing observed any of this before.** Restoring `pendingRolesOnTurn.includes(role)` for all call sites, which is the naive consolidation the issue proposes:

```
pnpm test
Test Files  93 passed | 1 skipped (94)
      Tests  1081 passed | 4 skipped (1085)
```

All green, while `currentTurn` after a scripted `agent()` then `user()` moves from `0` to `1`. That is the entire reason this PR ships a test as well as a refactor.

**The new test discriminates.** `script-call-role-pending.test.ts` reaches the state through the public API: `agent()` runs `consumeUntilRole(AGENT)`, which drains USER off `pendingRolesOnTurn` while leaving the user simulator on `pendingAgentsOnTurn`; the following `user()` then asks for a role the turn no longer lists.

| Source state | `script-call-role-pending.test.ts` | Full JS suite |
|---|---|---|
| this PR | 2 passed | 1081 passed |
| condition restored (Python semantics) | 1 failed / 1 passed | 1081 passed |

The suite cannot tell the two apart; the new test can.

**Also checked:** the old `nextAgentForRole` used `this.agents.indexOf(agent)` rather than the loop position. Equivalent here, and not a behaviour change: it returns the first occurrence, and the first occurrence is what the loop reaches first, since `pendingAgentsOnTurn` is a `Set` of instances and so cannot distinguish two copies of one adapter anyway.

## Notes

The single parameter stays positional. A `requireRolePending` flag was the first shape I tried, and the tautology above makes it dead weight; it would also have been exactly the magic boolean parameter #209 asks the repo to stop adding.

## Advisory note

### DIFF
--- a/javascript/src/execution/__tests__/script-call-role-pending.test.ts
+++ b/javascript/src/execution/__tests__/script-call-role-pending.test.ts
@@ -1,0 +1,95 @@
+/**
+ * Pins the one behavioural difference between the two agent lookups that #210
+ * consolidated into `findNextAgentForRole`.
+ *
+ * The automatic loop requires the role to still be pending on the turn; a
+ * scripted call does not. Before this test nothing observed that: flipping the
+ * scripted path to the automatic path's semantics left all 1079 other JS tests
+ * green, while silently adding a `newTurn()` and so changing turn counting.
+ *
+ * Reaching the case takes two scripted steps in one turn. `agent()` runs
+ * `consumeUntilRole(AGENT)`, which drains USER off `pendingRolesOnTurn` while
+ * leaving the user simulator on `pendingAgentsOnTurn`. The following `user()`
+ * then asks for a role the turn no longer lists, but whose agent is still
+ * available.
+ *
+ * Python's `_next_agent_for_role` always applies the guard, so the value
+ * asserted here is a TypeScript-only behaviour. It is pinned rather than
+ * aligned: aligning it is a turn-count change and belongs to whoever decides
+ * the parity question on #210, with this test failing to mark the decision.
+ */
+
+import { describe, expect, it } from "vitest";
+
+import {
+  AgentAdapter,
+  type AgentInput,
+  type AgentReturnTypes,
+  AgentRole,
+  UserSimulatorAgentAdapter,
+} from "../../domain";
+import { ScenarioExecution } from "../scenario-execution";
+
+class MockAgent extends AgentAdapter {
+  role = AgentRole.AGENT;
+  calls = 0;
+  async call(_input: AgentInput): Promise<AgentReturnTypes> {
+    this.calls += 1;
+    return { role: "assistant" as const, content: "agent reply" };
+  }
+}
+
+class CountingUserSim extends UserSimulatorAgentAdapter {
+  role = AgentRole.USER;
+  calls = 0;
+  async call(_input: AgentInput): Promise<AgentReturnTypes> {
+    this.calls += 1;
+    return `user turn ${this.calls}`;
+  }
+}
+
+async function runAgentThenUser() {
+  const sim = new CountingUserSim();
+  const agent = new MockAgent();
+  const turnsSeen: number[] = [];
+
+  const exec = new ScenarioExecution(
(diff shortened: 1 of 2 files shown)


langwatch-agent (commented):
No findings. The consolidation preserves the distinct scripted-call behavior and the regression test covers the previously unobserved turn-count path.

rogeriochaves (approved):

PR #685, opened 2026-06-18
fix(security): bump python/uv.lock security floors (cryptography, python-multipart, starlette, python-liquid, pydantic-settings)

langwatch-agent:
## What

Surgical `uv lock` bump in `python/uv.lock`: cryptography 48.0.0 -> 48.0.1, python-multipart 0.0.29 -> 0.0.32.

## Why

- **cryptography <48.0.1** (HIGH, #456)
- **python-multipart <0.0.30** (HIGH #460 + LOW #457/#458/#459)

Resolves #456, #457, #458, #459, #460.

## Notes

- cryptography is capped to `<49` so it lands on the patched **48.0.1** instead of the 6-day-old **49.0.0** major release (minimal churn / avoid a bleeding-edge major bump).
- Only these two packages change in the lock; `uv lock --check` passes.
- **starlette is intentionally deferred** - its fix (1.3.0/1.3.1) was published 2026-06-11/12 and is still inside the 7-day release-age window; it will follow once that clears (~tomorrow).

PR #873, opened 2026-08-03
fix(security): close all 117 open JS alerts across the scenario workspaces

langwatch-agent:
## What

Closes all **117** open JavaScript alerts across the scenario workspaces (60 HIGH, 47 MODERATE, 10 LOW). With #857 (the 38 on `python/uv.lock`) that is the entire scenario inbox, 155 of 155.

| manifest | alerts | how |
|---|---|---|
| `javascript/package-lock.json` | 35 | file removed, see below |
| `docs/pnpm-lock.yaml` | 24 | floors raised in `docs/pnpm-workspace.yaml` |
| `javascript/examples/openai-realtime-demo/pnpm-lock.yaml` | 20 | orphan removed, floors moved to the workspace root |
| `python/examples/lovable_clone/template/pnpm-lock.yaml` | 13 | floors + react-router 7 |
| `javascript/pnpm-lock.yaml` | 13 | floors raised in `javascript/package.json` |
| `python/examples/lovable_clone/template/package-lock.json` | 12 | floors + react-router 7 |

## Raising in place, not adding alongside

pnpm matches only the **first** selector for a package name, so adding a correctly-scoped floor next to a stale one leaves the stale one in charge. Main had exactly that shape: `docs/pnpm-workspace.yaml` carried an unscoped `hono: '>=4.12.25'`, which would have sat in front of anything new. Every stale entry is raised in place rather than shadowed.

Every selector is scoped to one major line (`"pkg@>=lo <fix": ">=fix <next"`). Two packages needed more than one because the tree carries parallel majors: `brace-expansion` (1.x, 2.x and 5.x) and `js-yaml` (3.x and 4.x). Targets are capped so a floor cannot float across a major, which was not academic here: `javascript/pnpm-lock.yaml` had resolved **fast-uri 4.1.2** off an uncapped `fast-uri@<=3.1.1 -> >=3.1.2`.

Each target is the highest fix across every advisory covering that package, not the first patched version of any single one, so a package with a chain of advisories lands above all of them in one move.

## Two lockfiles removed

**`javascript/package-lock.json`** carries the largest single share of the inbox, 35 alerts, and nothing installs from it. Every workflow uses `pnpm install --frozen-lockfile`; there is no `npm ci` or `npm install` anywhere in `.github/`; nothing in the repo references the file; and its entire commit history is release version bumps. It had drifted far enough that regenerating it moved `@openai/agents` 0.3.9 to 0.16.1 and removed 53 packages, and it cannot be regenerated at all without `--legacy-peer-deps` (a pre-existing zod peer conflict that reproduces on main untouched). Committing that would be shipping a lock nothing validates. The `pnpm-lock.yaml` beside it resolves the same packages at safe versions.

**`javascript/examples/openai-realtime-demo/pnpm-lock.yaml`** is an orphan. The directory is a declared member of the javascript workspace, so pnpm ignores its `pnpm.overrides` and writes the parent lock instead. Six of its seven floors were already superseded at the root; the seventh, `ajv`, moves there. Checked before removing.

## react-router 7 in the lovable_clone template

`react-router-dom` has no patched release in the 6.x line, so leaving 6.x was the only way to close it. The template uses only the v7-compatible surface (`BrowserRouter`, `Routes`, `Route` in `src/App.tsx`; `useLocation` in `src/pages/NotFound.tsx`), so no application code changed. `@remix-run/router` disappears from both lockfiles because v7 folded it in, which is why those alerts close by the package no longer existing.

`postcss` is a direct devDependency there and npm rejects an override that conflicts with a direct dependency, so its floor is carried by the direct pin.

## Two bumps that are not security fixes

**`@vitejs/plugin-react-swc` 3.11.0 to 4.3.3** in the lovable_clone template. 3.11.0 peers `vite ^4 || ^5 || ^6 || ^7` while the template is already on vite 8, so npm refused to resolve the tree with `ERESOLVE` and `package-lock.json` could not be regenerated at all. 4.3.3 peers `^8` as well. The plugin is used as a bare `react()` in `vite.config.ts`, unchanged, and the template builds.

**`toml` 3.0.0 to the 4.x line** in docs. 4.x is the only line with a fix, so this one is a major bump by necessity. The two are API-identical for this use: both export `parse` and nothing else, and both return the same object for a document with scalars and a table (checked directly against each published tarball). The consumer is `remark-mdx-frontmatter`, which parses page frontmatter, and the docs site prerenders every page unchanged on 4.3.0.

## The fast-uri floor is 3.1.6, not 3.1.5

Worth calling out because the alerts on this repo cannot tell you that on their own.

`javascript/pnpm-lock.yaml` resolves the 4.x line, so the alerts raised against it (#682-#685) all name `>= 4.0.0, < 4.1.3`. Reading the floor off those and capping to 3.x lands on **3.1.5**, which four separate HIGH advisories still cover: `>= 3.0.0, < 3.1.6` and `>= 3.1.2, < 3.1.6`, both fixed in 3.1.6. Those advisories are live and visible in this same inbox, raised as #686, #687 and #691-#694 against the two locks this PR deletes.

The floor is therefore `>=3.1.6 <4`, and the workspace resolves 3.1.7. Every resolved version this PR moves was then re-checked against every published advisory for its package rather than only the ones raised here, which is the check that caught this.

## Verified

- closure re-checked programmatically against each advisory's `vulnerableVersionRange`, not its `firstPatchedVersion`, over every resolved version in all four remaining locks: **0 of 117 left open**
- `pnpm install --frozen-lockfile` exit 0 in `javascript/`, `docs/` and the template
- `docs`: `eslint .` clean, `pnpm run build` exit 0, every page prerendered on toml 4.3.0
- template: `pnpm run build` exit 0 on vite 8 with plugin-react-swc 4.3.3, browserslist 4.28.9, postcss-selector-parser 6.1.4
- template: `npm install --package-lock-only` exit 0 with no peer conflict, which it could not do before the plugin bump
- template routing driven in a real browser on the built output: `/` renders, `/definitely-not-a-route` renders the 404 with its intentional `console.error` carrying the pathname (so `useLocation()` resolves), the "Return to Home" link client-side navigates, and browser-back returns to the 404
- lockfile version-set diffs reviewed per file; no package moves down

## Pre-existing failures, not from this PR

`pnpm typecheck` in `javascript/` fails with `TS2688: Cannot find type definition file for 'yauzl'`. Confirmed identical on a clean checkout of `origin/main` at the same commit this branch sits on, so it predates this change and is not caused by it.

https://claude.ai/code/session_017SsqT8ApEch8zUo7Aqc3y7


langwatch-agent:
Rebased onto main and substantially expanded: floors raised to current, now **105 of the 110** open JS alerts instead of 37. Two structural findings are written up in the description (an entire example's overrides were inert because it is a workspace member, and `javascript/` keeps two lockfiles whose override blocks had drifted). Dogfooded: 1083 tests pass, docs builds, 53 packages up with zero downgrades. CI green. Ready for human review.

langwatch-agent:
Added the react-router 7 migration for the `lovable_clone` template, which closes the last 5 alerts that were still open on it.

**This PR now closes 110 of the 110 open JS alerts in scenario**, up from 105. Nothing on the JS side is left behind after it merges.

I had previously written these five off as "a v6 to v7 migration, which is a different kind of change from a floor bump". That was too quick a judgement. The template only ever used the v7-compatible subset of the API:

- `BrowserRouter`, `Routes`, `Route` in `src/App.tsx`
- `useLocation` in `src/pages/NotFound.tsx`

All four are still exported by `react-router-dom` 7, so the migration needed **zero code changes** — only the version and the two overrides that pinned the 6.x line.

`@remix-run/router` drops out of both lockfiles entirely, because v7 folded it into `react-router`. That retires its overrides too. `cookie` and `set-cookie-parser` arrive as v7's own dependencies. Version-set audit on the pnpm lock: 2 up, 0 down, 1 removed, 2 added.

`react-router-dom` #551 is worth calling out: it has **no patched release in the 6.x line at all**, so leaving the 6.x range was the only way to close it. Staying on 6.30.4 would have kept it open permanently.

**Verified in a browser, not just on the build**, since a passing build says nothing about whether routes actually resolve:

- `npx tsc --noEmit` clean, `pnpm build` clean
- against the built preview: `/` renders "Welcome to Your Blank App"
- `/definitely-not-a-route` renders the 404 page, and its intentional `console.error` fires with the attempted path, which is what demonstrates `useLocation()` still resolves under v7

langwatch-agent:
Rebased onto current main, which added the `openai-realtime-demo` example and two release commits. `pnpm install --frozen-lockfile` in `javascript/` is exit 0 against the rebased lockfile, so nothing in main's four commits moved a dependency.

Re-ran the closure check across the whole scenario inbox afterwards. All 148 open alerts are live (no stale manifests in this repo), and this PR plus #857 close every one of them: 110 here, 38 there, 0 left over. The only two entries the checker cannot see a resolved version for are `@remix-run/router` #613 and #616, and that is because react-router 7 folded the package in and it no longer exists in either template lockfile.

Ready for human review.

langwatch-agent:
CI caught a real regression in the first push of this branch, and it is fixed in `b2c733f4`. Worth recording because the failure mode is not obvious.

`ci-checks (24.x)` failed at Lint with:

```
TypeError: expand is not a function
    at Minimatch.braceExpand (.../minimatch@3.1.5/minimatch.js:271:10)
```

The cause was my own override selector. I had written the floors as `pkg@<fix`, but a bare upper bound with no lower bound **also matches every lower major**: `brace-expansion@<5.0.7` matches 1.1.14 just as happily as 5.0.6. So `minimatch@3.1.5`, which declares `brace-expansion: ^1.1.7`, was handed brace-expansion **5.0.9**, whose export shape is not a callable, and eslint crashed before linting anything.

Main's original selectors had lower bounds (`brace-expansion@>=5.0.0 <5.0.6`, `js-yaml@>=4.0.0 <4.2.0`). I dropped that property while raising them to the current advisory versions. That was the mistake.

Every selector this PR touches now carries one, `pkg@>=N.0.0 <fix`, and the tree keeps each major line patched within itself:

```
minimatch@3.1.3 -> brace-expansion 1.1.18     (was 5.0.9)
minimatch@3.1.5 -> brace-expansion 1.1.18     (was 5.0.9)
minimatch@9.0.7 -> brace-expansion 5.0.9      (declares ^5.0.2, correct)
minimatch@10.x  -> brace-expansion 5.0.9
js-yaml: 3.15.1, 4.3.1, 5.2.2                 (3.x had vanished entirely)
```

I only added bounds to the selectors this PR introduces or raises. Pre-existing entries are untouched, deliberately: `esbuild@<0.28.1` and friends are 0.x, where the semver major is the minor, so the same rewrite would be wrong there.

Re-verified after the fix: closure still 90 of 90; `pnpm install --frozen-lockfile` exit 0 in all three workspaces; `pnpm lint:all` in `javascript/` no longer crashes and now reports exactly what a clean `origin/main` checkout reports (516 problems, 8 errors, all `import/no-unresolved` in `examples/vitest` that need the package built first); template `tsc --noEmit` and `pnpm build` exit 0, and the browser pass over the built output still goes 404 to Link to back cleanly.

Two failures remain on this branch that are **not** from it, both reproduced on an untouched `origin/main`: `pnpm typecheck` and the dts step of `pnpm build` fail with `TS2688: Cannot find type definition file for 'yauzl'`.

langwatch-agent:
Second CI finding, and this one is **not** from this branch: `javascript-complete` is red on `origin/main` too, with the identical failure. Fixed here in `d20e19db` because it otherwise keeps this PR unmergeable.

```
FAIL tests/scenario-expert-realtime.test.ts
TypeError: Cannot read properties of undefined (reading 'type')
  ❯ Module.create ../../node_modules/.pnpm/zod@3.25.76/node_modules/zod/v3/types.js:2474:68
  ❯ @openai+agents-core@0.16.0_ws@8.21.3_zod@3.25.76/src/types/protocol.ts:802:30
  ❯ ../openai-realtime-demo/agents/realtime-user-simulator.agent.ts:1:1
```

Same file, same error, same `1 failed | 34 passed | 21 skipped` on main's own run of `ci-checks (24.x)`. Note the resolution key in the path: `agents-core@0.16.0_zod@3.25.76`.

**Cause.** `@openai/agents` 0.16 declares a `zod@^4.0.0` peer. `javascript/examples/openai-realtime-demo/package.json` depends on `@openai/agents` but declares no zod of its own, so pnpm satisfied the peer from the `javascript` root, which is on `zod@^3.25.76`. The package's schema construction then fails at import, so the test dies before running an assertion. `examples/vitest` was unaffected because it already declares `zod: ^4.1.13`.

**Fix.** Declare the same `zod: ^4.1.13` in the demo. The importer moves from `0.16.0(ws@8.21.3)(zod@3.25.76)` to `0.16.0(ws@8.21.3)(zod@4.4.3)`. No package version changes anywhere in the lockfile, only that peer resolution:

```
-        version: 0.16.0(ws@8.21.3)(zod@3.25.76)
+        version: 0.16.0(ws@8.21.3)(zod@4.4.3)
```

**Verified locally.** The import-time TypeError is gone; the suite now gets into the test body and stops at `RealtimeAgentAdapter.connect requires an API key`, which is this sandbox having no `OPENAI_REALTIME_API_KEY`. CI has one, so this should now pass there.

Alert closure is unchanged at 90 of 90.

Worth flagging separately: this went red on main and stayed red. If the intent is for `javascript-complete` to gate merges, something is letting it through.

langwatch-agent:
Rebased onto current main. Two notes on state.

**Main absorbed the zod fix from this branch.** `javascript/examples/openai-realtime-demo/package.json` now carries `zod: ^4.1.13` on main, so the rebase dropped that commit as already upstream. This PR is back to a single commit, the 90-alert closure work. Re-verified after the rebase: closure still 90 of 90, and `pnpm install --frozen-lockfile` is exit 0 in `docs/`, `javascript/` and the template with no lockfile regeneration needed.

**The red on this PR is not from this PR.** `ci-checks (24.x)` fails with:

```
code: 'invalid_api_key', message: 'Incorrect API key provided: sk-proj-****GpwA'
```

`main` fails identically, with the same `5 failed | 30 passed | 21 skipped`, and has done on every run since 2026-08-20T13:03. This branch was green on the same content yesterday morning, before that started. Details and the run list are in langwatch/scenario#882; it needs the `OPENAI_API_KEY` secret rotated, which I have no access to.

So `javascript-complete` here will stay red until that secret is fixed, independently of anything in this branch.

langwatch-agent:
Rebased onto current main and re-verified the closure claim from scratch rather than trusting the title.

Against this branch's tree, all 90 JS alerts close:

```
45  closed because the resolved version is above the advisory range
43  closed because the manifest is deleted
 2  closed because the package no longer exists in the tree
```

The two in that last bucket are both `@remix-run/router` (#613, #616). That is real, not a parsing artifact: this PR moves `react-router-dom` from ^6.26.2 to ^7.18.0, and react-router 7 merged `@remix-run/router` back into `react-router`, so the package is gone rather than bumped.

Every check is an interval test of the resolved version against the alert's own `vulnerableVersionRange`, not a comparison to `firstPatchedVersion`, so an advisory with no tagged fix still gets judged correctly.

Zero of the 90 remain open. Together with #857, which covers `python/uv.lock`, that accounts for all 128 open alerts on this repo.

The `ci-checks / javascript-complete` red here is inherited from main, not caused by this PR: it is the rejected `OPENAI_API_KEY` tracked in #882. The same key is failing the langevals suite in the langwatch repo today, so it is an org-wide credential that needs rotating, which I cannot do.

Ready for human review.

langwatch-agent:
Green now: 3/3 required gates, no reds.

The earlier `python-complete / test (3.12)` red is worth recording rather than just re-running away, because it looked like it could have been real. The unit suite passed clean (1364 passed, 0 failed); the failure was the later `Test (Examples)` step, where `examples/test_lovable_clone.py::test_lovable_clone` hit the 300s `--timeout=300` cap. On the last green run of that job it took about 61 seconds, so a 5x blowup is not the kind of thing to wave through as noise.

What ruled it out as a real break, before re-running:

- `LovableAgent.clone_template()` does `shutil.copytree` and then deletes `node_modules`. There is no install and no build, so the template's `package.json` and lockfiles are inert for this test. The agent only reads and edits `src/index.css` and `Index.tsx`.
- The identical failure hit #857, which touches only `python/uv.lock` and no template file at all.
- The test is a live multi-agent loop (agent, user simulator, judge, `max_turns=5`) and is already marked `@pytest.mark.flaky(reruns=2)`.
- The unrelated `release-please` python branch failed the same `Test (Examples)` step on 2026-08-22, on a different test, with an LLM-judge `AssertionError: assert False`. So the step flakes across branches, not just here.

Re-run confirms it: 30 passed, 1 skipped, with `test_lovable_clone` passing in 274s and 290s respectively, both just under the cap. One of the two needed an internal rerun, which is the flaky marker doing its job. The cap is tight enough for this test that a slow provider day tips it over, which is a CI robustness point rather than anything about this PR.

No code changed between the failing and passing runs.

langwatch-agent:
## Human Review Brief

> [!CAUTION]
> Merging this deletes `javascript/package-lock.json` and raises dependency floors across five other manifests. Removing a committed lockfile changes what a fresh install resolves for everyone, including CI. If anything still installs through npm in that workspace, its dependency versions become whatever npm picks that day rather than what the lockfile pinned. That is the one line in this PR that can break a build for a reason unrelated to security.

| | |
|---|---|
| **Issue** | None linked. Clears 90 of the 128 open alerts on this repo; #857 clears the other 38 |
| **Size** | +510/-16268 across 10 files, the deletions being lockfiles |
| **Mode** | Targeted, on the lockfile removal |
| **Blast radius** | Dependency resolution in six manifests across the scenario workspaces |

<details>
<summary><b>What you are looking at, if this repo's dependency surface is not your area</b></summary>

`scenario` is not one JavaScript project, it is several, and that is the whole reason this diff is shaped the way it is. There is `javascript/` (the published `@langwatch/scenario` package), `docs/` (its own pnpm workspace), `javascript/examples/*` (each with its own lockfile), and even a Node template nested inside a Python example at `python/examples/lovable_clone/template/`. GitHub raises advisories per manifest, so a transitive package that is vulnerable in four lockfiles is four alerts, and closing "all 90" means visiting every one of those trees, not bumping one thing ninety times.

**The lever is overrides, not upgrades.** Almost none of these are direct dependencies. They arrive transitively through the OpenTelemetry SDK, LangChain, Hono and the rest, so waiting for each upstream to release is not a plan. `javascript/package.json` therefore carries two override blocks that do the same job for two package managers: a top-level `overrides` for npm and a `pnpm.overrides` for pnpm. Both are written as ranged rewrites (`"pkg@>=X <Y": ">=Y <Z"`) so an override applies to the vulnerable window and expires on its own instead of pinning the tree forever.

Three things follow that are worth knowing before reading 15,000 lines of lockfile.

- **An override is a claim about compatibility that nothing type-checks.** Forcing a transitive package forward is exactly as safe as that package's own semver discipline, and the consumer that pulled it in never agreed to the new version.
- **The two blocks can drift.** The same package appearing in one list and not the other is a real state, and this diff both adds entries to one and removes entries from the other.
- **Deleting a lockfile is a bigger decision than it reads as.** Removing `javascript/package-lock.json` says this tree is pnpm-only from here on. Anything that ran `npm install` against it, including the root `package.json`'s own `prepare` script, is affected.

Multiple ranged entries for one package (`brace-expansion` has three here) are normal rather than suspicious: separate major lines get separate advisories and separate fixed versions.

</details>

<details>
<summary><b>The pnpm detail that makes this non-obvious</b></summary>

The part of this PR that is genuinely worth knowing: pnpm matches only the **first** selector for a package name. So adding a correctly scoped floor next to a stale one leaves the stale one in charge, and the alert stays open while the diff looks like it fixed something. Main had exactly that shape in `docs/pnpm-workspace.yaml`.

This is why the change raises floors in place rather than appending them, and it is the kind of thing that will be re-broken by the next person who does not know it. Worth confirming the reason is written down somewhere more durable than this PR description.
</details>

<details>
<summary><b>The three questions</b></summary>

1. **Why does `javascript/package-lock.json` go away rather than get updated?** The table says "file removed, see below", which implies a considered reason, most likely that the workspace is pnpm and the npm lockfile was a leftover. Confirm nothing installs with npm there: not a CI job, not a Dockerfile, not a contributor's README instruction. If something does, this is the breakage.
2. **The react-router 7 bump in the lovable_clone template.** That is a major version with real migration notes, and it appears in two manifests. Templates get copied by users, so a template that does not run is worse than a template with an advisory. Confirm the template still builds.
3. **The orphan removal.** `openai-realtime-demo/pnpm-lock.yaml` goes away with its floors moved to the workspace root. Same question as 1, scoped smaller: confirm the demo still installs standalone if anyone runs it that way.
</details>

<details>
<summary><b>What not to spend time on</b></summary>

The 16268 deleted lines are lockfile content. Reading them proves nothing. The alert-to-manifest table in the description is the summary worth checking against the security tab, and that check takes a minute.
</details>

> [!NOTE]
> Advisory only. The hound does not review code line by line, does not approve and does not merge. Alongside #857 this closes the repo's entire alert inbox, which is worth someone's time this week.

langwatch-agent:
Rebased onto current main; this is mergeable again.

The conflict was a modify/delete on `javascript/package-lock.json`: this PR deletes it, and main's release-please commit for javascript 1.4.0 bumped the version field inside it.

Before resolving that as a delete, I re-checked the premise that the file is vestigial, since main touching it is exactly the evidence that would overturn it. It holds, and main's own history is the clearest argument:

- Every workflow installs with `pnpm install --frozen-lockfile`. There is no `npm ci` or `npm install` in `.github/workflows/` at all.
- The real dependency change in that window, moving `@elevenlabs/elevenlabs-js` to devDependencies (#945), updated `javascript/pnpm-lock.yaml` and did **not** touch `javascript/package-lock.json`. So a human changing dependencies already treats the npm lock as not-a-lockfile.
- The only thing that writes to it is release-please stamping a version, which simply stops happening once the file is gone.

That file carries 28 of the alerts this PR closes, all in dependencies nothing installs from.

`javascript/package.json` and `javascript/pnpm-lock.yaml` auto-merged, and I re-ran `pnpm install --lockfile-only` afterwards rather than trusting that: the re-resolution came back byte-identical, so the merge was sound against main's 1.4.0 and the elevenlabs move.

Closure re-verified against today's inbox on the rebased tree, unchanged at 90 of 90:

```
45  closed because the resolved version is above the advisory range
43  closed because the manifest is deleted
 2  closed because the package no longer exists in the tree
```

Ready for human review.

langwatch-agent:
Honest read of this PR's CI, and a correction to what I said yesterday.

**Yesterday's "green" was not a real green.** The heavy workflows had not run on this branch at all: the checks I read were `Conventional Commits` and the low-risk auto-approve, and `javascript-ci` / `python-ci` first ran today. I should have noticed that 18 checks was too few for this repo. Apologies for the false confidence.

**What actually ran today.** The unit suites pass: the javascript workspace is **1557 passed, 1 skipped across 122 files** in CI, matching 1548 locally. What is red is the examples suite in both `javascript-ci` and `python-ci`, and it is red because the shared AI gateway is out of budget, not because of anything in this change.

The python side says so in plain text, 27 of 30 example tests failing together:

```
FAILED examples/test_travel_agent.py::test_travel_agent - RuntimeError: [TravelAgent]
  litellm.RateLimitError: RateLimitError: OpenAIException -
  Your organization's AI gateway access is exhausted. Contact your LangWatch admin.
```

The javascript side hit the same wall at the same moment, and the five tests the gate named are all scenario runs that need a model.

**Why the gate called it a defect anyway, and the fix in this PR.** `classify-examples-failures.py` recognised exactly one shape of refusal, HTTP 402 `budget_exceeded`. A call that leaves through litellm comes back as a RateLimitError carrying the gateway's own sentence instead, and the 402 never reaches the JUnit report. So the classifier saw no budget marker and returned `mixed`, which the shell gate reports as *"The examples suite has failures that are not gateway budget refusals ... A budget outage does not excuse them."* That is the false alarm the script was written to prevent, so the marker set now includes the gateway's sentence.

The marker is the sentence, not the exception class, deliberately. Matching `ratelimiterror` would bury a real bug: an example that mishandles a genuine 429 from a provider should still fail. Checked against four reports:

| report | verdict |
|---|---|
| all failures are gateway-exhaustion refusals | `budget 2` |
| gateway refusal + one real assertion failure | `mixed`, naming the real one |
| a plain provider 429 | `mixed` |
| the original 402 `budget_exceeded` | `budget 1`, unchanged |

**This does not make the check green, and it should not.** This PR edits files under `examples/`, so `EXAMPLES_TOUCHED` is true, and the gate's own rule is that an outage plus a changed example means the change went unverified and the check stays red. All the fix changes is the reason: from "these five examples are broken", which was not true, to "the suite could not run", which is.

So: **not ready to merge today.** It needs a re-run once the daily gateway budget window resets, and then the examples suite has to actually pass. Everything else on the PR is verified, including a fresh audit of all 59 resolved versions against every published advisory for their package, which is what turned up the `fast-uri` floor correction also in this push.

langwatch-agent:
Re-ran with the classifier fix. It did what it should, and it did not make the check green. Refining what I said earlier, because the picture is more precise now.

**Python: from 27 defects to 3.** The suite still fails 27 of 30, all in the same outage. The gate now names only three:

```
The examples suite has failures that are not gateway budget refusals:
  examples.test_testing_remote_agents_sse::test_sse_response;
  examples.test_testing_remote_agents_stateful::test_stateful_conversation;
  examples.test_testing_remote_agents_streaming::test_streaming_response
```

**JavaScript: a different five.** The five it named before (image demo, api-service-mocking, custom-judge-llm, database-tool-mocking, error-handling) are now correctly classified as budget refusals. What remains is the remote-agent and realtime-voice set.

**Why that residue exists, and why I am not widening the matcher to swallow it.** Those examples do not call the model directly. They stand up their own HTTP endpoint and the *endpoint* calls the model. When the gateway refuses, the endpoint's response is cut short, and what reaches the test is a transport error:

```
[SSEAgentAdapter] Response payload is not completed:
  <TransferEncodingError: 400, message='Not enough data to satisfy transfer length header.'>
```

The gateway's sentence never appears in the report, so no honest marker can catch it. Adding `TransferEncodingError` to `BUDGET_MARKERS` would classify a genuine truncated-response bug as an infrastructure excuse, which is the exact failure mode the script's header warns about. I would rather the gate stay noisy here than teach it to hide a transport defect.

**Not caused by this PR.** Reproduced on `origin/main` (`edf70006`) with no model credentials, same test, same error:

```
FAILED examples/test_testing_remote_agents_sse.py::test_sse_response - Runtim...
1 failed, 1 warning, 2 rerun in 8.39s
```

Same failure on this branch. This PR changes no Python source at all; its only files under `python/` are three JavaScript manifests in the lovable_clone template.

**Where that leaves it.** Still not mergeable today, for the reason the gate was designed to enforce: this PR edits files under `examples/`, the suite could not run, so nothing verified the change. It needs a re-run after the daily gateway budget resets. Everything the outage does not touch is verified, including the javascript workspace at 1557 passed and a fresh audit of all 59 resolved versions against every published advisory for their package.

PR #783, opened 2026-07-10
feat(voice): instrument voice.stt.transcribe span (#776)

drewdrewthis:
> **Reviewer cockpit**
> - **Scariest thing** — STT is best-effort (a failure must NEVER fail the run). Two ways this diff could break that: (1) a provider exception now runs inside `voice_span`, whose `recordException` **exports** the message — OpenAI/ElevenLabs SDK errors embed the raw response body (+ a key fragment on a 401) → **telemetry leak** (found in review, **fixed**: sanitized at both sites); (2) a throw in `execute()`'s `finally` would **mask the scenario result** (the TS batch body is a `Promise.all` whose every element has its own try/catch → never rejects → `backfillSegmentTranscripts` stays non-throwing; the per-segment catch is now commented load-bearing).
> - **Start here** — `python/scenario/voice/adapter.py` `_ensure_transcript` and `javascript/src/execution/scenario-execution.ts` `transcribeSegment` / `backfillSegmentTranscripts`. They emit the *same* span at *different* trace depths on purpose (the #776 decision below).

## Why

Closes #776. #771 landed the base `voice.*` taxonomy but **deferred `voice.stt.transcribe`** because the STT seam is structurally asymmetric across the two SDKs and needed a position decision, not just code. This is PR6 of epic #770.

## What changed

- **`voice.stt.transcribe` = the atomic per-chunk STT span in BOTH languages** → *(alt: a distinct `voice.stt.*` name per language)* → same name + same attributes means one cross-language query works; the parent differs because each SDK calls STT from a structurally different place, encoded honestly in the trace **shape** + a `voice.stt.scope` attr.
- **Python: per-turn, nested under `voice.turn`** (`_ensure_transcript`, inside `call()`) → `voice.stt.scope="turn"`.
- **TS: per-run, under a NEW `voice.stt.backfill` batch parent** (`backfillSegmentTranscripts`, in `execute()`'s `finally`) → *(alt: bare spans)* → but `execute()` has **no run-root span** there, so bare spans would **orphan into N separate traces**; the batch span groups them and carries the run-correlation attrs (`scenario.run_id` / `langwatch.thread_id` / `langwatch.origin`) so the per-run trace is attributable to the run. `voice.stt.scope="run"`.
[...]

PR #880, opened 2026-08-09
feat(voice): instrument judge STT pre-pass

### LINKED ISSUE #785: feat(voice): instrument the judge pre-pass STT (voice.stt.transcribe undercounts)
Follow-up from the #783 review (part of #770).

#776/#783 instruments `voice.stt.transcribe` only on the **manifest back-fill** STT pass (TS `backfillSegmentTranscripts`, Py `_ensure_transcript`). A voice run makes a **second** STT pass — the **judge pre-pass** — which is left uninstrumented in BOTH languages:
- TS: `voice/judge-stt.ts` (`transcribeMessage`/`transcribeAudioMessages`), via `prepareJudgeInput` from `agents/judge/judge-agent.ts`.
- Py: `voice/_transcribe.py` (`transcribe_segments`/`_transcribe_one`), from `judge_agent.py`.

Consequence: any "how much STT is this run doing / how slow is STT" query via `voice.stt.transcribe` **undercounts** — it misses one of two provider passes over overlapping audio. (The judge pre-pass was deliberately scoped OUT of #776 because it is *symmetric* across both languages — no py-vs-ts position decision to make — but the undercount is worth closing.)

Wrap those call sites in `voice.stt.transcribe` identically (scope="judge" or similar), applying the SAME exception-sanitization guard #783 added (provider errors must not export raw response bodies to telemetry).

Also worth a small test-hardening rider: a mixed-outcome batch case (one segment succeeds while a sibling fails in the same `Promise.all`) — the failure-isolation is proven by separate tests but not composed.

prajvalrasik:
## Why

The judge pre-pass performs a second STT pass that was absent from `voice.stt.transcribe` telemetry in both runtimes, so STT volume and latency were undercounted. Closes #785.

## What changed

- Make telemetry opt-in at both judge boundaries so direct and public transcription callers keep their existing uninstrumented behavior.
- Sanitize provider exceptions before OpenTelemetry and application logs record them, while preserving the existing per-message failure isolation.
- Keep the TypeScript and Python attributes aligned and cover mixed success and failure batches in both runtimes.

## How it works

`prepareJudgeInput` and `JudgeAgent` opt into a `voice.stt.transcribe` span for each audio message or segment. Direct TypeScript and Python transcription callers remain unchanged and uninstrumented.

## Test plan

- `python/.venv/Scripts/python.exe -m pytest python/tests/voice/test_judge_stt_telemetry.py python/tests/test_judge_agent.py -q` - 28 passed
- `cd javascript && vitest run src/voice/__tests__/judge-stt-telemetry.test.ts` - 3 passed
- `cd javascript && eslint src/voice/__tests__/judge-stt-telemetry.test.ts`
- `cd javascript && tsc --noEmit`
- Python syntax compilation for the changed module and test passed.

## Human verification

This is an internal telemetry change with no UI surface. The regression tests verify the exported span attributes, mixed-batch isolation, absence of raw provider error content, and that direct transcription remains uninstrumented by default.

### DIFF
--- a/javascript/src/voice/__tests__/judge-stt-telemetry.test.ts
+++ b/javascript/src/voice/__tests__/judge-stt-telemetry.test.ts
@@ -7,7 +7,7 @@ import type { ModelMessage } from "ai";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 import { AudioChunk } from "../audio-chunk";
-import { prepareJudgeInput } from "../judge-stt";
+import { prepareJudgeInput, transcribeAudioMessages } from "../judge-stt";
 import { createAudioMessage } from "../messages";
 import type { STTProvider } from "../stt";
 
@@ -66,6 +66,23 @@ describe("judge pre-pass voice.stt.transcribe spans (#785)", () => {
       "account restored".length,
     );
     expect(spans[0]!.attributes["langwatch.span.type"]).toBe("span");
+  });
+
+  it("leaves direct transcription uninstrumented by default", async () => {
+    const stt: STTProvider = {
+      transcribe: vi.fn(async () => "direct transcript"),
+    };
+
+    await transcribeAudioMessages({
+      messages: [createAudioMessage(tone(1), "user") as ModelMessage],
+      stt,
+      includeAudio: false,
+    });
+
+    const spans = exporter
+      .getFinishedSpans()
+      .filter((span) => span.name === "voice.stt.transcribe");
+    expect(spans).toHaveLength(0);
   });
 
   it("keeps a successful sibling when one message fails and exports only a sanitized error", async () => {
--- a/javascript/src/voice/__tests__/judge-stt-telemetry.test.ts
+++ b/javascript/src/voice/__tests__/judge-stt-telemetry.test.ts
@@ -1,0 +1,110 @@
+/** Judge pre-pass STT span coverage for issue #785. */
+
+import { trace, SpanStatusCode } from "@opentelemetry/api";
+import { InMemorySpanExporter, SimpleSpanProcessor } from "@opentelemetry/sdk-trace-base";
+import { NodeTracerProvider } from "@opentelemetry/sdk-trace-node";
+import type { ModelMessage } from "ai";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+import { AudioChunk } from "../audio-chunk";
+import { prepareJudgeInput } from "../judge-stt";
+import { createAudioMessage } from "../messages";
+import type { STTProvider } from "../stt";
+
+function tone(marker: number): AudioChunk {
+  const data = new Uint8Array(4800);
+  data.fill(marker);
+  return new AudioChunk({ data });
+}
+
+function textPart(message: ModelMessage): string | undefined {
+  const content = (message as { content?: unknown }).content;
+  if (!Array.isArray(content)) return undefined;
+  const part = content.find(
+    (item) =>
+      item !== null &&
+      typeof item === "object" &&
+      (item as { type?: unknown }).type === "text",
+  ) as { text?: string } | undefined;
+  return part?.text;
+}
+
+describe("judge pre-pass voice.stt.transcribe spans (#785)", () => {
+  let exporter: InMemorySpanExporter;
+  let provider: NodeTracerProvider;
+
+  beforeEach(() => {
+    exporter = new InMemorySpanExporter();
+    provider = new NodeTracerProvider({
+      spanProcessors: [new SimpleSpanProcessor(exporter)],
+    });
+    trace.setGlobalTracerProvider(provider);
+  });
+
+  afterEach(async () => {
+    await provider.shutdown();
+    trace.disable();
+  });
+
+  it("emits a judge-scoped span with success attributes", async () => {
+    const stt: STTProvider = {
+      transcribe: vi.fn(async () => "account restored"),
+    };
+    await prepareJudgeInput({
+      messages: [createAudioMessage(tone(1), "assistant") as ModelMessage],
+      stt,
+    });
+
+    const spans = exporter
+      .getFinishedSpans()
+      .filter((span) => span.name === "voice.stt.transcribe");
+    expect(spans).toHaveLength(1);
+    expect(spans[0]!.attributes["voice.stt.scope"]).toBe("judge");
+    expect(spans[0]!.attributes["voice.stt.speaker"]).toBe("assistant");
+    expect(spans[0]!.attributes["voice.stt.audio_bytes"]).toBe(4800);
+    expect(spans[0]!.attributes["voice.stt.transcript_chars"]).toBe(
+      "account restored".length,
+    );
+    expect(spans[0]!.attributes["langwatch.span.type"]).toBe("span");
+  });
+
+  it("keeps a successful sibling when one message fails and exports only a sanitized error", async () => {
+    const rawError = "401 invalid key sk-secret body={provider response}";
+    const stt: STTProvider = {
+      async transcribe(audio) {
+        if (audio.data[0] === 1) throw new Error(rawError);
+        return "successful sibling";
+      },
+    };
+    const warn = vi.fn();
+    const messages = [
+      createAudioMessage(tone(1), "user") as ModelMessage,
+      createAudioMessage(tone(2), "assistant") as ModelMessage,
+    ];
+
+    const prepared = await prepareJudgeInput({ messages, stt, logWarn: warn });
+
+    expect(textPart(prepared.messages[0]!)).toBeUndefined();
+    expect(textPart(prepared.messages[1]!)).toBe("successful sibling");
+    const spans = exporter
+      .getFinishedSpans()
+      .filter((span) => span.name === "voice.stt.transcribe");
+    expect(spans).toHaveLength(2);
+    const failed = spans.find((span) => span.status.code === SpanStatusCode.ERROR)!;
+    const succeeded = spans.find((span) => span.status.code !== SpanStatusCode.ERROR)!;
+    expect(failed.attributes["voice.stt.scope"]).toBe("judge");
+    expect(failed.attributes["voice.stt.speaker"]).toBe("user");
+    expect(failed.attributes["voice.stt.transcript_chars"]).toBeUndefined();
+    expect(succeeded.attributes["voice.stt.transcript_chars"]).toBe(
+      "successful sibling".length,
+    );
+    const recorded = JSON.stringify(failed.events);
+    expect(recorded).toContain("STT provider failed");
+    expect(recorded).not.toContain("sk-secret");
+    expect(recorded).not.toContain("401");
+    expect(recorded).not.toContain("provider response");
+    expect(warn).toHaveBeenCalledOnce();
+    expect(warn.mock.calls[0]![0]).toContain("STT provider failed: Error");
+    expect(warn.mock.calls[0]![0]).not.toContain(rawError);
+  });
+});
--- a/python/scenario/voice/_transcribe.py
+++ b/python/scenario/voice/_transcribe.py
@@ -32,6 +32,7 @@ async def transcribe_segments(
     recording: VoiceRecording,
     provider: Optional[STTProvider] = None,
     only_missing: bool = True,
+    telemetry_scope: Optional[str] = None,
 ) -> None:
     """
     Run STT over recording.segments, mutating .transcript in place.
@@ -43,6 +44,9 @@ async def transcribe_segments(
         only_missing: If True (default), skip segments whose transcript is
             already set. If False, re-transcribe everything (e.g. to overwrite
             adapter-side STT with a different provider).
+        telemetry_scope: When set, emit a ``voice.stt.transcribe`` span for
+            each provider call using this scope. Public callers are
+            uninstrumented by default; the judge path opts in with ``judge``.
 
     Concurrency: transcribes segments concurrently with asyncio.gather. Each
     segment's STT call is independent. Empty-data segments are skipped.
@@ -62,7 +66,7 @@ async def transcribe_segments(
     ]
     if not targets:
         return
-    await asyncio.gather(*(_transcribe_one(p, s) for s in targets))
+    await asyncio.gather(*(_transcribe_one(p, s, telemetry_scope) for s in targets))
 
 
 def _try_get_provider() -> Optional[STTProvider]:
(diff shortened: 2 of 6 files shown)


rogeriochaves:
Reviewing this so it can land. I have not pushed anything and I have not marked it ready. It is your PR.

## CI is unblocked and green

The workflow runs were sitting at `action_required`. That is what fork PRs do until a maintainer approves them, so none of the checks had ever run since you opened it. I approved them. `python-complete`, `javascript-complete` and `docs-complete` all pass now.

I also checked it locally: it rebases cleanly on current `main`, and the Python tests pass on that rebase.

## 1. The Python span scope is hardcoded where the TypeScript one is gated

`_transcribe_one` always sets `voice.stt.scope="judge"`.

On the TS side the span is opt-in. `telemetryScope` is only set by `prepareJudgeInput`, so a direct `transcribeAudioMessages` caller stays uninstrumented. That is exactly what the PR description says it does.

Python has no equivalent gate. `transcribe_segments` is a public export in `scenario/voice/__init__.py`, and its own docstring names a second intended caller: "Used by the judge fallback path for non-multimodal judges and as an opt-in to save_segments for richer manifests." So any non-judge caller would emit a span labelled `judge`.

Nothing is mislabelled today. `judge_agent.py:496` is the only production caller, so the label is correct in practice. But it goes wrong the moment a second caller appears, and it is the kind of py-vs-ts asymmetry langwatch/scenario#785 asked to avoid.

Suggestion, mirroring the TS shape:

```python
async def transcribe_segments(
    recording: VoiceRecording,
    provider: Optional[STTProvider] = None,
    only_missing: bool = True,
    telemetry_scope: Optional[str] = None,
) -> None:
```

Emit the span only when `telemetry_scope` is set, and pass `telemetry_scope="judge"` from `judge_agent.py:496`. Then both runtimes opt in at the judge boundary, instead of one opting in and the other assuming.

## 2. The WARNING log now loses the provider's own message

Worth knowing, not a change request.

Before, the outer handler formatted the original exception, so a default-level log read roughly:

```
STT failed for agent segment at 1.20s: ElevenLabs STT HTTP 401 (see DEBUG log for response body)
```

Now it formats the sanitized `RuntimeError`, so it reads:

```
STT failed for agent segment at 1.20s: STT provider failed: RuntimeError
```

and the real detail only exists at DEBUG.

This is defensible. It is a faithful copy of the #783 guard in `voice/adapter.py`, which does the same thing, and keeping response bodies out of telemetry is the whole point. Just flagging that operators lose a level of detail at default verbosity, and that the sanitized text reads oddly when the source was already a `RuntimeError`.

## Test coverage

Good. Both runtimes, mixed success and failure inside one batch, and the sanitization asserted against the raw key and body rather than just the message shape. That last part is the one people usually skip.

## To land it

Mark it ready for review, then get one approving review. The auto-approve workflow skips fork PRs by design, so it will never approve this one.

One red check you can ignore: `Validate PR Title` fails with `Resource not accessible by integration`. That is `conventional-commits.yml` setting `validateSingleCommit: true` while granting only `pull-requests: read` and `statuses: write`, so it breaks on any one-commit PR. Your title is fine, and it is not a required check.

prajvalrasik:
Thanks for the careful review. I addressed the Python scope asymmetry in e28991d: `transcribe_segments` now accepts an optional `telemetry_scope` and remains uninstrumented by default, while `JudgeAgent` explicitly opts in with `telemetry_scope="judge"`.

I also added coverage proving a direct public call still transcribes without emitting a judge span, and updated the judge-boundary call assertions. The branch is rebased onto current `main`; the focused checks pass (48 Python tests, 10 TypeScript tests, Pyright, Ruff, ESLint, and `tsc`). I'll mark it ready once the refreshed CI completes.

langwatch-agent:
## Human Review Brief

**Mode: Targeted Review.** Closes #785, and the PR does what the issue asks: the judge pre-pass STT was absent from `voice.stt.transcribe`, so volume and latency were undercounted.

### What you are looking at, if the voice subsystem is not your area

### Decisions being ratified

1. **Telemetry is opt-in at the judge boundaries rather than on by default in the transcribe path.** Direct and public transcription callers stay uninstrumented, so the span means "a judge transcribed this", not "a transcription happened". Anyone reading `voice.stt.transcribe` volume later has to know that.
2. **Provider exceptions are sanitized before OpenTelemetry records them.** That decides what an operator can see when a provider fails, permanently, and it is the kind of filter that quietly removes the one field someone needed.
3. **Attributes are kept aligned across the TypeScript and Python runtimes.** Two implementations of one attribute set means the alignment has to be tested, not intended.

### Must Check

- **What sanitization drops.** Confirm the provider's error type or status code survives, even if the message does not. A sanitized exception that reduces to a generic string makes the new telemetry unable to answer why STT failed.
- **Mixed success and failure batches.** The body says per-message failure isolation is preserved. Confirm a failing segment produces a span with an error status rather than no span, otherwise the undercount this PR fixes comes back for exactly the failing case.

### Ask Author

If the opt-in flag is ever set on a direct caller, do the two runtimes emit identical attributes? That is the moment the alignment claim gets tested for real.

python/scenario/voice/_transcribe.py
```diff
@@ -76,10 +82,41 @@ def _try_get_provider() -> Optional[STTProvider]:
...
+                "voice.stt.audio_bytes": len(segment.audio),
+            },
+        ) as stt_span:
+            try:
+                text = await provider.transcribe(AudioChunk(data=segment.audio))
+            except Exception as exc:
+                # Provider SDK errors can include response bodies and key fragments.
+                # Keep the raw detail local and let telemetry record only a minimal
+                # provider-agnostic exception, matching the #783 STT guard.
+                logger.debug(
```
langwatch-agent: **P2 — Do not log the raw provider exception in the sanitized path.** `exc_info=True` serializes the original exception message and traceback into debug logs. STT SDK failures can contain provider response bodies or credential fragments—the exact data this change prevents from reaching telemetry—so enabling debug logging reintroduces that disclosure under a second observable path. Log only the exception type (or remove this debug entry) before raising the sanitized `RuntimeError`.

langwatch-agent (commented):
Found one privacy issue in the Python STT failure path. I reviewed this external contribution statically; I did not execute contributor code or install dependencies.

LangWatch-Review: verdict=findings sha=e28991d907daaac970b3d5c30d50954799ba1076 p0=0 p1=0 p2=1 p3=0

python/scenario/voice/_transcribe.py
```diff
@@ -76,10 +82,41 @@ def _try_get_provider() -> Optional[STTProvider]:
...
+                    "scenario.voice.transcribe: STT provider error detail",
+                    exc_info=True,
```
drewdrewthis: **[review-clerk, Fix]** This `logger.debug(..., exc_info=True)` still serializes the raw provider exception (message + traceback) into the `scenario.voice` DEBUG logger — the exact content this PR sanitizes out of the OTel span. If DEBUG logging is ever enabled for this logger (local debugging, a verbose log level in prod), provider response bodies / key fragments (e.g. `sk-...`) leak through a second, unsanitized channel, defeating the PR's own stated goal.

This concurs with the still-unresolved `langwatch-agent` review thread on this same line (posted 2026-08-13T06:52:00Z, `isResolved=false`) — flagging as a review-clerk finding too so it participates in this review's own gate. Suggest: drop `exc_info=True` (log only `type(exc).__name__`, already available at line 115) or move the raw detail behind an explicit opt-in flag, never a bare DEBUG level on this logger.

javascript/src/voice/__tests__/judge-stt-telemetry.test.ts
```diff
@@ -0,0 +1,110 @@
...
+    const recorded = JSON.stringify(failed.events);
+    expect(recorded).toContain("STT provider failed");
+    expect(recorded).not.toContain("sk-secret");
+    expect(recorded).not.toContain("401");
+    expect(recorded).not.toContain("provider response");
+    expect(warn).toHaveBeenCalledOnce();
+    expect(warn.mock.calls[0]![0]).toContain("STT provider failed: Error");
+    expect(warn.mock.calls[0]![0]).not.toContain(rawError);
+  });
+});
```
drewdrewthis, line 127: **[review-clerk, Fix/New AC]** Missing test: nothing here proves a direct (non-judge) `transcribeAudioMessages` call stays uninstrumented — i.e. emits zero `voice.stt.transcribe` spans when `telemetryScope` is omitted.

The Python side has this exact test (`test_direct_transcription_is_uninstrumented_by_default` in `python/tests/voice/test_judge_stt_telemetry.py`), and the PR description explicitly claims parity: "Direct TypeScript and Python transcription callers remain unchanged and uninstrumented." Right now only the Python half of that claim is evidenced. Add a JS sibling test calling `transcribeAudioMessages` (or the underlying non-judge path) without `telemetryScope` and asserting `exporter.getFinishedSpans()` has no `voice.stt.transcribe` span.

drewdrewthis (commented):
Automated pr-review pass. Inline findings below; verdict comment upserted separately.

drewdrewthis (changes requested):
See verdict comment below for full details. NOT-READY: 2 new review-clerk findings (DEBUG-log leak of raw provider exception; missing JS parity test) plus 1 pre-existing unresolved thread from langwatch-agent. CI has not actually run on this SHA — see verdict.

drewdrewthis:
## Review verdict: READY

**Reviewed at:** `7e936e78` · **Run:** pr-review (reviewer: @review-clerk)

<details>
<summary>Review notes (2 passes + CI check)</summary>

- **Pass 1 (correctness):** `voiceSpan`/`voice_span` usage is correct on both runtimes — pre-existing guarded-span infra, not reinvented (Pass 3 clean). Provider exceptions are sanitized (`STT provider failed: <ExceptionType>`) before being recorded on the span or logged, consistent with the existing #783 pattern already used in `python/scenario/voice/adapter.py` and `javascript/src/execution/scenario-execution.ts`. The outer per-message/per-segment `catch`/`except` still fires regardless of the inner sanitization, so failure isolation and the transcript-cache negative-sentinel behavior (#735 P2) are preserved.
- **Pass 2 (structural):** `telemetryScope`/`telemetry_scope` is opt-in and defaults to uninstrumented for direct/public transcription callers, matching the PR description. No duplicated helpers; attribute names (`voice.stt.scope`, `voice.stt.speaker`, `voice.stt.audio_bytes`, `voice.stt.transcript_chars`) match 1:1 across TypeScript and Python.
- **CI (check-runs, not rollup):** all gating jobs (`python-complete`, `javascript-complete`, `docs-complete`, `test (3.12)`, `ci-checks (24.x)`, PR-title validation) are `SUCCESS` at this SHA; `build`/`comment`/`firefighting`/`preflight`/`evaluate`/`dismiss-firefighting-approval` are `SKIPPED` (gated, not applicable).
- Reviewed from a fresh clone of `pull/880/head` at `7e936e781f64126afa4c460818df449f0f7c1540` (no shared/live checkout used). This SHA postdates the 2026-09-17T04:10Z human DISMISSED review, so this is a fresh re-review of the current head, not a re-assertion of the stale one.

</details>

---
_Verdict is prose, not a GitHub approval. Scope: review findings only — READY means no unresolved blocking review threads at this SHA. It is **not** a merge-readiness signal._

langwatch-agent (commented):
No actionable correctness, security, or integrity finding in the external diff. The judge-only instrumentation leaves public direct transcription unchanged and replaces provider errors before telemetry/logging observes them. I performed static review only; contributor code and dependencies were not executed.

LangWatch-Review: verdict=clean sha=fac25413a1562c1f16b4dcb415cc8df7a056a5a5 p0=0 p1=0 p2=0 p3=0

langwatch-agent (commented):

LangWatch-Review: verdict=clean sha=fac25413a1562c1f16b4dcb415cc8df7a056a5a5 p0=0 p1=0 p2=0 p3=0

python/scenario/voice/_transcribe.py
```diff
@@ -76,10 +82,37 @@ def _try_get_provider() -> Optional[STTProvider]:
...
+        ) as stt_span:
+            try:
+                text = await provider.transcribe(AudioChunk(data=segment.audio))
+            except Exception as exc:
+                # Provider SDK errors can include response bodies and key fragments.
+                # Record only a provider-agnostic exception, matching the #783
+                # STT guard, so logs and telemetry expose the same safe detail.
+                raise RuntimeError(
+                    f"STT provider failed: {type(exc).__name__}"
+                ) from None
```
drewdrewthis, line 112: **[review-clerk, Fix]** CI blocker (PR-level, anchored here for visibility): `python-ci`, `javascript-ci`, and `docs-ci` all show `conclusion: failure` with **zero jobs executed** at `fac25413a1562c1f16b4dcb415cc8df7a056a5a5` (`gh api repos/langwatch/scenario/actions/runs/31887242004/jobs` → `total_count: 0`). This is the fork-PR workflow-approval gate expiring unapproved, not a real test failure — but it means no pytest/vitest has run against this SHA at all. A maintainer must re-approve the workflow runs (Actions tab → "Approve and run") before this PR's test claims can be trusted. This is a re-flag of the same CI gate noted in the prior NOT-READY verdict at `e28991d9` — it was never re-approved after this push.

drewdrewthis (commented):
Automated pr-review pass on fac25413a (post requested-changes push). The 3 prior review-clerk/langwatch-agent inline threads are resolved — the DEBUG-log leak and missing JS parity test are both fixed. New blocker: python-ci/javascript-ci/docs-ci all show conclusion=failure with zero jobs run at this SHA (workflow-approval gate never granted a second time) — no test in this diff has actually executed. See verdict comment for detail.

drewdrewthis (dismissed):
See verdict comment below for full details. NOT-READY: 1 new blocking finding (CI never actually ran at this SHA — python-ci/javascript-ci/docs-ci show 0 jobs executed, workflow approval gate not re-granted). All 3 prior review-clerk/langwatch-agent inline threads are resolved.

langwatch-agent (commented):
No new actionable findings. The earlier raw-exception debug-log finding is fixed. Static review covered judge-only opt-in, both runtimes’ span wrappers, per-segment failure isolation, sanitized exception recording, and the added telemetry tests. Direct transcription stays uninstrumented by default; raw provider errors are not attached as causes on the new sanitized path. No external branch code, dependencies, or tests executed; visible checks report no failed or pending checks. Runtime exporter behavior remains unverified by this review.

LangWatch-Review: verdict=clean sha=7e936e781f64126afa4c460818df449f0f7c1540 p0=0 p1=0 p2=0 p3=0

drewdrewthis (commented):
READY — see verdict: https://github.com/langwatch/scenario/pull/880#issuecomment-5278074811

Issue #888, opened 2026-08-11
Red-team report fails open: analyzer failures render green, JS/Python status mismatch hides broken runs, early-exit files won attacks as held


sergioestebance:
## What this is

The red-team report pipeline fails open in four independent ways: analyzer failures render compromised runs as green, a JS/Python wire mismatch silently hides broken runs from the dashboard, judge failures are filed as significant security breaks, and the early-exit path files a *successful attack* as "attacks held". For a security-testing feature, every one of these points the wrong way.

All findings adversarially verified end-to-end (writer → saved report → Streamlit dashboard).

## 1. Analyzer failure renders "RISK · NONE" in green

`python/scenario/report/_save.py:185-187` coerces any `break_severity` outside `{"none","partial","significant","complete"}` — missing, null, typo, non-string — to `"none"`. Worse, the exception path does the same: `_save.py:262-277` catches every analyzer failure (429, timeout, auth, JSON error) and keeps the defaults from `:249-257`, which already carry `break_severity: "none"`. The dashboard's compound risk matrix maps (any severity, "none") → risk "none" (`report/app.py:35-47`), rendered green (`:59`). So a confirmed compromise whose analyzer call failed renders **"RISK · NONE" in green**; the only trace is a `failure_summary` string "(analyzer failed: ...)".

This is the default path, not opt-in — the pytest plugin auto-saves with `analyze=True` for any test containing a RedTeamAgent (`python/scenario/pytest_plugin.py:22-64`).

The status-based fallback at `app.py:74-81` that would handle this correctly is dead code: every writer in the repo emits a value passing the `in _BREAK_ORDER` check first.

## 2. JS writes "broken", Python expects "broke" — broken runs vanish

`javascript/src/red-team-report.ts:111` writes status `"broken"`; the Python writer and dashboard use `"broke"` (`report/_save.py:299`, `app.py:908-910`). Both writers share the same output directory and dashboard by design (`red-team-report.ts:5`). A JS-written broken run matches none of the dashboard's three status buckets: it is dropped from the Held/Compromised/Errored tiles (while still in Total, so the tiles stop adding up), excluded from "Findings Requiring Attention" and Prioritized Fixes, and appears only as a red X in the per-test checklist. A JS suite where 3 of 10 runs break renders "Held 7 / Compromised 0 / Errored 0". Silent under-reporting of exactly the runs the report exists to surface.

## 3. Every non-success is filed as a "significant" break

[...]

Issue #887, opened 2026-08-11
Judge verdict pipeline loses the third state: unreachable INCONCLUSIVE, forced checkpoints, unvalidated criteria, JS/Python divergence


sergioestebance:
## What this is

Follow-up family to #886. Verifying that bug surfaced four more defects in the same pipeline — the judge's verdict handling loses or mangles the third state at every stage, and the JS and Python SDKs disagree with each other about what a verdict means. Together they mean a scenario can fail with an explanation that says it passed, pass with zero criteria verified, or die at a checkpoint the conversation hadn't reached yet.

All findings below were adversarially verified by tracing the full code path (including the AI SDK's actual invalid-tool-call behavior).

## 1. The emit boundary makes INCONCLUSIVE unreachable

`javascript/src/execution/scenario-execution.ts:2558` maps every result as `result?.success ? Verdict.SUCCESS : Verdict.FAILURE`; all five `emitRunFinished` call sites funnel through it. Python mirror: `python/scenario/scenario_executor.py:1836-1839`. `Verdict.INCONCLUSIVE` exists in the wire schema (`javascript/src/events/schema.ts:12`) and is emitted by neither SDK — a judge that honestly says "can't tell" is indistinguishable from a hard failure on every downstream surface. (The platform separately emits INCONCLUSIVE on user cancellation, so the value in stored data currently means "cancelled", never "judge hedged" — a second reason to make the SDKs emit it truthfully.)

## 2. Forced checkpoints turn "can't tell yet" into a dead scenario

A scripted `judge()` checkpoint sets `judgmentRequest`, which forces `toolChoice: finish_test` (`javascript/src/agents/judge/judge-agent.ts:566-569`) — the judge cannot continue. The checkpoint consumer (`scenario-execution.ts:2261-2277`) terminates the whole scenario on any non-success. So a mid-script checkpoint whose criteria the conversation hasn't reached yet ends the run as failed. Python mirror: `scenario_executor.py:1688-1704` + `judge_agent.py:644`.

Two verified nuances:
- Platform-triggered runs are scriptless, so this bites code-based SDK users only.
- The force silently disappears when the trace is large: `invokeLLMWithDiscovery` overwrites `toolChoice` to `"required"` (`judge-agent.ts:625-631`) — same checkpoint, opposite semantics, decided by trace size.

## 3. JS accepts malformed criteria payloads — including unverified success

[...]

PR #889, opened 2026-08-11
fix(judge): treat an unforced inconclusive finish_test as continue instead of failing the run

### LINKED ISSUE #886: Judge ends the run mid-conversation with finish_test(verdict: inconclusive), reported as FAILED — looks like the user simulator stopped responding
## What happens

When the judge decides a conversation is *too early to call*, it can call `finish_test` with `verdict: "inconclusive"` instead of `continue_test`. Any `finish_test` call is terminal, and every verdict except `"success"` is reported as a failure. The simulation ends right after the agent's first reply, and in the simulations UI the last visible message is the agent asking the user a question — so it looks like the user simulator simply stopped responding.

Anyone running simulations hits this as intermittent one-turn FAILED runs where nothing actually failed: the judge's own recorded reasoning explicitly concludes the verdict should be inconclusive rather than success or failure — and the run is still reported as FAILED with the inconclusive criteria counted as unmet.

## Mechanism

Three pieces combine, in both SDKs:

1. **The `finish_test` tool advertises `"inconclusive"` as a valid final verdict** — `javascript/src/agents/judge/judge-agent.ts:270-272`, `python/scenario/judge_agent.py:619`. A strict judge model that concludes "my current assessment is inconclusive" reads that enum as the way to *report* the assessment, not as a way to *end the test*. Reasoning-family judge models (observed with gpt-5.1) do this readily on criteria phrased as end-states (e.g. "agent asks the user to confirm before applying the change"), because on turn 1 those are necessarily inconclusive.

2. **Every `finish_test` call is terminal, even when nothing forced it.** In the default `[proceed()]` script the judge is invoked each turn with `toolChoice: "required"` and `continue_test` freely available (`judge-agent.ts:566-569`), but `parseToolCalls` returns a terminal result for any `finish_test` (`judge-agent.ts:753-776`). Only a literal `continue_test` call continues the conversation.

3. **The verdict enum is collapsed to a boolean.** `success = verdict === "success"` (`judge-agent.ts:769`, `python/scenario/judge_agent.py:1077`), criteria marked `"inconclusive"` are bucketed into `unmetCriteria` (`judge-agent.ts:764-766`), and the run-finished event maps the result to a binary SUCCESS/FAILURE (`javascript/src/execution/scenario-execution.ts:2558`). An honest "inconclusive" becomes an indistinguishable FAILED.

The forced-verdict paths (`isLastMessage`, `enforceJudgement`, discovery `forceVerdict`) are the only places an inconclusive final verdict is legitimate — there the model genuinely has no `continue_test` option.

## Reproduction

Deterministic, unit level:

1. Invoke the judge mid-conversation (not last turn, no `judgmentRequest`) and have the LLM return a `finish_test` tool call with `verdict: "inconclusive"` and some criteria `"inconclusive"`.
2. `parseToolCalls` returns `{ success: false, unmetCriteria: [...] }` — terminal — instead of `null` (continue).
3. The run finishes FAILED after that turn.

Live: any multi-step scenario whose criteria are end-state checks, run with the default `proceed()` script and a reasoning-family judge model, intermittently ends after the first exchange with reasoning stating the verdict "must be inconclusive".

## Fix

- In non-forced judge calls, treat `finish_test(verdict: "inconclusive")` as `continue_test`: return `null` from `parseToolCalls` so the conversation proceeds. Keep `"inconclusive"` meaningful only where the verdict is forced (`isLastMessage` / `enforceJudgement` / `forceVerdict`).
- Mirror the same change in `python/scenario/judge_agent.py` (same trap, same shape).
- Optional hardening: system prompt line "never call finish_test with an inconclusive verdict while the conversation can continue" — weaker than the code guard, not a substitute.
- Downstream: the platform vendors the JS SDK as a tarball, so the fix needs a re-vendor + redeploy there; the platform UI could also distinguish "ended early as inconclusive" from a criteria failure (separate issue).

## Acceptance criteria

- A mid-conversation `finish_test` with `verdict: "inconclusive"` does not end the run; the conversation continues to the next turn (JS and Python).
- A forced judgment (last turn / explicit `judge()` step / discovery exhaustion) can still return an inconclusive verdict, and its result is reported as not-success exactly as today.
- Regression test at the `parseToolCalls` (JS) / tool-call-handling (Python) level executing the real code path, plus one execution-level test proving the conversation proceeds past an inconclusive judge turn.

## 🤖 For Agents

Root cause (verified): `javascript/src/agents/judge/judge-agent.ts:753-776` — `case "finish_test"` returns a terminal result unconditionally; `:769` `success: verdict === "success"`; enum at `:270-272`. Python mirror: `python/scenario/judge_agent.py:604-619` (enum), `:1009` (default verdict), `:1077` (boolean collapse). Whether the call was forced is already computable at the call site: `judge-agent.ts:540` (`isLastMessage`), `:554` (`enforceJudgement`), `forceVerdict` at `:700`. Plan: thread a `forced: boolean` into `parseToolCalls`; when `!forced && verdict === "inconclusive"`, log and return `null`. Do not change the event schema. Write the failing test first; it must execute `parseToolCalls`, not string-compare prompts.

sergioestebance:
## For humans

When the judge in a simulation isn't sure yet, it sometimes ends the whole run with an "inconclusive" verdict instead of letting the conversation continue. The run then shows as FAILED right after the agent's last message — which looks exactly like the simulated user stopped responding. This PR makes "can't tell yet" mean "keep going": the conversation continues, and the judge only delivers a final verdict when one is actually required. Fixes #886.

## Cause

`parseToolCalls` treated **any** `finish_test` tool call as terminal and mapped every verdict ≠ `"success"` to a failure (`javascript/src/agents/judge/judge-agent.ts:753-776`; Python mirror `python/scenario/judge_agent.py`). The `finish_test` schema offers `"inconclusive"` as a verdict, so a judge model reporting "too early to tell" ended the run as FAILED — with `continue_test` freely available the whole time.

## Fix

- A `finish_test` with `verdict: "inconclusive"` (or a missing verdict, which defaults to inconclusive) is treated as `continue_test` — the judge returns the continue contract (`null` in JS, `[]` in Python) — **unless the judgment was forced**: last turn, an explicit `judge()` checkpoint, or discovery exhaustion. Forced judgments keep their terminal semantics and results.
- One `judgmentRequired` source of truth feeds both the `toolChoice` pinning and the new `verdictForced` flag, in both SDKs, and both are deliberately independent of `hasCriteria` (review round, `6a82e3b2`): a required judgment cannot be dodged via `continue_test`, so a criteria-less judge on the last turn delivers its terminal result instead of silently sliding into the generic max-turns failure. This is the one deliberate behavior change on a forced path vs. main — previously a criteria-less last-turn judge was left unpinned.
- The judge system prompt now says to prefer `continue_test` while the conversation can continue, and explicitly permits an honest inconclusive when a final verdict is required (both SDKs, so JS — which has no last-turn counter-message — doesn't pressure the model into fabricating a definitive verdict).

No event-schema change, no new wire verdict values.

## Failing test first

`javascript/src/agents/judge/__tests__/judge-agent.test.ts` (before the fix):

```
- Expected: null
+ Received: { "metCriteria": [], "success": false,
    "unmetCriteria": ["Agent requests approval before applying the change"] }
```

`javascript/src/execution/__tests__/judge-inconclusive-continues.test.ts` (real `ScenarioExecution` + real judge, only the LLM stubbed):

```
AssertionError: expected 1 to be 2   // judge called once — the run died at turn 1
```

Python `tests/test_judge_inconclusive_continues.py` failed the same way (`ScenarioResult` returned where `[]` was expected).

## After

- JS: 1086 passed (full `src` suite) + typecheck + `lint:all` clean
- Python: 653 passed (full suite minus on-demand voice tiers)
- New coverage: unforced inconclusive → continue (unit + execution level), forced paths pinned terminal (last turn, checkpoint, discovery exhaustion — both SDKs), unforced inconclusive during large-trace discovery → continue, and a required large-trace judgment that answers continue_test is forced to a terminal verdict rather than dodging into the max-turns failure (both SDKs; review round `77ce1dbd`).

## Sweep

Already done exhaustively before this fix — the same tri-state-collapse shape across both SDKs and the platform is mapped and filed as #887 (judge verdict pipeline), #888 (red-team report), and platform-side langwatch/langwatch#6833/#6834/#6835. Deliberately **not** fixed here (separate defect classes, separate reviewers):

- Emitting `INCONCLUSIVE` on forced final verdicts (today it still collapses to FAILURE at the emit boundary) — #887
- JS criteria-payload hardening parity with Python's #161 fix — #887
- Platform fold mapping inconclusive→FAILURE — langwatch/langwatch#6834

## Delivery note

The platform vendors this SDK as a tarball (`vendor/langwatch-scenario-1.0.0.tgz`); a follow-up PR in langwatch/langwatch re-vendors it after this merges — without that, production never sees this fix.

### DIFF
--- a/javascript/src/agents/judge/judge-agent.ts
+++ b/javascript/src/agents/judge/judge-agent.ts
@@ -576,8 +576,9 @@ export class JudgeAgent extends JudgeAgentAdapter {
     // judgment cannot be dodged via continue_test, criteria or not — the
     // criteria-less enforced case already returned above, and a criteria-less
     // last-turn judge still owes its terminal verdict. Known gap: the
-    // large-trace discovery path relaxes toolChoice to "required" below and
-    // still accepts continue_test there (inherited; tracked in #887).
+    // Standard path pins finish_test; the large-trace path relaxes it to
+    // "required" so the judge can use discovery tools, but still guarantees a
+    // finish_test outcome for a required judgment (see invokeLLMWithDiscovery).
     const toolChoice: ToolChoice<typeof tools> = judgmentRequired
       ? { type: "tool", toolName: "finish_test" }
       : "required";
@@ -598,6 +599,7 @@ export class JudgeAgent extends JudgeAgentAdapter {
       tools,
       toolChoice,
       isLargeTrace,
+      judgmentRequired,
     });
 
     // A verdict is "forced" when this judgment is required to be final: the
@@ -645,14 +647,22 @@ export class JudgeAgent extends JudgeAgentAdapter {
    *
    * When the trace is large, toolChoice is relaxed to "required" so the
    * judge can freely pick discovery tools (expand_trace/grep_trace) before
-   * being forced to a terminal decision.
+   * being forced to a terminal decision. A required judgment (last turn or an
+   * explicit judge() checkpoint) is still guaranteed a finish_test outcome:
+   * if discovery ends on continue_test — or exhausts its steps — we force a
+   * final verdict rather than let the run dodge into the generic max-turns
+   * failure (#886; the large-trace half of the same terminal contract).
    */
   private async invokeLLMWithDiscovery({
     isLargeTrace,
+    judgmentRequired,
     ...params
-  }: InvokeLLMParams & { isLargeTrace: boolean }): Promise<{
+  }: InvokeLLMParams & {
+    isLargeTrace: boolean;
+    judgmentRequired: boolean;
+  }): Promise<{
     completion: InvokeLLMResult;
-    /** True when the discovery loop exhausted and forceVerdict pinned finish_test. */
+    /** True when forceVerdict pinned finish_test (exhaustion, or a dodged required judgment). */
     verdictWasForced: boolean;
   }> {
     if (isLargeTrace) {
@@ -674,11 +684,43 @@ export class JudgeAgent extends JudgeAgentAdapter {
       })),
     });
 
-    if (isLargeTrace && this.discoveryExhausted(completion)) {
-      return { completion: await this.forceVerdict(params), verdictWasForced: true };
+    if (isLargeTrace) {
+      if (this.discoveryExhausted(completion)) {
+        return { completion: await this.forceVerdict(params), verdictWasForced: true };
+      }
+      // A required judgment must not be dodged via continue_test on a large
+      // trace: discovery relaxed the pin, so if the judge continued instead of
+      // finishing, force the final verdict now.
+      if (
+        judgmentRequired &&
+        !this.completionCalledTool(completion, "finish_test")
+      ) {
+        this.logger.debug(
+          "required judgment continued during large-trace discovery - forcing verdict"
+        );
+        return { completion: await this.forceVerdict(params), verdictWasForced: true };
+      }
     }
 
     return { completion, verdictWasForced: false };
+  }
+
+  /**
+   * True when `toolName` was called anywhere in the (possibly multi-step)
+   * completion — the aggregate `steps` array when present, else the final
+   * `toolCalls`. Mirrors how {@link discoveryExhausted} inspects the loop.
+   */
+  private completionCalledTool(
+    completion: InvokeLLMResult,
+    toolName: string
+  ): boolean {
+    const steps = completion.steps;
+    if (steps && steps.length > 0) {
+      return steps.some((step) =>
+        step.toolCalls?.some((tc) => tc.toolName === toolName)
+      );
+    }
+    return Boolean(completion.toolCalls?.some((tc) => tc.toolName === toolName));
   }
 
   /**
--- a/javascript/src/agents/judge/judge-agent.ts
+++ b/javascript/src/agents/judge/judge-agent.ts
@@ -236,6 +236,7 @@ ${criteriaList}
 <rules>
 - Be strict, do not let the conversation continue if the agent already broke one of the "do not" or "should not" criteria.
 - DO NOT make any judgment calls that are not explicitly listed in the success or failure criteria, withhold judgement if necessary
+- If you do not yet have enough information for a definitive verdict, call the continue_test tool to let the conversation play out - do NOT call finish_test with an "inconclusive" verdict while the conversation can still continue. When a final verdict is explicitly required (the conversation cannot continue), deliver your honest verdict - "inconclusive" is acceptable there
 </rules>
 `.trim();
 }
@@ -563,8 +564,16 @@ export class JudgeAgent extends JudgeAgentAdapter {
       };
     }
 
+    // A final judgment is required when the conversation cannot continue past
+    // this call: the last turn, or an explicit judge() step. Note this is
+    // about the CONTRACT, not the transport — on large traces
+    // invokeLLMWithDiscovery relaxes toolChoice to "required" so the model can
+    // use discovery tools, but a checkpoint/last-turn judgment is still
+    // terminal.
+    const judgmentRequired = isLastMessage || enforceJudgement;
+
     const toolChoice: ToolChoice<typeof tools> =
-      (isLastMessage || enforceJudgement) && hasCriteria
+      judgmentRequired && hasCriteria
         ? { type: "tool", toolName: "finish_test" }
         : "required";
 
@@ -576,7 +585,7 @@ export class JudgeAgent extends JudgeAgentAdapter {
       isLargeTrace,
     });
 
-    const completion = await this.invokeLLMWithDiscovery({
+    const { completion, verdictWasForced } = await this.invokeLLMWithDiscovery({
       model: mergedConfig.model,
       messages,
       temperature: mergedConfig.temperature,
@@ -586,7 +595,16 @@ export class JudgeAgent extends JudgeAgentAdapter {
       isLargeTrace,
     });
 
-    return this.parseToolCalls(completion, criteria);
+    // A verdict is "forced" when this judgment is required to be final: the
+    // last turn, an explicit judge() step, or the discovery loop exhausting
+    // its steps (forceVerdict pinned finish_test). Only a forced judgment may
+    // legitimately end the run on an inconclusive verdict (#886). Deliberately
+    // independent of hasCriteria: a criteria-less judge on the last turn still
+    // delivers its terminal result rather than silently continuing into the
+    // generic max-turns failure.
+    const verdictForced = judgmentRequired || verdictWasForced;
+
+    return this.parseToolCalls(completion, criteria, { verdictForced });
   }
 
   /**
@@ -627,7 +645,11 @@ export class JudgeAgent extends JudgeAgentAdapter {
   private async invokeLLMWithDiscovery({
     isLargeTrace,
     ...params
-  }: InvokeLLMParams & { isLargeTrace: boolean }): Promise<InvokeLLMResult> {
+  }: InvokeLLMParams & { isLargeTrace: boolean }): Promise<{
+    completion: InvokeLLMResult;
+    /** True when the discovery loop exhausted and forceVerdict pinned finish_test. */
+    verdictWasForced: boolean;
+  }> {
     if (isLargeTrace) {
       params.toolChoice = "required";
       params.stopWhen = [
@@ -648,10 +670,10 @@ export class JudgeAgent extends JudgeAgentAdapter {
     });
 
     if (isLargeTrace && this.discoveryExhausted(completion)) {
-      return this.forceVerdict(params);
+      return { completion: await this.forceVerdict(params), verdictWasForced: true };
     }
 
-    return completion;
+    return { completion, verdictWasForced: false };
   }
 
   /**
@@ -739,7 +761,8 @@ export class JudgeAgent extends JudgeAgentAdapter {
 
   private parseToolCalls(
     completion: InvokeLLMResult,
-    criteria: string[]
+    criteria: string[],
+    { verdictForced }: { verdictForced: boolean }
   ): JudgeResult | null {
     let args: FinishTestArgs | undefined;
     if (completion.toolCalls?.length) {
@@ -755,6 +778,20 @@ export class JudgeAgent extends JudgeAgentAdapter {
           args = toolCall.input as FinishTestArgs;
 
           const verdict = args.verdict || "inconclusive";
+
+          // "Can't tell yet" is not a verdict (#886). When nothing forced the
+          // judge to finish — continue_test was freely available — an
+          // inconclusive finish_test used to end the run as FAILED, which in
+          // the UI reads as the user simulator going silent mid-conversation.
+          // Treat it as continue_test and let the conversation play out; a
+          // FORCED judgment (last turn, judge() checkpoint, discovery
+          // exhaustion) keeps its terminal behavior unchanged.
+          if (!verdictForced && verdict === "inconclusive") {
+            this.logger.debug(
+              "finish_test returned an inconclusive verdict without a forced judgment - continuing the conversation"
+            );
+            return null;
+          }
           const reasoning = args.reasoning || "No reasoning provided";
           const criteriaArgs = args.criteria || {};
           const criteriaValues = Object.values(criteriaArgs);
--- a/javascript/src/agents/judge/judge-agent.ts
+++ b/javascript/src/agents/judge/judge-agent.ts
@@ -572,10 +572,12 @@ export class JudgeAgent extends JudgeAgentAdapter {
     // terminal.
     const judgmentRequired = isLastMessage || enforceJudgement;
 
-    // Pinned on judgmentRequired ALONE: a required judgment must not be
-    // dodgeable via continue_test, criteria or not — the criteria-less
-    // enforced case already returned above, and a criteria-less last-turn
-    // judge still owes its terminal verdict.
+    // Pinned on judgmentRequired ALONE: on the standard path a required
+    // judgment cannot be dodged via continue_test, criteria or not — the
+    // criteria-less enforced case already returned above, and a criteria-less
+    // last-turn judge still owes its terminal verdict. Known gap: the
+    // large-trace discovery path relaxes toolChoice to "required" below and
+    // still accepts continue_test there (inherited; tracked in #887).
     const toolChoice: ToolChoice<typeof tools> = judgmentRequired
       ? { type: "tool", toolName: "finish_test" }
       : "required";
--- a/javascript/src/agents/judge/judge-agent.ts
+++ b/javascript/src/agents/judge/judge-agent.ts
@@ -572,10 +572,13 @@ export class JudgeAgent extends JudgeAgentAdapter {
     // terminal.
     const judgmentRequired = isLastMessage || enforceJudgement;
 
-    const toolChoice: ToolChoice<typeof tools> =
-      judgmentRequired && hasCriteria
-        ? { type: "tool", toolName: "finish_test" }
-        : "required";
+    // Pinned on judgmentRequired ALONE: a required judgment must not be
+    // dodgeable via continue_test, criteria or not — the criteria-less
+    // enforced case already returned above, and a criteria-less last-turn
+    // judge still owes its terminal verdict.
+    const toolChoice: ToolChoice<typeof tools> = judgmentRequired
+      ? { type: "tool", toolName: "finish_test" }
+      : "required";
 
     this.logger.debug("Calling LLM", {
       model: mergedConfig.model,
--- a/python/scenario/judge_agent.py
+++ b/python/scenario/judge_agent.py
@@ -604,6 +604,7 @@ If you do have enough information, use the finish_test tool to determine if all 
 <rules>
 - Be strict, do not let the conversation continue if the agent already broke one of the "do not" or "should not" criterias.
 - DO NOT make any judgment calls that are not explicitly listed in the success or failure criteria, withhold judgement if necessary
+- If you do not yet have enough information for a definitive verdict, call the continue_test tool to let the conversation play out - do NOT call finish_test with an "inconclusive" verdict while the conversation can still continue. When a final verdict is explicitly required (the conversation cannot continue), deliver your honest verdict - "inconclusive" is acceptable there
 </rules>
 """,
             },
@@ -702,11 +703,26 @@ if you don't have enough information to make a verdict, say inconclusive with ma
                 reasoning="TestingAgent was called as a judge, but it has no criteria to judge against",
             )
 
+        # A final judgment is required when the conversation cannot continue
+        # past this call: the last turn, or an explicit judgment_request. This
+        # is about the CONTRACT, not the transport — the discovery loop applies
+        # the pinned tool_choice only on its final step, but a checkpoint/
+        # last-turn judgment is still terminal.
+        judgment_required = is_last_message or enforce_judgment
+
         tool_choice: Any = (
             {"type": "function", "function": {"name": "finish_test"}}
-            if (is_last_message or enforce_judgment) and has_criteria
+            if judgment_required and has_criteria
             else "required"
         )
+
+        # Only a forced judgment may legitimately end the run on an
+        # inconclusive verdict (#886); discovery exhaustion forces separately
+        # in _force_verdict. Deliberately independent of has_criteria: a
+        # criteria-less judge on the last turn still delivers its terminal
+        # result rather than silently continuing into the generic max-turns
+        # failure.
+        verdict_forced = judgment_required
 
         # Multi-step discovery loop for large traces
         if is_large_trace:
@@ -718,6 +734,7 @@ if you don't have enough information to make a verdict, say inconclusive with ma
                 working_messages=working_messages,
                 effective_criteria=effective_criteria,
                 input_messages=input.messages,
+                verdict_forced=verdict_forced,
             )
 
         # Standard single-call path for small traces
@@ -733,7 +750,13 @@ if you don't have enough information to make a verdict, say inconclusive with ma
             **self._extra_params,
         )
 
-        return self._parse_response(response, effective_criteria, messages, input_messages=input.messages)
+        return self._parse_response(
+            response,
+            effective_criteria,
+            messages,
+            input_messages=input.messages,
+            verdict_forced=verdict_forced,
+        )
 
     def _completion_with_reasoning_off_retry(self, **kwargs: Any) -> ModelResponse:
         """
@@ -914,6 +937,7 @@ if you don't have enough information to make a verdict, say inconclusive with ma
         working_messages: Sequence[ChatCompletionMessageParam],
         effective_criteria: List[str],
         input_messages: Sequence[Any],
+        verdict_forced: bool,
     ) -> AgentReturnTypes:
         """
         Runs the multi-step discovery loop for large traces.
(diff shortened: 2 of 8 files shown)


sergioestebance:
## Ruthless review — self-authored, so findings fixed in-branch and recorded here; approval stays human

**Verdict: no P0/P1 in the final diff.** Two review passes ran: an independent adversarial pass (separate agent context) on the near-final diff produced 8 findings, all addressed before commit; this final pass verified the post-review edits by executing the real code paths. Independence caveat, stated plainly: the final pass was performed by the author's context — the executed tests, not the author's judgment, are the evidence.

### What the first pass caught and this diff now fixes (recorded so the fixes are judgeable)

1. **Prompt rule pressured forced verdicts (was: silent verdict flip risk on the last turn, JS only).** The rule now explicitly permits an honest "inconclusive" when a final verdict is required — both SDKs, same sentence.
2. **`verdictForced` comment claimed toolChoice pinning that the large-trace path removes.** Reworked: one `judgmentRequired` source of truth feeds both the pinning and the forced flag, and the comments now describe the contract (checkpoint/last-turn judgments are terminal) rather than the transport.
3. **Criteria-less judge lost its terminal result on the last turn.** Forcing is now independent of `hasCriteria` — a zero-criteria judge's last-turn verdict stays terminal instead of sliding into the generic max-turns failure.
4. **Discovery paths untested.** Added both SDKs: unforced inconclusive during large-trace discovery → continues; discovery-exhaustion forced inconclusive → terminal. The JS exhaustion test kills the `verdictWasForced: false` mutation that previously no test would catch.
5. **Execution test coupled to the global span collector.** Now passes an explicit `JudgeSpanCollector`.

### Executed evidence (final committed tree)

- Failing-first outputs recorded in the PR body; post-fix: JS full `src` suite 1086 passed, typecheck + `lint:all` clean; Python 653 passed.
- Forced/unforced boundary pinned by 8 dedicated tests across both SDKs (mid-conversation, missing verdict, checkpoint, last turn, discovery ×2 per SDK).
- `git show a12bd1ba --stat -- javascript/src/events python/scenario/_events python/scenario/types.py` → empty: no wire/event schema change.

### Non-blocking residuals (P2/P3, deliberately not widening the PR)

- **P2** — Python has no executor-level "run continues past a hedge" test (JS does). Mitigation accepted: the fix returns `[]`, byte-identical to the `continue_test` contract at `judge_agent.py` (`continue_test → []`), whose executor handling existing proceed tests already cover.
- **P3** — the system-prompt rule is a text change visible to *forced* calls too. Intended (it is what fixes residual finding 1), but it means "forced path byte-identical" holds for code semantics, not prompt bytes.
- **P3** — `#886` remains half-delivered until the platform re-vendors the SDK tarball; noted in the PR body, tracked as the follow-up.

---

Blessed. 🙏 The judge may now say "not yet" without the run being condemned — hedging continues the conversation, and only a judgment that was truly demanded can end it.

> "Judge not, that you be not judged. For with the judgment you pronounce you will be judged." — Matthew 7:1-2

Traced the forced/unforced boundary through both SDKs and both discovery paths, executed the failing-first and passing runs, confirmed the wire schema untouched. Ship it — residuals are the Python executor-test gap and the prompt-text note above. Approval is the human's to give; this comment is the review record, not the approval.

javascript/src/agents/judge/judge-agent.ts
```diff
@@ -563,8 +564,16 @@ export class JudgeAgent extends JudgeAgentAdapter {
...
+    // invokeLLMWithDiscovery relaxes toolChoice to "required" so the model can
+    // use discovery tools, but a checkpoint/last-turn judgment is still
+    // terminal.
+    const judgmentRequired = isLastMessage || enforceJudgement;
+
     const toolChoice: ToolChoice<typeof tools> =
-      (isLastMessage || enforceJudgement) && hasCriteria
+      judgmentRequired && hasCriteria
         ? { type: "tool", toolName: "finish_test" }
         : "required";
```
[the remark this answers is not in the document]
sergioestebance: Deliberate, and inherited rather than introduced: the toolChoice pinning condition is byte-identical to main — a criteria-less last-turn judge could already pick continue_test before this PR, landing in the same generic max-turns result. This PR froze the forced-path transport on purpose (`judgmentRequired` governs the CONTRACT — what happens when finish_test arrives — not the pinning); forcing finish_test on an empty-criteria judge would change forced-call wire behavior beyond this fix scope. If pinning semantics should change, that belongs with the wider judge-verdict work in #887.

python/scenario/judge_agent.py
```diff
@@ -702,12 +703,27 @@ async def call(
...
+        # the pinned tool_choice only on its final step, but a checkpoint/
+        # last-turn judgment is still terminal.
+        judgment_required = is_last_message or enforce_judgment
+
         tool_choice: Any = (
             {"type": "function", "function": {"name": "finish_test"}}
-            if (is_last_message or enforce_judgment) and has_criteria
+            if judgment_required and has_criteria
             else "required"
         )
```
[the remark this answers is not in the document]
sergioestebance: Same as the JS thread: the has_criteria condition is unchanged from main and the described window (criteria-less last-turn judge choosing continue_test) behaves exactly as before this PR. The PR deliberately changes only what an arriving finish_test(inconclusive) means; pinning changes are out of scope here and tracked with the wider judge-verdict work in #887.

python/tests/test_judge_inconclusive_continues.py
```diff
@@ -0,0 +1,129 @@
...
+
+async def _run_judge(
+    *,
+    verdict: Optional[str],
+    judgment_request: Optional[JudgmentRequest],
+    current_turn: int,
+    max_turns: int = 10,
+) -> Any:
+    """Drive JudgeAgent through a finish_test tool call and return the raw result."""
+    ScenarioConfig.default_config = ScenarioConfig(default_model="openai/gpt-4")
```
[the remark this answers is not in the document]
sergioestebance: Fixed in b175f262 — the helper now saves and restores the prior ScenarioConfig.default_config, and returns AgentReturnTypes instead of Any.

javascript/src/agents/judge/judge-agent.ts
```diff
@@ -563,8 +564,16 @@ export class JudgeAgent extends JudgeAgentAdapter {
...
+    // this call: the last turn, or an explicit judge() step. Note this is
+    // about the CONTRACT, not the transport — on large traces
+    // invokeLLMWithDiscovery relaxes toolChoice to "required" so the model can
+    // use discovery tools, but a checkpoint/last-turn judgment is still
+    // terminal.
+    const judgmentRequired = isLastMessage || enforceJudgement;
+
     const toolChoice: ToolChoice<typeof tools> =
-      (isLastMessage || enforceJudgement) && hasCriteria
+      judgmentRequired && hasCriteria
```
langwatch-agent: **P1 — A final-turn judge without criteria is still allowed to continue.** `judgmentRequired` is true here, and `verdictForced` below assumes that guarantees a terminal result, but the retained `&& hasCriteria` makes `toolChoice` `"required"` for the supported `judgeAgent({ criteria: [] })` case. The model can therefore call `continue_test`; `parseToolCalls` returns `null` and execution falls through to the generic max-turn failure—the exact behavior the new comment says this avoids. Pin `finish_test` on `judgmentRequired` alone (the explicit checkpoint/no-criteria early return remains unchanged) and add the criteria-less last-turn regression.

python/scenario/judge_agent.py
```diff
@@ -702,12 +703,27 @@ async def call(
...
+        # past this call: the last turn, or an explicit judgment_request. This
+        # is about the CONTRACT, not the transport — the discovery loop applies
+        # the pinned tool_choice only on its final step, but a checkpoint/
+        # last-turn judgment is still terminal.
+        judgment_required = is_last_message or enforce_judgment
+
         tool_choice: Any = (
             {"type": "function", "function": {"name": "finish_test"}}
-            if (is_last_message or enforce_judgment) and has_criteria
+            if judgment_required and has_criteria
```
langwatch-agent: **P1 — Python has the same terminal-contract hole.** On a last turn with no effective criteria, `judgment_required` and `verdict_forced` are true but this `and has_criteria` leaves the model free to emit `continue_test`. `_parse_response` then returns `[]`, so the run reaches the generic max-turn failure rather than a forced final judgment. Make the forced `finish_test` choice depend only on `judgment_required`, mirroring the intended contract, and cover this path in Python as well.

langwatch-agent (commented):
2 P1s — the no-criteria final-turn contract remains broken in both SDKs; see inline comments.

javascript/src/agents/judge/judge-agent.ts
sergioestebance: Fixed in 6a82e3b2 — toolChoice now pins finish_test on judgmentRequired alone; the criteria-less enforced case still early-returns, and a criteria-less last-turn regression test asserts the pinned tool choice plus the terminal result.

python/scenario/judge_agent.py
sergioestebance: Fixed in 6a82e3b2 — tool_choice pins finish_test on judgment_required alone, mirrored from JS, with the criteria-less last-turn regression test covering the pinned choice and terminal ScenarioResult.

javascript/src/agents/judge/judge-agent.ts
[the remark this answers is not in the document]
sergioestebance: Correction: reversing the earlier skip. langwatch-agent raised the same point and the decisive argument stands — the new code comment claimed a terminal result the unpinned toolChoice could not guarantee. Fixed in 6a82e3b2 exactly as you proposed (pin on judgmentRequired alone), with the criteria-less last-turn regression test.

python/scenario/judge_agent.py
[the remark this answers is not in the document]
sergioestebance: Correction: reversing the earlier skip — fixed in 6a82e3b2 (tool_choice pinned on judgment_required alone, plus the criteria-less last-turn regression test). Your proposed fix was right.

python/tests/test_judge_inconclusive_continues.py
```diff
@@ -0,0 +1,191 @@
...
+    assert isinstance(result, ScenarioResult)
+    assert result.success is False
+
+
+@pytest.mark.asyncio
+async def test_criteria_less_last_turn_judge_pins_finish_test_and_stays_terminal():
+    """A required judgment is not dodgeable via continue_test even with no
+    criteria — otherwise the run falls through to the generic max-turns
+    failure and the judge's reasoning is lost."""
+    ScenarioConfig.default_config = ScenarioConfig(default_model="openai/gpt-4")
```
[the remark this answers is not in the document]
sergioestebance: Fixed in e1e023be — the new test now saves and restores the prior ScenarioConfig.default_config, same treatment as b175f262. Fair catch: the hand-rolled setup reintroduced the exact pattern you already flagged.

sergioestebance:
## Prove-them-wrong: live before/after on a real LangWatch stack

Reproduced the bug and verified the fix end-to-end against a running LangWatch instance (haven), with runs reported by the **scenario SDK** exactly as a customer's tests would report them, then read back from the Simulations UI.

**Method.** One scenario shape (an agent that asks clarifying questions before acting, three end-state approval criteria), run twice — once with the **published `@langwatch/scenario@1.0.0`** (the version the platform currently vendors) and once with **this PR's build**. The bug is intermittent by nature — it only bites when the judge chooses `finish_test("inconclusive")` early instead of `continue_test` — so to make the A/B deterministic and isolate the SDK as the *only* variable, the judge's turn‑1 decision is pinned to an inconclusive verdict (the exact shape a real judge emits). Everything else is the real SDK path: scenario execution, user simulator (`gpt-5-mini`), agent, judge model wiring (`gpt-5.1`), and event reporting to LangWatch. Fictional data throughout.

Buggy 1.0.0 — run ends FAILED after the agent's first question

Fixed (this PR) — same inconclusive turn 1, conversation continues to Turn 2 and PASSES

| # | Claim | Evidence | Status |
|---|-------|----------|--------|
| 1 | With the vendored SDK, a mid-conversation inconclusive verdict ends the run FAILED after a single exchange — the reported symptom | `01-buggy-1.0.0-failed-turn1.png`: Criteria 0/3, **FAILED**, only Turn 1 present, last message is the agent's clarifying question, reasoning "…the overall verdict at this point must be inconclusive, not success or failure" | PASS (bug reproduced) |
| 2 | With this PR, the identical turn‑1 inconclusive verdict lets the conversation continue and resolve | `02-fixed-PR889-continues-passed.png`: **TURN 2** present, Criteria 3/3, **PASSED** 100% | PASS (fix verified) |

Both runs used identical scripts and an identical pinned turn‑1 verdict; the **only** difference was the SDK version — so the SDK is the cause and this PR is the fix.

### Findings
- Console error on the run-detail drawer: `scenarios.getById → "Scenario not found"`. Environmental, not a PR‑889 regression — the reproduction reports runs via the SDK without first creating a matching **scenario template** in the library, so the drawer's template lookup 404s. Present identically on both the buggy and fixed runs. Worth a separate look (a run detail should degrade gracefully when its template was never saved), unrelated to this fix.

### Not covered
- Full platform `target.type:"code"` path (worker child process executing the vendored tarball): not exercised here — that lands with the separate re-vendor PR in `langwatch/langwatch`, which swaps `vendor/langwatch-scenario-1.0.0.tgz` to this build. The SDK code path proven above is the same judge logic that child process runs.
- Natural (non-pinned) judge reproduction: one run with a real unforced `gpt-5.1` judge did **not** trigger the early finish (it chose `continue_test`), which is consistent with the intermittency — the pinned A/B above is the deterministic proof.

**Verdict: 2/2 proven, 0 disproven, 0 unverified — on `https://app.<redacted-slug>.langwatch.localhost` (local haven stack, dogfood account).**

sergioestebance:
CI note: the red on `javascript-complete` / `python-complete` is entirely the **Test (Examples)** suite hitting the LLM gateway daily budget cap (`budget_exceeded`, `fault: customer` — 330 occurrences in the last run). Every FAIL is in `examples/` and traces to that cap; no core unit/execution test fails. The identical tree passed the same suite on the prior commit `09da869a` (the only delta since is a one-line test-hygiene change that makes no LLM calls). Re-running reproduced the same cap. This clears when the daily budget resets or an admin raises it — it is not a code issue on this PR. Core suites are green locally (JS 1086 passed, Python 653 passed).

javascript/src/agents/judge/judge-agent.ts
```diff
@@ -563,10 +564,23 @@ export class JudgeAgent extends JudgeAgentAdapter {
...
+    // invokeLLMWithDiscovery relaxes toolChoice to "required" so the model can
+    // use discovery tools, but a checkpoint/last-turn judgment is still
+    // terminal.
+    const judgmentRequired = isLastMessage || enforceJudgement;
+
+    // Pinned on judgmentRequired ALONE: on the standard path a required
+    // judgment cannot be dodged via continue_test, criteria or not — the
+    // criteria-less enforced case already returned above, and a criteria-less
+    // last-turn judge still owes its terminal verdict. Known gap: the
+    // large-trace discovery path relaxes toolChoice to "required" below and
```
langwatch-agent: **P1 — Do not let discovery bypass a required final judgment.** On a last turn or explicit `judge()` checkpoint, `toolChoice` is initially pinned, but this method overwrites it with `"required"`. An intermediate large-trace response can therefore call `continue_test`; `parseToolCalls` returns `null`, and the run reaches the generic max-turn failure instead of returning the required judge result. This is the same contract that the standard-path fix just hardened, so leaving it to #887 means the fix remains conditional on trace size. Preserve the terminal constraint for `judgmentRequired` (for example, only permit discovery before a final forced finish, without accepting `continue_test`) and add a large-trace last-turn/checkpoint regression.

python/scenario/judge_agent.py
```diff
@@ -702,12 +703,34 @@ async def call(
...
+        # is about the CONTRACT, not the transport — the discovery loop applies
+        # the pinned tool_choice only on its final step, but a checkpoint/
+        # last-turn judgment is still terminal.
+        judgment_required = is_last_message or enforce_judgment
+
+        # Pinned on judgment_required ALONE: on the standard path a required
+        # judgment cannot be dodged via continue_test, criteria or not — the
+        # criteria-less enforced case already returned above, and a
+        # criteria-less last-turn judge still owes its terminal verdict. Known
+        # gap: the discovery loop applies this pin only on its final step and
```
langwatch-agent: **P1 — The Python discovery loop has the same terminal-judgment escape.** For a last turn or `judgment_request`, `tool_choice` is correctly set to `finish_test`, but every non-final discovery iteration substitutes `"required"`. A model can choose `continue_test` there, `_parse_response` returns `[]`, and the caller falls through to max-turn handling rather than receiving the required judgment. This cannot be deferred as an inherited gap because this PR promises forced verdicts remain terminal. Keep discovery available while preventing/overriding `continue_test` for `judgment_required`, and test a large-trace last-turn and checkpoint path.

langwatch-agent (commented):
2 P1s — required judgments can still continue through the large-trace discovery path in both SDKs; see inline comments. CI’s failing example suites are attributable to the shared gateway virtual-key daily budget, not this diff.

javascript/src/agents/judge/judge-agent.ts
sergioestebance: Fixed in 77ce1dbd — closed here rather than deferred, since this PR now makes the terminal-contract promise explicitly. invokeLLMWithDiscovery now forces a final verdict when a required judgment (last turn / judge() checkpoint) answers continue_test during large-trace discovery, so it can no longer dodge into the max-turns failure. Regression test added: large-trace last-turn continue_test -> forced finish_test, terminal result.

python/scenario/judge_agent.py
sergioestebance: Fixed in 77ce1dbd — Python mirror: _run_discovery_loop forces a verdict via _force_verdict when a required judgment answers continue_test on a non-final discovery step, instead of returning [] and falling through to max-turns. Regression test added (large-trace judgment_request continue_test -> forced finish_test, terminal ScenarioResult).

drewdrewthis:
## Review verdict: READY

Run: pr-review (reviewer: @review-clerk)

Third-party review of langwatch/scenario#889 (author sergioestebance, confirmed not this session). Full diff read plus surrounding source (`judge-agent.ts`, `judge_agent.py`, `scenario-execution.ts`) checked out locally, not diff-in-isolation.

### Focus: non-termination / max-turns determinism
Traced the specific concern end-to-end:
- The executor's `currentTurn >= maxTurns` hard stop (`scenario-execution.ts`, `_step()`) is fully independent of the judge's continue/finish decision — an inconclusive-continues loop cannot bypass it. No infinite-loop risk.
- Discovery-loop exhaustion (`discoveryExhausted()` / `forceVerdict()` JS, `_force_verdict` Python) closes the trace-discovery path deterministically.
- One gap was live during this review: on a large trace, `invokeLLMWithDiscovery` relaxed `toolChoice` to `"required"`, so a required (last-turn/checkpoint) judgment could answer `continue_test` and dodge into the generic max-turns failure instead of its real verdict. The author caught and fixed this in-flight (commit `77ce1dbd`, "force a verdict when a required large-trace judgment tries to continue") in both SDKs, with new regression tests (`judge-agent.test.ts`, `test_judge_discovery_exhaustion.py`) that exercise exactly this path. Verified the fix is structurally correct: JS's `completionCalledTool(completion, "finish_test")` check and Python's `verdict_forced and terminal_call.function.name == "continue_test"` check both correctly gate on the `judgmentRequired`/`judgment_required` signal (traced `verdict_forced = judgment_required` at the Python call site). Terminal contract now holds on both the standard and large-trace paths.

### Correctness / safety
- Criteria-less last-turn judge (the one deliberate behavior change vs. main, per PR body): traced `parseToolCalls`/`_parse_response` with `criteria=[]` — `metCriteria`/`unmetCriteria` both resolve to `[]` safely, `success` driven purely by `verdict === "success"`. No crash risk.
- `enforceJudgement && !hasCriteria` still short-circuits before reaching the tool call, so the criteria-less-and-forced case is handled separately and correctly.
- No secrets, no unsafe eval, no injected user content reaching a sink — nothing security-relevant in this diff.

### Structure / design
- `judgmentRequired`/`judgment_required` and `verdictForced`/`verdict_forced` are named and threaded identically across both SDKs — good mirror discipline.
- Judge-internal state-machine logic; nothing here reinvents existing SDK/tool functionality.
- Known-adjacent defects (INCONCLUSIVE never reaching the wire, JS criteria-payload hardening parity) are explicitly out of scope, disclosed in the PR body, and already filed as #887/#888 and platform-side langwatch/langwatch#6833-6835 — correctly not bundled into this fix.

### Tests
New coverage on both SDKs: unforced inconclusive → continue (unit + execution-level, real `ScenarioExecution`), forced paths stay terminal (last turn, checkpoint, discovery exhaustion), unforced inconclusive during large-trace discovery → continue, and the large-trace required-judgment-dodges-continue fix above. All traced as testing the actual changed branches, not just re-asserting existing behavior.

### CI — RED at this sha, but the failure is environmental, not this PR
CI at head sha `77ce1dbd6e107ed08216930260cda532768a2d79` has resolved: `ci-checks (24.x)`, `javascript-complete`, `test (3.12)` and `python-complete` are FAILURE; `changes`, `Analyze (javascript-typescript)`, `Analyze (python)`, `Validate PR Title`, `docs-complete` are SUCCESS.

The failure does not implicate this change. In both SDKs the real test suites pass and only the live-LLM example suites fail:

| Job | Step | Result |
|---|---|---|
| `ci-checks (24.x)` | `Tests` | **93 files / 1088 tests passed**, 4 skipped |
| `ci-checks (24.x)` | `Test (Examples)` | 26 files / 36 tests FAILED |
| `test (3.12)` | `Run tests` | **1219 passed**, 25 skipped, 0 failed |
| `test (3.12)` | `Test (Examples)` | 26 failed, 3 passed |

Both files this PR touches on the JS side are green — `src/agents/judge/__tests__/judge-agent.test.ts` (32 tests) and `src/execution/__tests__/judge-inconclusive-continues.test.ts` — including the new large-trace regression cases.

Every example-suite failure traces to one upstream response: `402` — "The virtual key spending limit (per day) for this request has been reached", `"type":"budget_exceeded"`, `"fault":"customer"`, budget scope `virtual_key`, window `day`. The JS log alone carries 178 occurrences of that message. The failures hit tests unrelated to this change (`weather-agent`, `vegetarian-recipe-agent`, `database-tool-mocking`, `voice/*`, `test_lovable_clone`, `test_azure_api_gateway`), which confirms an account-level LLM spend cap rather than a code regression.

Action for maintainers: raise the LangWatch virtual-key daily budget, or re-run the examples jobs once the daily window rolls over. No change to this branch is required to clear it. This review is still not CI sign-off.

### Blocking findings
None. No inline review-thread comments were posted — zero unresolved threads exist, satisfying the READY test.

drewdrewthis (commented):
Third-party review complete (pr-review + review-methodology procedures). Verdict: READY. Zero blocking findings — no inline threads posted. CI still pending at review time; not treated as sign-off. Full writeup in the verdict comment: https://github.com/langwatch/scenario/pull/889#issuecomment-5252940524

drewdrewthis (approved):
lgtm

PR #906, opened 2026-08-14
fix: evaluation score should be passed / total, not passed / failed

CTWalk:
## Why

The judgment score reported to LangWatch divides the passed criteria by the
**failed** count instead of the total, so a half-failing verdict gets the same
score as a flawless one.

```
3 passed, 0 failed  ->  score = 1.0
3 passed, 3 failed  ->  score = 1.0     <-- indistinguishable
3 passed, 1 failed  ->  score = 3.0
0 passed, 3 failed  ->  score = 0.0
```

Every verdict where passes equal failures — `(1,1)`, `(2,2)`, `(3,3)` — collapses
onto a perfect 1.0, so the evaluation span can't be used to tell a healthy run
from a half-broken one.

The expression, `scenario_executor.py:1056-1061`:

```python
score=(
    len(agent_response.passed_criteria)
    / len(agent_response.failed_criteria)
    if agent_response.failed_criteria
    else 1.0
),
```

## What changed

- **Denominator is `passed + failed`, not `failed`.** That matches the
  denominator the package already uses for this same ratio elsewhere:
  `pytest_plugin.py:222-225` prints `Passed Criteria: n/total` with
  `total = len(passed_criteria) + len(failed_criteria)`. The span score and the
  terminal report now agree.
- **Kept the existing `if ... failed_criteria else 1.0` guard** rather than
  adding a separate zero-check. It already guarantees a non-zero denominator,
  and a judgment with no criteria at all reaches this line — dividing by the
  total there would be `0/0`. The new test pins that case so the guard can't be
  dropped later.
- **`passed + failed` is exactly the criteria count**, so the new denominator
  can't undercount: `JudgeAgent` puts every criterion in one of the two buckets
  (`judge_agent.py:1344-1352` — a criterion passes only on an explicit `"true"`,
  everything else is a failure).
- `passed=agent_response.success` is untouched and stays correct.

## Test plan

- New `test_evaluation_score_is_fraction_of_criteria_passed` in
  `python/tests/test_scenario_executor.py`. It captures the score by
  `monkeypatch`ing `LangWatchSpan.add_evaluation`, so no network call.
- It asserts three things: the four ratios `1.0 / 0.75 / 0.5 / 0.0`; the
  collision directly (half-failing must not score the same as flawless); and
  the zero-criteria judgment, which is what keeps the guard necessary.
- Confirmed RED on `main` — with the fix reverted and the test kept, it fails at
  `assert 3.0 == 0.75`, i.e. the 3-passed / 1-failed row above.
- `cd python && uv run pytest tests/test_scenario_executor.py` → 13 passed.
- Full suite minus the modules that need live provider credentials
  (`uv run pytest tests --ignore=tests/voice --ignore=tests/test_event_reporter.py`)
  → **668 passed, 10 skipped**. `uv run pyright` clean on both changed files.

No playable artifact — this is a metric on the evaluation span. The observable
proof is the captured score itself, before and after, for the same judgments:

| passed | failed | before | after |
|---|---|---|---|
| 3 | 0 | 1.0 | 1.0 |
| 3 | 1 | **3.0** | 0.75 |
| 3 | 3 | **1.0** | 0.5 |
| 0 | 3 | 0.0 | 0.0 |
| 0 | 0 | 1.0 | 1.0 |

The `after` column is exactly what the new test asserts. The `before` column is
what those same assertions produce when the fix is reverted — the two bold cells
are the bug: a 3-of-4 verdict scoring 3.0, and a half-failing verdict scoring a
perfect 1.0.

## Anything surprising?

- **Reports only, never a verdict.** This value goes to the evaluation span and
  nowhere else — one call site, `scenario_executor.py:1051`. It doesn't affect
  `result.success` or any assertion, so no existing test changes behaviour.
- `1 - failed / total` is algebraically the same fix — happy to switch to it if
  you'd prefer that reading.
- The TypeScript package has no equivalent call, so there's no mirror PR to
  send for this one.

### DIFF
--- a/python/scenario/scenario_executor.py
+++ b/python/scenario/scenario_executor.py
@@ -1055,7 +1055,10 @@ class ScenarioExecutor:
                         details=agent_response.reasoning,
                         score=(
                             len(agent_response.passed_criteria)
-                            / len(agent_response.failed_criteria)
+                            / (
+                                len(agent_response.passed_criteria)
+                                + len(agent_response.failed_criteria)
+                            )
                             if agent_response.failed_criteria
                             else 1.0
                         ),
--- a/python/tests/test_scenario_executor.py
+++ b/python/tests/test_scenario_executor.py
@@ -290,6 +290,47 @@ class InlineCriteriaMockJudgeAgent(JudgeAgent):
 
 
 @pytest.mark.asyncio
+async def test_evaluation_score_is_fraction_of_criteria_passed(monkeypatch):
+    """The score reported for a judgment is the fraction of its criteria that
+    passed. Dividing by the failure count instead made every verdict with as
+    many passes as failures score 1.0 — identical to a flawless one."""
+    import langwatch.telemetry.span as lwspan
+
+    scores = []
+    monkeypatch.setattr(
+        lwspan.LangWatchSpan,
+        "add_evaluation",
+        lambda self, **kw: scores.append(kw.get("score")),
+    )
+
+    async def score_for(*criteria: str) -> float:
+        scores.clear()
+        await ScenarioExecutor(
+            name="evaluation score",
+            description="test",
+            agents=[
+                MockAgent(),
+                MockUserSimulatorAgent(model="none"),
+                InlineCriteriaMockJudgeAgent(model="none", criteria=list(criteria)),
+            ],
+            script=[user("hello"), agent(), judge()],
+        ).run()
+        return scores[0]
+
+    assert await score_for("a", "b", "c") == 1.0
+    assert await score_for("a", "b", "c", "x fail") == 0.75
+    assert await score_for("a", "b", "c", "x fail", "y fail", "z fail") == 0.5
+    assert await score_for("x fail", "y fail", "z fail") == 0.0
+
+    # The collision this guards: half-failing must not equal flawless.
+    assert await score_for("a", "x fail") != await score_for("a")
+
+    # A judgment with no criteria at all keeps the existing guard — dividing by
+    # the total would be 0/0 here.
+    assert await score_for() == 1.0
+
+
+@pytest.mark.asyncio
 async def test_inline_criteria_checkpoint_pass_continues():
     """When inline criteria pass, the script should continue past the judge step."""
     call_count = {"agent": 0}


langwatch-agent (commented):
No actionable correctness, security, or integrity finding in the external diff. The denominator now represents all evaluated criteria and preserves the existing empty-criteria behavior. I performed static review only; contributor code and dependencies were not executed.

LangWatch-Review: verdict=clean sha=328d91f3d82c4660eebf2de1665cb11cb2ed27f4 p0=0 p1=0 p2=0 p3=0

langwatch-agent (commented):

LangWatch-Review: verdict=clean sha=328d91f3d82c4660eebf2de1665cb11cb2ed27f4 p0=0 p1=0 p2=0 p3=0

langwatch-agent:
> [!CAUTION]
> **A three-line fix that silently rescores every evaluation a customer has ever run, on their next SDK upgrade.**
>
> The old expression divided passed criteria by failed criteria. So `3 passed, 1 failed` reported a score of `3.0`, and `3 passed, 3 failed` reported `1.0`, identical to a flawless run. After this, those become `0.75` and `0.5`.
>
> Both of those movements break something real. Anyone with a threshold, a CI gate or an alert on evaluation score has been sitting on numbers that were mostly above `1.0` or pinned at exactly `1.0` for half-broken runs. A gate like "fail the build under 0.8" was effectively never firing and will start firing. Dashboards that plotted this on a zero-to-one axis had points off the top of the chart and will now be in range, which will read as a regression in agent quality rather than a correction in arithmetic. Nothing in the SDK will tell them which it was.
>
> That is not an argument against merging. The old number was meaningless: every verdict with as many passes as failures collapsed onto a perfect score, so the span could not distinguish a healthy run from a half-broken one, which is the only thing it is for. It is an argument that this needs a release note that says scores will move and why, and probably not a silent patch bump.

| | |
| --- | --- |
| **Mode** | `Targeted Review`. Three lines of production code. The review is entirely about the blast radius and the release, not the arithmetic. |
| **Issue** | No linked issue. The PR body carries the full derivation with a collision table, which is better evidence than most issues would have provided. |
| **State** | +45 / -1 across 2 files, of which 41 lines are the new test. CI green, `mergeable`, `BLOCKED` only for want of the required approval. Requested reviewer `sergioestebance`. Outside contributor, so nobody here is accountable for it by default. Open since 14 August. |
| **Evidence** | Complete for the change itself. The test asserts four ratios, the collision directly, and the zero-criteria case, and it was confirmed red on `main` failing at `assert 3.0 == 0.75`. The claim that `passed + failed` is exactly the criteria count is backed by a code reference: `JudgeAgent` puts every criterion in one bucket or the other, passing only on an explicit `"true"`. |
| **Where to look** | The release and its notes · the zero-criteria guard, which preserves a different unearned `1.0` · whether anything downstream has a stored threshold |

<details>
<summary><b>What you are looking at, if the Python SDK is not your area</b></summary>

ACME writes a scenario test: a simulated user talks to their agent, then a judge agent checks the transcript against a list of criteria. Each criterion either passes or fails. The SDK reports the result to LangWatch as an evaluation on a span, with a score.

The score is meant to be "what fraction of the criteria passed". It was computed as passed divided by failed.

```
3 passed, 0 failed  ->  1.0   (guard, correct by accident)
3 passed, 1 failed  ->  3.0   (above the maximum)
3 passed, 3 failed  ->  1.0   (perfect, and half the criteria failed)
0 passed, 3 failed  ->  0.0   (correct by accident)
```

| Term | What it means here |
| --- | --- |
| **Criteria** | The list of things the judge checks, one boolean each. |
| **`passed_criteria` / `failed_criteria`** | The two buckets. Every criterion lands in exactly one; passing requires an explicit `"true"`. |
| **`add_evaluation`** | The call that puts the verdict on the span, carrying `passed` and `score` separately. |
| **`passed`** | The boolean verdict, from `agent_response.success`. Untouched by this PR and always correct. |

```mermaid
flowchart LR
  J["JudgeAgent evaluates N criteria"] --> P["passed_criteria"]
  J --> F["failed_criteria"]
  P --> S["score"]
  F --> S
  S -->|"before: passed / failed"| B["3.0, or 1.0 for half-broken"]
  S -->|"after: passed / (passed + failed)"| A["0.75, or 0.5 for half-broken"]
  J --> V["passed = success, unchanged"]
```

The small thing that is load-bearing: **the `if agent_response.failed_criteria else 1.0` guard is the only reason this line never divided by zero.** It is kept deliberately, and the new test pins it so nobody deletes it as redundant later. With the new denominator it is doing different work than it was: it now exists purely for the zero-criteria case, since `passed + failed` cannot be zero any other way.

</details>

<details>
<summary><b>The decisions being ratified</b></summary>

**The denominator becomes `passed + failed`.** It matches the denominator the package already prints in the terminal report: `pytest_plugin.py` shows `Passed Criteria: n/total` with `total = passed + failed`. So the span score and the thing the developer reads on their own console now agree, which they did not before. That agreement is the strongest argument for this exact form over any other.

**The guard is kept rather than replaced with a zero check.** A judgment with no criteria at all reaches this line, and with the new denominator that would be `0/0`. Keeping the existing guard is the smaller change and the test makes it load-bearing rather than incidental.

**`passed` is left alone.** Correct. The boolean verdict was never wrong, and conflating it with the score is how these two drift.

**Scope is one expression.** Nothing else is touched. For a fix from an outside contributor to a published package, that is the right size.

</details>

<details>
<summary><b>Worth asking the author, or deciding internally</b></summary>

**Is a zero-criteria judgment really a perfect score?** The guard returns `1.0` when there are no criteria, so a judge that judged nothing reports a flawless run. That is the same shape as the bug being fixed: an unearned `1.0` that a dashboard cannot distinguish from a real one. Leaving it is defensible, since changing it is a separate decision about whether an absent score should be `None` rather than a number. It is worth writing down as a known state rather than leaving it to be rediscovered.

**What does the release look like?** The title is `fix:`, which cuts a patch. A patch that moves every customer's evaluation scores is a surprising thing to receive from a patch. Whatever the version, the release note should say plainly that scores previously above `1.0` were a bug, that half-failing verdicts previously reported `1.0`, and that both now report the true fraction.

**Does anything on the platform side have a stored threshold on this score?** Alerts, monitors, saved views or trigger conditions built while the value could exceed `1.0` were tuned against the old scale. That is a platform question rather than an SDK one, and it is cheap to check before this ships.

**The JavaScript SDK does not report an evaluation score at all.** Only the Python path emits one, so there is no matching bug to fix and no divergence introduced here. Worth knowing the two SDKs report different things on this surface, which is a separate gap somebody may want tracked.

**Nobody here owns this PR by default.** It is from an outside contributor, and it has been open since 14 August with only bot reviews on it. The assigned reviewer is the accountability, which is the point of the assignment.

</details>

<details>
<summary><b>Evidence and files</b></summary>

| File | What it is |
| --- | --- |
| `python/scenario/scenario_executor.py` | The change. Three lines, in the `score=` argument to `add_evaluation` inside `_call_agent`. |
| `python/tests/test_scenario_executor.py` | `test_evaluation_score_is_fraction_of_criteria_passed`. Captures the score by monkeypatching `LangWatchSpan.add_evaluation`, so no network call. Asserts `1.0 / 0.75 / 0.5 / 0.0`, then the collision directly, then the zero-criteria case. |

Confirmed red on `main` with the fix reverted and the test kept: it fails at `assert 3.0 == 0.75`, which is the three-passed one-failed row. `uv run pytest tests/test_scenario_executor.py` gives 13 passed, and the full suite passes minus the modules needing live provider credentials.

</details>

> [!NOTE]
> **The invariant: a score is only useful if two different outcomes cannot produce the same number.** The old expression violated that for every verdict with equal passes and failures. The fix is right, and the work left is telling people their numbers moved.

CTWalk:
One thing I should have put in the description: this changes the score value users see, not just the formula. Runs that reported above `1.0` come back into range, and half-failing runs go from `1.0` to `0.5` — so anyone with a stored threshold or alert on evaluation score will see their numbers move on upgrade. Release-please would cut a patch from the `fix:` title; happy to retitle if you'd rather it land flagged as breaking, though the version call is yours.

Also deliberate: a judgment with zero criteria still returns `1.0` through the existing guard. That's an unearned `1.0` of the same shape as the bug, but changing it is a separate decision about whether an absent score should be `None` — happy to open an issue for that separately.

One question on process: as this is from a fork, the CI runs are waiting on a maintainer to approve them, so `python-complete` hasn't reported yet. Anything you'd like me to do on my side to help that along?

langwatch-agent:
## Human Review Brief

> [!CAUTION]
> Merging this changes the numeric value of an evaluation score that is already being written to LangWatch. Every historical span carries the old formula's number, and every new one carries the new formula's. They are not comparable, and nothing in the diff marks the boundary. Any saved chart, threshold or regression baseline built on this score is measuring two different things either side of the merge commit.

| | |
|---|---|
| **Issue** | None linked, but the defect is demonstrated in the description |
| **Size** | +45/-1 across 2 files |
| **Mode** | Fast skim on the fix, targeted on the consequence |
| **Blast radius** | The `score` on every evaluation span the Python SDK emits |
| **Author** | External contributor |

<details>
<summary><b>What you are looking at, if the scenario executor is not your area</b></summary>

A scenario run is a scripted conversation driven by `ScenarioExecutor`. It steps through agents (the agent under test, a simulated user, and a judge), and at the end the **judge** returns a `ScenarioResult` carrying `success`, `reasoning`, and two lists: `passed_criteria` and `failed_criteria`. Those lists are the judgement; everything numeric downstream is derived from them.

Scenario also emits OpenTelemetry spans, and LangWatch consumes them. `span.add_evaluation(...)` is the call that attaches the judge's verdict to the run's span, and its `score` field is the number that ends up on the LangWatch simulation view and in anything aggregating across runs. The `passed` boolean and the `score` come from the same result but travel separately, so they can disagree without anything complaining.

Two things about that field make a small arithmetic change worth reading carefully rather than waving through.

- **It is an exported number, not an internal one.** Historical runs already carry whatever the old expression produced. Changing the formula changes what new runs mean relative to old ones, and nothing restates the old ones.
- **The `else 1.0` branch is doing real work.** The expression is guarded on `failed_criteria` being non-empty, so an all-pass run never evaluates the division at all. That guard is why the original could look plausible: the divide-by-zero case was handled, and the remaining cases were merely wrong rather than crashing.

One asymmetry worth knowing rather than assuming: this is Python-only. The TypeScript SDK has no `addEvaluation` call at all, so it emits no score for the judge. Parity between the two SDKs is normally a maintained property here, which makes this a gap rather than a deliberate difference, but it does mean there is no second copy of this expression to fix alongside.

</details>

<details>
<summary><b>The bug is real and the demonstration is clean</b></summary>

```python
score=(len(passed_criteria) / len(failed_criteria) if failed_criteria else 1.0)
```

Dividing passes by failures rather than by the total. The consequences fall out immediately:

| verdict | old score |
|---|---|
| 3 passed, 0 failed | 1.0 |
| 3 passed, 3 failed | 1.0 |
| 3 passed, 1 failed | 3.0 |
| 0 passed, 3 failed | 0.0 |

A half-failing verdict is indistinguishable from a flawless one, every balanced verdict collapses onto 1.0, and the score can exceed 1.0, which for something named "score" is on its own a strong signal something is off. The fix uses `passed + failed`, which is also the denominator `pytest_plugin.py` already uses when it prints `Passed Criteria: n/total`. So the package was internally inconsistent and the printed number was the correct one.
</details>

<details>
<summary><b>What to decide, since the code change is obvious</b></summary>

1. **Is this a breaking change for the SDK's semver?** The function signature does not change; the meaning of a returned number does. Different teams draw that line differently. If anyone has an alerting threshold on this score, it moves under them silently.
2. **Should the discontinuity be recorded?** A line in the changelog saying the score formula changed on this version costs nothing and saves someone a confusing afternoon in three months.
3. **Zero criteria.** Check the new expression handles `passed + failed == 0`. The old code guarded division by returning 1.0 when there were no failures, and that guard covered the empty case by accident. A new denominator needs its own guard, and "no criteria evaluated" scoring 1.0 is a defensible choice but should be a chosen one.
</details>

> [!NOTE]
> Advisory only. The hound does not review code line by line, does not approve and does not merge. Good catch from an external contributor; the code is four lines and the comparability question is the actual review.

PR #907, opened 2026-08-14
fix(voice): discard interrupted Gemini turn residue

### LINKED ISSUE #760: voice/gemini-live (TS): interrupt() leaves stale queue items that can truncate the recovery turn
### Status of this report

Found by a **pattern sweep** after fixing the equivalent bug in the Twilio adapter (#697). The code path below was established by reading the source; it has **not been reproduced at runtime** and no test exercises it. Treat the mechanism as plausible-and-specific, not confirmed — a repro test is the first task.

### The pattern

The bug fixed in Twilio: a producer enqueues a terminal signal into a **long-lived** queue; if nothing is draining at that moment, the item stays buffered and the **next** turn's first `receiveAudio` consumes it. `drainAgentResponse` breaks on an empty chunk, so the next turn truncates to silence and its real audio is stranded.

### Where it appears to recur

`javascript/src/voice/adapters/gemini-live.ts`:

- `private queue: QueueItem[] = []` (`:157`) is long-lived — reset only in `connect()` (`:237`) and `disconnect()` (`:277`).
- `interrupt()` (`:511-524`) wakes the in-flight `dequeue()` via `resolveNext`, but **never clears `this.queue`**.
- `sendAudio()` (`:295-316`) resets `lastAgentTranscript` / `iterHadAudio` — not the queue.
- The SDK's `onmessage` keeps calling `enqueue()` (`:534-542`) for the cut-off turn: trailing audio, or that turn's belated `turnComplete`.

Call site: `scenario-execution.ts:2017` calls `adapter.interrupt()` and then `adapter.sendAudio(chunk)` with no purge in between. The recovery turn's `receiveAudio` → `dequeue()` → `shift()` (`:549`) would then read the previous turn's residue first.

### Why Python is not affected

`python/scenario/voice/adapters/gemini_live.py:459-500` — `interrupt()` **actively drains** the receive iterator up to `turn_complete` (2s bound) before resetting `_recv_iter = None`. The TypeScript port did not carry this over. So this is also a PY/TS parity gap.

### Prior art in this repo

`ElevenLabsAgentAdapter` (TS) already defends against exactly this: it hands the empty terminal to an **active waiter only**, and otherwise drops it rather than queueing it (`elevenlabs.ts:894-900`), with `reconcilePendingAudio()` (`:857-868`) handling genuine leftover audio at the pre-`sendAudio` hook (#747). Twilio (#697) instead purges the queue at receive-loop entry. Either shape would work here.

### Suggested first step

A test in `gemini-live.test.ts`: interrupt a turn, deliver a late message for the old turn while nothing is draining, then start a new turn and assert its first `receiveAudio` returns the **new** turn's audio, not a stale empty chunk.

### Also surfaced by the same sweep (separate, lower priority)

- `python/scenario/voice/adapters/pipecat.py` — `_recv_loop` (`:252-288`) has no `finally` and never enqueues a terminal sentinel on any exit path, and `recv_audio` (`:232-235`) is a bare `queue.get()`. It is the one Python adapter that never adopted the #648 end-of-turn pattern: a `recv_audio` at true end-of-call times out rather than returning cleanly. Bounded (not silent truncation), but a real parity gap — the TS port has the clean version via `interruptPhase` (`pipecat.ts:334,365-368`).

ach3rry:
## Why

A late Gemini Live event can arrive after `interrupt()` has already released the old turn's receiver. Without a protocol-aware turn boundary, that event can become the recovery turn's first queue item and an empty `turnComplete` truncates the real reply.

Closes #760

## What changed

- Keep interrupted-turn cleanup armed through the old turn's terminal `serverContent` boundary, including callbacks delivered after recovery sending starts.
- Keep recovery-send protection active until all three `sendRealtimeInput()` calls succeed, so a failed send can be retried safely.
- Preserve transport and session events while removing stale turn content.
- Cover the public callback-to-adapter path with regressions for a buffered late terminal event, a post-snapshot callback, and a failed-send retry.

## How it works

`interrupt()` records two independent facts: the old turn's terminal boundary is still pending, and a complete recovery send is still pending. Buffered server content is removed before each recovery attempt. Newly delivered server content is discarded until the old turn's `turnComplete` arrives; later messages are then eligible for the recovery receiver. Recovery-send state is cleared only after `activityStart`, audio, and `activityEnd` are all sent successfully.

The existing abort sentinel still releases any in-flight receiver immediately.

## Test plan

- Red on `c9b58bbf`: the original regression fails with `expected 0 to be greater than 0`.
- Red on prior PR head `b8bb005`: the post-snapshot callback and failed-send retry regressions both fail with `expected 0 to be greater than 0`.
- `pnpm exec vitest run src/voice/adapters/__tests__/gemini-live.test.ts` — 17 passed.
- `pnpm exec vitest run` — 1119 passed, 4 skipped.
- `pnpm run typecheck`
- `pnpm exec eslint src/voice/adapters/gemini-live.ts src/voice/adapters/__tests__/gemini-live.test.ts`
- `git diff --check`

No playable artifact is needed for this queue-lifecycle fix. The offline regressions drive the real adapter boundary through the mocked Gemini SDK callback and cover the interrupted-turn → late terminal event → recovery-turn sequence, including the two asynchronous and retry boundaries identified during review.


langwatch-agent:
## Human Review Brief

**Mode: Targeted Review.** Closes #760. Two files, and the whole change is a protocol-aware turn boundary. Scope matches the issue exactly.

### What you are looking at, if the voice subsystem is not your area

### Decisions being ratified

1. **Interrupted-turn cleanup stays armed until the old turn's terminal `serverContent` arrives**, including callbacks delivered after recovery sending has already started. The boundary is the protocol event, not the local `interrupt()` call. That is the right seam and it is the thing to get right.
2. **Recovery-send protection holds until all three `sendRealtimeInput()` calls succeed**, so a failed send is retryable. Three is a magic number from the Gemini Live protocol; it becomes a constant this code depends on.
3. **Transport and session events are preserved while stale turn content is dropped.** The filter distinguishes classes of event rather than clearing the queue, which is what keeps a recovery from losing connection state.

### Must Check

- **The old turn's `turnComplete` never arriving.** If the transport dies between `interrupt()` and the terminal boundary, cleanup stays armed forever and every subsequent event is discarded. Confirm there is a release path on transport close or error, not only on the happy terminal event.
- **The three-call assumption.** If the protocol or the adapter ever sends a different number, the recovery-send guard either releases early or never. Worth a named constant with a comment rather than a bare `3`, if it is not already.

### Ask Author

Is the buffered-late-terminal-event regression driven through the public callback-to-adapter path, or does it reach in? The body says the public path, which is what makes the test worth having.

`Probably Fine`: 181 added lines, two files, three named regressions covering the reported failure and two adjacent ones.

langwatch-agent (commented):
No additional actionable correctness, security, or integrity finding in the external diff. The cleanup remains armed across a failed recovery send and filters both buffered and late server-content residue. I performed static review only; contributor code and dependencies were not executed.

LangWatch-Review: verdict=clean sha=afb55fb18d108b0066b1611edec9bd74cfe5330f p0=0 p1=0 p2=0 p3=0

ach3rry:
Thanks for the focused review. I confirmed the three points:

- The cleanup gate only drops items carrying `serverContent`. `onerror` and `onclose` are enqueued unchanged; `receiveAudio()` surfaces the error or the end-of-turn sentinel, and `connect()` resets all interruption and recovery flags before the next session. A missing old `turnComplete` therefore cannot hide transport termination or leak cleanup state into a reconnected session.
- There is no numeric three-call counter. `sendAudio()` explicitly sends `activityStart`, audio, and `activityEnd` in sequence, and clears recovery state only after all three calls return. A throw leaves the state armed for retry.
- The late-boundary regression uses the SDK `onmessage` callback captured from `live.connect()` and schedules it with `queueMicrotask` during `sendRealtimeInput()`; it does not reach into adapter internals.

The subsequent static review is clean, so I have not added extra state or a synthetic numeric constant.

langwatch-agent (commented):

LangWatch-Review: verdict=clean sha=afb55fb18d108b0066b1611edec9bd74cfe5330f p0=0 p1=0 p2=0 p3=0

langwatch-agent:
> [!NOTE]

<details open>
<summary><b>What you are looking at, if the voice path is not your directory</b></summary>

Scenario's voice support drives a realtime model over a streaming connection: audio frames go up, audio and events come back, and the library turns that stream into the turn structure the rest of scenario reasons about. A **turn** is one participant speaking until they stop.

The hard part is that the stream does not announce turns cleanly. Frames arrive continuously, silence is not a message, and an interruption means the model was mid-turn when the user started talking again. So the adapters carry a small state machine that decides where one turn ends and the next begins, driven by events like speaking-started and speaking-stopped plus the frames themselves.

That state machine is where nearly every voice defect lives, and it fails in a characteristic way: not a crash, but a turn boundary in the wrong place, which surfaces much later as a judge reading a transcript that does not match what was said. When reading a voice fix, the question is almost always "what does this do to the boundary", not "does this code work".
</details>

langwatch-agent (commented):
Static review complete: no blocking correctness or security findings in the current queue-boundary change. I did not execute this untrusted fork's code or tests. Residual risk is live Gemini callback ordering beyond the mocked adapter boundary.\n\nLangWatch-Review: verdict=clean sha=0562e0d02dbfc1f51176d43d30f2757fb64fe00f p0=0 p1=0 p2=0 p3=0

PR #909, opened 2026-08-16
feat: judge fetches remote traces with per-turn trace propagation

### LINKED ISSUE #910: Span-collection error reason says still incomplete when only infrastructure-logged LLM spans arrived
When the agent under test does not adopt the propagated trace context, the turn trace can still contain LLM spans that scenario-side infrastructure (a gateway or logging proxy) wrote into it. Those spans are not in the local collector, so they count as remote, and the settle deadline reports the trace as "still incomplete" instead of "no agent spans arrived".

The judge outcome stays correct: trace-dependent criteria go inconclusive or fail either way. But the reason string points the operator at ingestion lag when the real problem is that the agent never joined the trace.

Observed in the platform falsification runs for langwatch/scenario#909 (trace context stripped on the agent side; the gateway-logged user simulator and judge LLM spans on the turn trace counted as remote).

Idea: classify spans from the scenario's own model-call path (gateway metadata, span attributes the SDK already stamps) as infrastructure when deciding between the two deadline reasons, so "no agent spans arrived" fires when only such spans are present.

rogeriochaves:
## What

When the agent under test runs behind an HTTP endpoint, the judge saw only the transcript. Tool calls, database writes, and retrievals happened on the remote side, so criteria about internal behavior were unverifiable. Customer report: teams that test deployed agents need the judge to verify real behavior, not transcript claims.

This PR makes trace propagation and remote trace fetching SDK capabilities in both languages:

- `AgentInput` exposes W3C trace propagation headers built from the active turn context: `propagation_headers` (Python) and `propagationHeaders` (TypeScript). An HTTP adapter forwards them, and the remote agent that adopts them writes its spans into the same trace the conversation messages already reference.
- Opt-in config enables fetching: `fetch_remote_traces` + `trace_wait_timeout` (Python), `fetchRemoteTraces` + `traceWaitTimeoutMs` (TypeScript). The judge fetches every turn's trace from `GET {LANGWATCH_ENDPOINT}/api/trace/{id}`, filters scenario infrastructure spans, dedupes against locally collected spans, and feeds the rest into its existing trace digest with progressive discovery.

## Two-phase judge

The judge now runs in two phases, remote traces on or off:

1. **Decision call** (mid-conversation): decides only whether the conversation has collected enough information. Two argument-free tools, `continue_test` and `make_verdict`. No reasoning field and no per-criterion schema, so the judge cannot pre-commit to pass or fail before it sees the full evidence. No remote fetching and no waiting. Below the `min_turns` floor the call is skipped entirely.
2. **Verdict call**: entered from a `make_verdict` decision, the last turn, or an explicit judge step. With remote fetching on, this is the only fetch site: the judge settle-waits once, then delivers the verdict with the full trace digest and `finish_test` pinned.

Conversation flow returns to normal multi-turn behavior: no early-finish pressure, no mid-conversation fetch rounds, no second-verdict re-invocation machinery.

## Settle semantics

- A trace settles when it holds at least one remote span (not one of the scenario's own locally collected spans) and every fetched agent span's parent resolves within the fetched plus local spans. Span-count stability is deliberately not a signal: ingestion arrives in chunks that can be tens of seconds apart.
- A failed poll retries until the deadline instead of failing the trace. Each request has a fixed 30 second bound; the loop owns the deadline.
- At the deadline, the judge keeps every span that arrived plus a synthetic `langwatch.span_collection.error` span whose reason separates "still incomplete" from "no agent spans arrived". The judge rule maps those to inconclusive, never passed from transcript claims.
- A voluntary inconclusive verdict continues the conversation (#886), except when no trace of the run ever settled: then it is terminal, because more turns cannot produce the missing evidence.
- Spans this process itself started never count as remote evidence. The judge span collector registers every span at start, and the fetcher treats a platform echo of one of them as local, even when its ancestor chain crosses the still-open turn span or its instrumentation carries no thread id. A trace holding only echoes now reports "no agent spans arrived" instead of "still incomplete" (fixes #910).
- The default wait budget is 30 seconds (was 60). When the settle-wait ends with incomplete traces, the verdict call offers a one-shot `wait_for_traces` tool: calling it re-arms the failed traces, retracts their synthetic error spans, and settle-waits once more under `trace_wait_extension` / `traceWaitExtensionMs` (default: the wait budget; the platform passes its 30s cap). The tool is then withdrawn, so the re-entered verdict must decide.

## Specs and tests

- `specs/remote-trace-fetching.feature` and `specs/judge-two-phase.feature`, all scenarios bound to tests in both SDKs.
- Full suites green in both SDKs, pyright and typecheck clean.
- Proven live end to end against a LangWatch platform run with an MCP-backed HTTP agent: golden runs pass with span-citing reasoning, and falsification runs (trace context stripped, or wrong account parameter) fail exactly the trace-dependent criteria. Evidence and screenshots are on the platform PR.

## Related

- ADR: `docs/adr/006-remote-trace-fetching.md`
- Docs: `docs/docs/pages/advanced/remote-traces.mdx`, `how-judging-works.mdx`, blackbox testing and configuration pages
- Platform adoption PR: langwatch/langwatch#7070 (deletes the platform's private fetch wrapper, passes a per-project wait budget)

Fixes #910
- Demo agent used for the live proof: langwatch/altura-demo#TBD_ALTURA


javascript/src/tracing/remote-trace-fetcher.ts
```diff
@@ -0,0 +1,490 @@
...
+    auth: TraceApiAuth;
+  }): Promise<LangWatchApiSpan[]> {
+    const url = `${auth.endpoint.replace(/\/$/, "")}/api/trace/${traceId}`;
+
+    const response = await this.fetchFn(url, {
+      headers: {
+        // Dual-emit auth: /api/trace historically authenticates with
+        // X-Auth-Token, while newer deployments accept a Bearer token.
+        // Sending both keeps the fetch working across LangWatch versions.
+        "X-Auth-Token": auth.apiKey,
```
langwatch-agent: **P1 — Preserve `LangwatchConfig.projectId` on the trace fetch.** `LangwatchConfig` explicitly supports bearer/PAT credentials that require `X-Project-Id`, and the event reporter sends that header, but this new fetch path reduces its auth state to endpoint/key and sends neither the project ID nor Basic auth. Because `Authorization: Bearer` is present, the platform auth middleware resolves that branch first and a PAT/org key without `X-Project-Id` is rejected; remote evidence is then silently replaced by the synthetic error span. Carry `projectId` in `TraceApiAuth` and emit `X-Project-Id` when configured (with a regression test using a bearer/PAT-style configuration).

javascript/src/tracing/remote-trace-fetcher.ts
```diff
@@ -0,0 +1,490 @@
...
+        Authorization: `Bearer ${auth.apiKey}`,
+      },
+      // A fixed per-request bound against a hung API. The settle loop owns
+      // the shared deadline: shrinking this signal to the remaining budget
+      // aborted in-flight polls at the deadline edge, which recorded a
+      // misleading hard-failure reason instead of the timeout reason.
+      signal: AbortSignal.timeout(REQUEST_TIMEOUT_MS),
```
langwatch-agent: **P1 — Enforce the shared deadline while an API request is in flight.** The advertised `traceWaitTimeoutMs` is a total verdict budget, and the platform caller intentionally clamps it down to 10 seconds for low-lag projects. This fixed 30-second `AbortSignal` is not raced with the deadline, so one stalled request makes a 10-second run wait 30 seconds before the loop can observe `remaining <= 0`. The Python implementation has the same issue with its fixed 10-second client timeout. Bound each request by the remaining shared budget (rounding to a positive integer for `AbortSignal.timeout`), and add a slow/hanging-fetch test for a smaller configured budget in both SDKs.

langwatch-agent (commented):
Found two blocking issues in the current head. Both affect the new remote-trace handoff used by langwatch/langwatch#7070. This was a static review; I did not execute PR code locally.

LangWatch-Review: verdict=findings sha=cf3d628845011648901ad04d39fd35ae31fd3d30 p0=0 p1=2 p2=0 p3=0

javascript/src/tracing/remote-trace-fetcher.ts
rogeriochaves: Fixed in d3b93898. `TraceApiAuth` now carries `projectId`, resolved from the run `langwatch` config first and then `LANGWATCH_PROJECT_ID`, and the fetch sends `X-Project-Id` whenever it is set. That matches what the event reporter already does, so a PAT or organization key now reaches the trace API scoped the same way on both paths. Two tests cover it: the header is present with a bearer-style config that names a project, and absent when no project is configured. The Python fetcher already sent the header and is unchanged.

javascript/src/tracing/remote-trace-fetcher.ts
rogeriochaves: Fixed in d3b93898, in both SDKs.

Each request is now bounded by whichever is nearer, the fixed request guard or what is left of the shared deadline, so a stalled request can no longer outlive a short budget: a 10 second run gives up at 10 seconds instead of waiting the full request timeout. The computed bound is rounded up before it reaches a timer, which keeps the fractional budget case (1.25 * p95 + 5s) working.

The abort runs on the fetcher own `AbortController`, and the timer is cleared in a `finally` that also covers the response body read, so a server that stalls mid-body is bounded too. Python needed nothing extra there because `asyncio.wait_for` already wraps the whole coroutine.

On the reason wording you raised before: the abort raises a `DeadlineAbortError` sentinel that the settle loop excludes from failure-reason selection, so a run that only ran out of time still reports "no agent spans arrived" instead of a fetch that "kept failing". Hanging-fetch tests with a short budget were added in both SDKs.

javascript/src/domain/core/config.ts
```diff
@@ -23,6 +23,21 @@ export const scenarioProjectConfigSchema = z
...
+    /**
+     * Project-wide default for `ScenarioConfig.traceWaitTimeoutMs`. A per-run
+     * value on the scenario config wins over this default.
+     */
+    traceWaitTimeoutMs: z.number().positive().optional(),
+    /**
+     * Project-wide default for `ScenarioConfig.traceWaitExtensionMs`. A
+     * per-run value on the scenario config wins over this default.
+     */
+    traceWaitExtensionMs: z.number().positive().optional(),
```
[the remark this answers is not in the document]
rogeriochaves: Fixed in d3b93898, both ends.

Python: `allow_inf_nan=False` on both `Field(gt=0)` declarations, so `trace_wait_timeout` and `trace_wait_extension` reject positive infinity, which `gt=0` accepted on its own.

JavaScript: `ScenarioConfig` is a plain interface with no runtime validation, so the judge now filters each budget source before use. A value that is not a finite positive number is dropped with a warning and the next source is used, ending at the 30 second default. That covers `Infinity`, `NaN`, zero and negatives, from the per-run config and from the project config.

You were right that `domain/core/config.ts` needed no change: Zod already rejects non-finite numbers there. Tests were added for both SDKs.

python/tests/test_remote_trace_fetcher.py
```diff
@@ -0,0 +1,921 @@
...
+
+        assert elapsed < 5.0
+        [error_span] = collector.get_spans_for_thread(THREAD_ID)
+        reason = str(
+            dict(error_span.attributes or {})[
+                "langwatch.span_collection.error.reason"
+            ]
+        )
+        assert "no agent spans arrived" in reason
+
```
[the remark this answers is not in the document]
rogeriochaves: Fixed in 9950bdf3. Added `assert "aborted" not in reason`. You are right that it was the weaker assertion: the deadline sentinel exists precisely so the reason does not read as a failing fetch, and the test passed without checking that half. It now matches `TestDeadlineMidPoll` and the JavaScript test.

python/tests/test_judge_force_verdict_hardening.py
```diff
@@ -203,6 +203,50 @@ def fake_completion(**kwargs):
...
+        tools = [
+            {"type": "function", "function": {"name": "expand_trace"}},
+            {"type": "function", "function": {"name": "wait_for_traces"}},
+            {"type": "function", "function": {"name": "finish_test"}},
+        ]
+
+        captured: dict = {}
+
+        def fake_completion(**kwargs):
+            captured.update(kwargs)
```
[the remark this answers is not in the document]
rogeriochaves: Fixed in 9950bdf3. `captured` is now `Dict[str, object]`, `fake_completion` is annotated `(**kwargs: object) -> SimpleNamespace` (the helper returns a SimpleNamespace, not a MagicMock), and the captured tools are cast to `List[Dict[str, Any]]` before indexing. Scoped pyright is clean on the file.

langwatch-agent:
## Human Review Brief

**Mode: Deep Review.** Closes #910 (a span-collection error-reason bug). That issue is one line of the diff; the PR is a 10,325-line SDK capability in two languages. The linked issue does not describe the change, so treat the body as the spec and #910 as a rider.

### Decisions being ratified

1. **Trace propagation becomes public SDK surface in both runtimes**: `propagation_headers` / `propagationHeaders` on `AgentInput`, built from the active turn context. Once released the header set, the naming split and the "built from the active turn" semantics are a contract in two languages that must move together.
2. **The judge fetches over the network, opt-in, with a timeout.** `fetch_remote_traces` + `trace_wait_timeout` make the judge's verdict depend on an HTTP call to `GET {LANGWATCH_ENDPOINT}/api/trace/{id}` completing in time. A judge that is sometimes network-bound is a different thing from one that is not, and every flake from here has a new candidate cause.
3. **The judge runs in two phases.** That is a control-flow change to the component every scenario's verdict comes from, and it applies to runs that never enable fetching.
4. **Scenario infrastructure spans are filtered and remote spans deduped against local ones.** The dedupe key is the load-bearing detail: too loose and real remote work vanishes, too tight and the digest doubles.

### Must Check

- **The dedupe key between locally collected and remotely fetched spans.** This is the single place the feature silently produces a wrong judgement rather than an error. Confirm it is identity-based (span id) rather than heuristic, and what happens when the remote agent adopted the trace but not the span ids.
- **Timeout behaviour on the verdict.** When `trace_wait_timeout` expires with partial spans, does the judge return a verdict on what it has, or fail closed? Given #910 is about an error reason that lied about completeness, a partial fetch that reads as complete is the same bug one layer out.
- **Two-phase judge on runs with fetching disabled.** Confirm the phase split is a no-op for them, with a test that would fail if it were not. Most users will never enable this and all of them get the new control flow.

### Ask Author

- Does the propagation header set leak anything about the scenario run to the remote agent that a customer would not expect to send?
- 10,325 lines across 49 files. Which parts are generated or vendored, so review effort lands on the judge and the dedupe rather than on scaffolding?

### Production risk

The judge decides pass or fail. Making its input depend on a network fetch introduces a failure mode where a scenario's verdict changes because a trace was slow, not because the agent behaved differently. The opt-in flag is the right containment; the two-phase change is the part that reaches everyone.

python/scenario/scenario_executor.py
```diff
@@ -201,6 +216,14 @@ def __init__(
                       Overrides global configuration for this scenario.
             debug: Whether to enable debug mode with step-by-step execution.
                   Overrides global configuration for this scenario.
+            fetch_remote_traces: Whether the judge fetches the traces the
+                  agent under test reported to LangWatch for this
+                  conversation's trace ids (default: False). Requires the
+                  agent adapter to forward ``AgentInput.propagation_headers``.
+                  Overrides global configuration for this scenario.
+            trace_wait_timeout: Maximum seconds the judge waits at verdict
+                  time for remote traces to arrive and stabilize (default: 60).
```
drewdrewthis: **Fix:** Stale default in docstring. This says `trace_wait_timeout` "default: 60", but the actual runtime default is 30s (`DEFAULT_TRACE_WAIT_TIMEOUT_SECONDS = 30.0` in `_tracing/remote_trace_fetcher.py`), and the PR description itself says "The default wait budget is 30 seconds (was 60)." The same stale text also appears at line 2115 in the module-level `run()` docstring. Please update both to say 30.

drewdrewthis (changes requested):
Formalizing blocking status: see the inline thread on python/scenario/scenario_executor.py:225 (stale `trace_wait_timeout` default docstring, contradicts the PR's own description). Full verdict in the review-verdict comment below.

drewdrewthis:
## Review verdict: NOT-READY

### Blocking
1. `python/scenario/scenario_executor.py:225` and `:2115` — stale `trace_wait_timeout` docstrings say "default: 60"; the actual runtime default is 30s (`DEFAULT_TRACE_WAIT_TIMEOUT_SECONDS = 30.0` in `_tracing/remote_trace_fetcher.py`), and the PR's own description says "The default wait budget is 30 seconds (was 60)." Inline thread posted on line 225.

### Non-blocking
- `RemoteTraceFetcher._state()` (Python) does a check-then-act on the per-thread dict outside `self._registry_lock` in `_states_for_thread()`/`_state()`. Under asyncio's single-threaded cooperative model this is not exploitable (no `await` between the check and the mutation), so not blocking, but worth a comment noting the lock is decorative there.
- `wait_extension_available` in `judge_agent.py` is not additionally gated on `evidence_exhausted`/`verdict_is_terminal`. Reads as intentional (the tool is offered whenever there's still budget and unsettled traces), not a bug — flagging in case that was accidental.
- Investigated the PR description's claim "An HTTP adapter forwards them" against a possible built-in HTTP adapter class — none exists in the SDK, but `docs/docs/pages/advanced/remote-traces.mdx` correctly documents this as a user-authored adapter pattern. No doc/code overclaim; false lead, no action needed.
- TypeScript side (`remote-trace-fetcher.ts`, `judge-agent.ts`, `config.ts`) mirrors the Python settle semantics and default (`DEFAULT_TRACE_WAIT_TIMEOUT_MS` = 30000) correctly — no equivalent stale-docstring defect found there.
- CI: all gating checks green (python-complete, test (3.12), javascript-complete, ci-checks (24.x), build (24.x), docs-complete, CodeQL analyze x2, CodeRabbit, Validate PR Title) as of head SHA.

python/scenario/scenario_executor.py
rogeriochaves: Valid, fixed in 9ec55a4d. Both sites now say 30, matching `DEFAULT_TRACE_WAIT_TIMEOUT_SECONDS = 30.0` and the PR description.

I swept for the same staleness rather than only patching the two lines you named. The `.mdx` source pages under `docs/docs/pages` are already correct. The remaining "default 60" hits are all under `docs/docs/dist`, which is gitignored build output, so they are a stale local artifact and not part of the repo. The TypeScript side matches too, as you noted.

I also took your non-blocking note on `RemoteTraceFetcher._state`. Rather than adding a comment saying the lock is decorative, I made the lock cover both steps: `_states_for_thread` took the lock to fetch the per-thread dict and released it before the membership test, so two callers could both miss and both create, and the loser state object would be dropped while its caller still held a reference to it. It is one `with self._registry_lock` over the `setdefault` pair now. Checked first that no caller reaches `_state` while already holding the lock, since `threading.Lock` is not reentrant.

69 tests pass across the fetcher, judging, two-phase, forced-verdict and span-collector suites; pyright adds no error on either file.

rogeriochaves:
Merged with an admin override: CI was fully green and both review items were fixed in 9ec55a4d (docstring defaults corrected repo-wide, and the state lock made to actually cover the check-then-create). @drewdrewthis thank you for the review. If anything reads wrong on a second look, a follow-up PR goes out.

PR #916, opened 2026-08-17
fix: require a langwatch SDK that does not exit its host on a signal

0xdeafcafe:
## What

Raises the `langwatch` floor from `>=0.16.1` to `>=1.5.0` (upper bound unchanged) in both `package.json` and `javascript/package.json`, and updates the lockfile. Resolved version goes **0.37.0 → 1.5.0**.

## Why

Every `langwatch` release below 1.5.0 registers signal handlers that call `process.exit(0)` in a `finally` once their own OTel flush completes. From 0.37.0, the version this repo's lockfile actually resolved:

```js
process.on("SIGINT",  () => { void gracefulShutdown({ signal: "SIGINT",  exitAfter: true }); });
process.on("SIGTERM", () => { void gracefulShutdown({ signal: "SIGTERM", exitAfter: true }); });
// ... finally { if (exitAfter) process.exit(0) }
```

Node runs *every* listener registered for a signal, so this ends the process out from under everybody else. On pod termination:

- a scenario run in flight is cut off mid-turn;
- any other `SIGTERM` listener still draining is abandoned;
- and the process reports **exit 0** — success — for work that never finished.

`0.16.1`, the old floor, is worse still: it registers only `SIGTERM`, so a local Ctrl+C never flushes at all.

`langwatch@1.5.0` fixed this. It flushes, removes its own listeners, and then either leaves the process to the remaining listeners or re-raises the signal so Node applies the default action — reporting the signal (128+n) rather than faking a 0. Exiting is now opt-in behind `UNSAFE_exitProcessAfterAutoShutdown`.

This matters for this library in particular because `setupScenarioTracing` discards the handle `setupObservability()` returns and never calls `shutdown()` itself, so **the SDK's auto-shutdown is this library's entire shutdown path.** Raising the floor is the fix; there are no call-site changes.

## Why a floor bump and not just a lockfile update

The range already permitted 1.5.0, so this could have been a lockfile-only change. It shouldn't be: with a `>=0.16.1` floor, any consumer resolving from a stale lockfile or an older transitive constraint silently gets a host-killing SDK back. The floor is what makes that unrepresentable.

## Verification

Against `langwatch@1.5.0`:

- `pnpm typecheck` — clean
- `pnpm build` — clean (CJS + ESM + DTS)
- `pnpm smoke:dist` — CJS and ESM dist both load
- `pnpm lint:lib` — clean
- `pnpm test` — **1116 passed**, 4 skipped

Every export this library uses is unchanged across the 0.x → 1.x boundary — `setupObservability`, `SetupObservabilityOptions`, `LangWatchTraceExporter`, `attributes`, `getLangWatchTracer` — and `setupObservability(options?): ObservabilityHandle` keeps its signature.

## Note for the LangWatch platform

This PR does **not** by itself change what the LangWatch app runs. That app does not consume this package from npm — it vendors `platform/app/vendor/langwatch-scenario-1.1.0.tgz`, whose own `langwatch` dependency resolves to **1.0.0**, which still exits its host. `langwatch` is a regular dependency here, not a peer dependency, so the vendored copy does not inherit the app's own SDK version. Picking this up there needs a re-vendored tarball (or dropping the vendoring) once this lands and a release is cut.


sergioestebance (changes requested):
This PR has merge conflicts with the base branch. Please rebase and resolve them so we can proceed with the review.

PR #919, opened 2026-08-18
perf(vad): frame audio through a Uint8Array view, not a number[] splice (#525)

### LINKED ISSUE #525: perf(vad): replace number[] ring-buffer with Uint8Array offset tracking
## Description

\`javascript/src/voice/vad.ts:111\` declares \`private buf: number[] = []\` as the ring buffer for incoming audio frames. \`process()\` iterates \`AudioChunk.data\` (a \`Uint8Array\`) byte-by-byte and pushes into \`number[]\`, then \`classifyFrame\` takes \`number[]\`.

## Context

- The surrounding type system uses \`Uint8Array\` (\`AudioChunk.data: Uint8Array\`). The \`number[]\` is a scatter copy that fights the typed-array model.
- \`splice(0, BYTES_PER_FRAME)\` inside the hot loop is O(n) — for a 30 ms frame at 24 kHz that's 1440 elements shifted on every frame iteration.

Not a correctness bug, but a performance smell in code added in PR #515.

## Scope

Replace the \`number[]\` ring-buffer with one of:
- A single \`Uint8Array\` plus offset/length tracking.
- Concatenate-then-slice using \`Uint8Array.set()\`.
- A circular-buffer abstraction (probably overkill at 191 lines).

Measure frame-processing latency before/after on a representative clip.

Surfaced by /review on PR #515. Single reviewer: hygiene.

Related: #516, #515.

---
*Classification: Refactor*
*Status: stub — investigation pending. Issue is not-ready until \`/investigate\` completes.*

langwatch-agent:
## Ticket

Closes #525.

`vad.ts:111` held incoming audio in `private buf: number[]`. `process()` unpacked a `Uint8Array` into it one `push` at a time, then took each frame with `slice()` and dropped it with `splice(0, BYTES_PER_FRAME)`, which shifts every remaining element once **per frame**.

## Change

The buffer is a `Uint8Array` with a length. Frames are read as `subarray` views, which copy nothing, and the remainder is compacted once per `process()` call instead of once per frame. `classifyFrame` takes a `Uint8Array`; its arithmetic is untouched, since a `Uint8Array` element is already the 0-255 byte the old code read.

Capacity starts at two frames and doubles only if a single chunk exceeds it, which the steady state never does.

## Measurement

The ticket asks for before and after on a representative clip. 60 seconds of 24 kHz PCM16 delivered in 20 ms chunks, which is the shape a realtime transport actually emits and deliberately does not divide the 30 ms frame, so the buffer always carries a remainder. Median of 7 runs, same machine, same process:

| | median |
|---|---|
| `number[]` + `splice` | **18.2 ms** |
| `Uint8Array` + offset | **3.3 ms** |

Runs before: 17.9, 18.0, 18.2, 18.2, 19.4, 20.3, 23.7. After: 3.1, 3.2, 3.2, 3.3, 3.4, 4.0, 5.1.

## Evidence

**The framing had no direct coverage.** The existing `vad-fallback.test.ts` drives the one-shot fallback warning, not what `process()` does with bytes. So ten tests were written **against the original `number[]` implementation** and pass unchanged against this one. That is what makes this a refactor rather than a rewrite, and it is checkable: the test file applies cleanly to `main`.

**Five mutations, five caught:**

| # | Mutation | Fails |
|---|---|---|
| M1 | never compact, so consumed bytes are re-read | the re-classification test |
| M2 | drop the leftover instead of carrying it | the split-frame test |
| M3 | off-by-one on the frame boundary (`>` for `>=`) | five tests |
| M4 | drop the PCM16 sign extension | the quiet-negative test |
| M5 | `reserve` never grows the buffer | two tests |

**Two of those escaped on the first pass, and the tests were wrong, not the mutations.** Worth stating because both were tests I had written believing they covered exactly the thing they did not:

- The sign-extension test used amplitude `-8000`. Read signed that is loud; read unsigned as `57536` it is also loud. It passed with or without the sign extension. The discriminating fixture is `-1`: signed it is an RMS of 1 and silent, unsigned it is `65535` and the loudest sample there is. The test now uses `-1` and asserts silence.
- Every framing test used uniform audio, so re-reading consumed bytes produced the same verdict and M1 passed. Alternating loud and silent frames makes the reprocessing observable as extra speech transitions, which is also what an uncompacted buffer would leak.

**Tests.** 487 pass across `src/voice` (49 files, 1 skipped).

**Types.** `tsc --noEmit` reports 1 error on this branch and 1 on `main`, measured in a clean `main` worktree against the same `node_modules`. It is `TS2688 Cannot find type definition file for 'yauzl'` on both, and names no file in this diff.

**Lint.** `eslint` exits 0 on both changed files.

## Deployment impact

None. Internal buffering of an SDK-side detector; the public surface (`process(chunk)`, `onSpeechStart` / `onSpeechEnd`, `isSpeaking`) and the detection thresholds are unchanged.

---

### DIFF
--- a/javascript/src/voice/__tests__/vad-framing.test.ts
+++ b/javascript/src/voice/__tests__/vad-framing.test.ts
@@ -1,0 +1,211 @@
+/**
+ * Framing and hysteresis behaviour of the SDK-side VAD fallback.
+ *
+ * The ring buffer had no direct coverage: the existing suite drives the
+ * one-shot fallback warning, not what `process()` does with bytes. These
+ * tests were written against the original `number[]` implementation and pass
+ * unchanged against the `Uint8Array` one, which is what makes the swap a
+ * refactor rather than a rewrite.
+ *
+ * Everything below is expressed in whole frames or deliberate fractions of
+ * one, because the only thing that should decide when a frame is classified
+ * is how many bytes have arrived, never how they were chopped into chunks.
+ */
+import { describe, expect, it } from "vitest";
+import { AudioChunk, PCM16_SAMPLE_RATE } from "../audio-chunk";
+import { WebRTCVadFallback } from "../vad";
+
+const FRAME_MS = 30;
+const SAMPLES_PER_FRAME = (PCM16_SAMPLE_RATE * FRAME_MS) / 1000;
+const BYTES_PER_FRAME = SAMPLES_PER_FRAME * 2;
+
+/** PCM16 little-endian bytes for `sampleCount` samples all at `amplitude`. */
+function pcm16(amplitude: number, sampleCount: number): Uint8Array {
+  const bytes = new Uint8Array(sampleCount * 2);
+  for (let i = 0; i < sampleCount; i++) {
+    const value = amplitude < 0 ? amplitude + 0x10000 : amplitude;
+    bytes[i * 2] = value & 0xff;
+    bytes[i * 2 + 1] = (value >> 8) & 0xff;
+  }
+  return bytes;
+}
+
+const chunkOf = (bytes: Uint8Array): AudioChunk =>
+  ({ data: bytes }) as AudioChunk;
+
+/** Loud enough to clear the default RMS threshold of 500. */
+const LOUD = 8000;
+const SILENT = 0;
+
+function makeDetector(options: { hysteresisFrames?: number } = {}) {
+  const starts: number[] = [];
+  const ends: number[] = [];
+  const vad = new WebRTCVadFallback("test-adapter", {
+    ...options,
+    onSpeechStart: () => starts.push(1),
+    onSpeechEnd: () => ends.push(1),
+  });
+  return { vad, starts, ends };
+}
+
+describe("WebRTCVadFallback framing", () => {
+  describe("given fewer bytes than one frame", () => {
+    describe("when the chunk is processed", () => {
+      it("classifies nothing, because a partial frame is not a frame", () => {
+        const { vad, starts } = makeDetector({ hysteresisFrames: 1 });
+
+        vad.process(chunkOf(pcm16(LOUD, SAMPLES_PER_FRAME - 1)));
+
+        expect(vad.isSpeaking).toBe(false);
+        expect(starts).toHaveLength(0);
+      });
+    });
+  });
+
+  describe("given a frame split across two chunks", () => {
+    describe("when the second chunk completes it", () => {
+      it("classifies once the bytes are all in, not once per chunk", () => {
+        const { vad, starts } = makeDetector({ hysteresisFrames: 1 });
+        const frame = pcm16(LOUD, SAMPLES_PER_FRAME);
+
+        vad.process(chunkOf(frame.slice(0, 10)));
+        expect(vad.isSpeaking).toBe(false);
+
+        vad.process(chunkOf(frame.slice(10)));
+
+        expect(vad.isSpeaking).toBe(true);
+        expect(starts).toHaveLength(1);
+      });
+    });
+  });
+
+  describe("given several frames in one chunk", () => {
+    describe("when the chunk is processed", () => {
+      it("classifies each frame, so hysteresis can be satisfied in one call", () => {
+        const { vad, starts } = makeDetector({ hysteresisFrames: 3 });
+
+        vad.process(chunkOf(pcm16(LOUD, SAMPLES_PER_FRAME * 3)));
+
+        expect(vad.isSpeaking).toBe(true);
+        expect(starts).toHaveLength(1);
+      });
+    });
+  });
+
+  describe("given a chunk carrying a whole frame plus a remainder", () => {
+    describe("when a later chunk completes the remainder", () => {
+      it("carries the leftover bytes across calls rather than dropping them", () => {
+        const { vad, starts } = makeDetector({ hysteresisFrames: 2 });
+        const loud = pcm16(LOUD, SAMPLES_PER_FRAME);
+        const half = SAMPLES_PER_FRAME; // in bytes, half a frame
+
+        vad.process(chunkOf(pcm16(LOUD, SAMPLES_PER_FRAME + SAMPLES_PER_FRAME / 2)));
+        expect(vad.isSpeaking).toBe(false);
+
+        // The remaining half frame, which only completes because the leftover
+        // survived the first call.
+        vad.process(chunkOf(loud.slice(0, half)));
+
+        expect(vad.isSpeaking).toBe(true);
+        expect(starts).toHaveLength(1);
+      });
+    });
+  });
+
+  describe("given hysteresis of three frames", () => {
+    describe("when only two loud frames arrive", () => {
+      it("stays silent, because the run has not been held long enough", () => {
+        const { vad, starts } = makeDetector({ hysteresisFrames: 3 });
+
+        vad.process(chunkOf(pcm16(LOUD, SAMPLES_PER_FRAME * 2)));
+
+        expect(vad.isSpeaking).toBe(false);
+        expect(starts).toHaveLength(0);
+      });
+    });
+
+    describe("when speech is followed by enough silence", () => {
+      it("flips back and fires the end callback exactly once", () => {
+        const { vad, starts, ends } = makeDetector({ hysteresisFrames: 3 });
+
+        vad.process(chunkOf(pcm16(LOUD, SAMPLES_PER_FRAME * 3)));
+        expect(vad.isSpeaking).toBe(true);
+
+        vad.process(chunkOf(pcm16(SILENT, SAMPLES_PER_FRAME * 3)));
+
+        expect(vad.isSpeaking).toBe(false);
+        expect(starts).toHaveLength(1);
+        expect(ends).toHaveLength(1);
+      });
+    });
+  });
+
+  describe("given a quiet negative-amplitude frame", () => {
+    describe("when it is classified", () => {
+      it("reads it as silence, which only holds if the sign survives", () => {
+        // -1 is the discriminating fixture, and -8000 is not. Read signed, -1
+        // is an RMS of 1 and silent; read unsigned it is 65535 and the
+        // loudest sample there is. A frame at -8000 is loud under BOTH
+        // readings, so it passes whether or not the sign extension exists.
+        const { vad, starts } = makeDetector({ hysteresisFrames: 1 });
+
+        vad.process(chunkOf(pcm16(-1, SAMPLES_PER_FRAME)));
+
+        expect(vad.isSpeaking).toBe(false);
+        expect(starts).toHaveLength(0);
+      });
+    });
+
+    describe("when it is loud", () => {
+      it("still reads as speech", () => {
+        const { vad } = makeDetector({ hysteresisFrames: 1 });
+
+        vad.process(chunkOf(pcm16(-8000, SAMPLES_PER_FRAME)));
+
+        expect(vad.isSpeaking).toBe(true);
+      });
+    });
+  });
+
+  describe("given consumed bytes", () => {
+    describe("when a later chunk arrives", () => {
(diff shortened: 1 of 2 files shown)


langwatch-agent (commented):
Reviewed the byte framing, remainder compaction, growth path, signed PCM16 classification, and hysteresis behavior. The refactor keeps frame boundaries and callbacks intact while removing the per-frame copy/shift work; the focused regression tests cover partial frames, remainders, repeated consumption, and negative samples. No blocking issue found. Residual operational consideration: an unusually large single audio chunk grows the retained buffer for that detector instance, as documented by the implementation.

LangWatch-Review: verdict=clean sha=7a38a3ca76f97a6b39614e1cb3bb7ef0fe409e88 p0=0 p1=0 p2=0 p3=0

drewdrewthis (commented):
Reviewed at `7a38a3ca`. No blocking findings.

The change is correct. `subarray` frames plus one `copyWithin` per `process()` call replaces a per-frame `splice` that shifted the whole remainder, and `classifyFrame`'s arithmetic is genuinely untouched because a `Uint8Array` element was already the 0-255 byte the old code read. Compaction, offset walk and the growth path all line up.

I checked the load-bearing claim rather than taking it, since it is what makes this a refactor and not a rewrite. The tests use only the public surface, and `vad.ts` on `main` carries the same `hysteresisFrames` option, the same `FRAME_MS`, `SAMPLES_PER_FRAME` and `BYTES_PER_FRAME`, and the same `0x8000` sign extension. So "these tests apply cleanly to `main`" holds.

I also checked the `-1` argument, because it is the interesting one. Signed, `-1` gives an RMS of 1, well under the 500 threshold. Unsigned, it is `65535` and clears the threshold by two orders of magnitude. `-8000` clears it either way. The fixture swap is right, and the reasoning in the comment is right.

Writing up two mutations that escaped, and stating plainly that the tests were wrong rather than the mutations, is a better artifact than a clean mutation table would have been. It is also the part a reader should trust most, because it is the part that cost something to admit.

Three notes, none blocking.

1. **The buffer never shrinks.** `reserve` doubles and keeps. One outsized chunk, say a transport that batches or a caller that hands over a whole file, pins that capacity for the detector's lifetime. The steady state you describe never grows, so this is a tail case, not a leak in normal operation. If you want it bounded, the cheap version is a reset to the two-frame default in `process()` when `bufLen` is 0 and `buf.length` is far above it.

2. **The last test's name overclaims.** "classifies exactly as many frames as the bytes allow" asserts only that the detector settled on speech and fired once, which your own comment concedes. It still kills the mutations that matter, since dropping leftovers with 101-sample chunks would never complete a frame at all. But the name promises a frame count that is not asserted, and a later reader may trust it.

3. **The test file redeclares `FRAME_MS = 30` and derives its own `BYTES_PER_FRAME`** rather than importing them, because `vad.ts` does not export them. Low risk, since a frame-size change would break these tests loudly rather than quietly, but it is a second copy of a constant that has to stay in step.

On the measurement: 18.2 ms to 3.3 ms over 60 s of audio, with the raw runs shown and a chunk size chosen specifically so it does not divide the frame. That is the right shape of benchmark for this claim.

Not claimed: I did not run the benchmark, the test suite, or `tsc`. I read the diff, `vad.ts` on `main`, and reasoned about the fixtures.

One process note, not about the code: the `APPROVED` state here is the in-repo low-risk automation, not a human sign-off, and this PR says explicitly it is advisory and not to be merged by the bot.

drewdrewthis:
## Review verdict: READY

Reviewed at `7a38a3ca76f97a6b39614e1cb3bb7ef0fe409e88`.

No blocking findings.

The change is correct: `subarray` frames plus one `copyWithin` per `process()` call replace a per-frame `splice` that shifted the whole remainder, and `classifyFrame`'s arithmetic is untouched because a `Uint8Array` element was already the 0-255 byte the old code read.

Verified rather than taken on trust:

- **"These tests apply cleanly to `main`"** - the load-bearing claim, since it is what makes this a refactor. The tests use only the public surface, and `vad.ts` on `main` carries the same `hysteresisFrames` option, the same frame constants, and the same `0x8000` sign extension. Holds.
- **The `-1` fixture argument.** Signed, `-1` is an RMS of 1, far under the 500 threshold; unsigned it is `65535` and clears it by two orders of magnitude. `-8000` clears it either way. The swap is right and so is the reasoning.

Documenting two escaped mutations, and saying plainly that the tests were wrong rather than the mutations, is worth more than a clean table would have been.

Non-blocking:

1. **The buffer never shrinks.** `reserve` doubles and keeps, so one outsized chunk pins that capacity for the detector's lifetime. A tail case, not a normal-operation leak.
2. **The last test's name overclaims** - "classifies exactly as many frames as the bytes allow" asserts only that the detector settled on speech and fired once. It still kills the mutations that matter.
3. **The test file redeclares `FRAME_MS` and derives its own `BYTES_PER_FRAME`** because `vad.ts` exports neither. A second copy of a constant that must stay in step, though drift would fail loudly.

Not claimed: I did not run the benchmark, the test suite, or `tsc`. Static read of the diff and of `vad.ts` on `main`.

The `APPROVED` state on this PR is the in-repo low-risk automation, not a human sign-off, and the PR itself says it is advisory and not to be merged by the bot.

langwatch-agent:
## Human Review Brief

> [!CAUTION]
> **This is mergeable right now, and no human has approved it.** `mergeStateStatus` is `CLEAN` and the review decision reads `APPROVED`, but the only approving review is the repo's own `low-risk-change` automation. <@drewdrewthis> reviewed at head on 20 Aug and posted no blocking findings, as a `COMMENTED` review. So nothing is standing, and the merge button is live anyway. That is the low-risk policy working as designed rather than a defect, and it is worth knowing before someone assumes the green state implies a sign-off.
>
> **The risk in the change itself is small, and it is not what the title suggests.** `subarray` returns a view into the shared buffer rather than a copy, so aliasing becomes possible where it structurally was not before. I checked: the view never escapes. `classifyFrame` reads it inside the loop and returns a boolean, `observe` receives the boolean, and the `copyWithin` compaction runs only after the loop. Nothing holds a frame across a mutation. Worth knowing that this is the property to preserve, because a future change that stashes a frame for later would break silently rather than loudly.
>
> **The buffer grows and never shrinks.** `reserve` doubles when a single chunk exceeds capacity, which the steady state never does, and there is no path back down. One outsized chunk sets a permanent high-water mark for that detector's lifetime. On an SDK-side detector that is almost certainly fine, and it is a decision rather than an oversight.
>
> **The author is an agent.** Nobody can answer a design question on this PR or address a finding, so the questions below need whoever adopts it rather than a reply.

| | |
|---|---|
| **Mode** | Targeted Review. Two files, and the interesting one is the test file. |
| **Issue** | [#525](https://github.com/langwatch/scenario/issues/525) `P3 - low` `refactor` `tech-debt`. Scope matches exactly, including the before-and-after measurement the ticket asked for. |
| **State** | Every latest-per-name check green. `MERGEABLE` / `CLEAN`. Zero unresolved threads. Automation approval only. |
| **Evidence** | The strongest part, and unusual: ten tests written against the *original* implementation, which pass unchanged against the new one and apply cleanly to `main`. |
| **Where to look** | Nothing in the diff. Read the test file, then decide whether the merge policy should let an agent-authored change through on an automation approval. |

<details><summary><b>📊 What the numbers do and do not say</b> · ~1 min</summary>
<br/>

The measurement is 18.2 ms down to 3.3 ms, median of seven runs, over 60 seconds of 24 kHz PCM16 delivered in 20 ms chunks.

Read that as a ratio and it is 5.5x. Read it as a proportion and both numbers are noise: 18 ms of work across 60 seconds of audio is 0.03% of realtime. Nobody was experiencing this as a symptom, and the PR does not claim they were.

The value is the shape, not the constant. `splice(0, BYTES_PER_FRAME)` shifts the whole remainder once **per frame**, so cost grows with how much backlog the buffer is carrying. The benchmark deliberately uses 20 ms chunks against 30 ms frames so a remainder always exists, which is honest, and it is still the steady state, where the backlog is at most one frame and the quadratic barely shows.

The case this actually protects is the one not measured: a consumer that stalls, a burst that arrives while frames are not being drained, a long session on a slow device. That is where per-frame shifting of a growing remainder stops being 18 ms. Removing a quadratic is worth doing on the strength of the shape alone, and the benchmark is the weakest argument in the PR rather than the strongest.

</details>

<details><summary><b>🎯 The best thing here is a test story, not a diff</b> · ~2 min</summary>
<br/>

The framing had no direct coverage. The existing `vad-fallback.test.ts` drives the one-shot fallback warning, not what `process()` does with bytes. So the buffer that decides when speech starts and stops was, until this PR, untested.

Ten tests were written **against the original `number[]` implementation** and pass unchanged against the new one. That is what makes this a refactor rather than a rewrite, and the claim is checkable rather than asserted: the test file applies cleanly to `main`.

Then five mutations, five caught. And the part worth reading twice:

**Two escaped on the first pass, and the tests were wrong, not the mutations.**

- The sign-extension test used amplitude `-8000`. Read signed that is loud. Read unsigned as `57536` it is *also* loud. The test passed with the sign extension and without it, so it was proving nothing. The discriminating fixture is `-1`: signed it is an RMS of 1 and silent, unsigned it is `65535` and the loudest sample available. The test now uses `-1` and asserts silence.
- Every framing test used uniform audio, so re-reading consumed bytes produced the same verdict and the never-compact mutation survived. Alternating loud and silent frames makes the reprocessing observable as extra speech transitions, which is exactly what an uncompacted buffer would leak into production.

Both are the same failure: a fixture that cannot distinguish the two behaviours it is meant to separate. That is the hardest class of bad test to notice, because it is green and it looks specific.

| Term | What it means here |
|---|---|
| **VAD** | Voice activity detection. Deciding whether a slice of audio is speech or silence. This is the SDK-side fallback for when the transport does not provide it. |
| **PCM16** | Raw audio as signed 16-bit little-endian samples. Two bytes per sample, and the sign matters: read unsigned, quiet negative samples look like maximum amplitude. |
| **Frame** | A 30 ms window, the unit that gets classified. At 24 kHz that is 1,440 bytes, and transports emit 20 ms chunks, so a frame rarely arrives whole. |
| **Hysteresis** | Requiring N consecutive frames to agree before flipping state, so one loud frame does not start speech. |
| **`subarray` vs `slice`** | `subarray` is a view over the same memory and copies nothing. `slice` copies. Swapping one for the other is the entire performance change. |

</details>

<details><summary><b>⚖️ What approving this makes permanent</b> · three decisions · ~1 min</summary>
<br/>

1. **Frames are views, not copies.** The invariant that keeps that safe is that a frame is consumed before the next compaction, which holds today and is not enforced by anything. A comment records it; a type cannot.
2. **Capacity doubles and never shrinks.** Deliberate, documented, and only reachable via an outsized chunk.
3. **Tests written against the old implementation first.** This is the methodology decision and the one worth carrying beyond this PR. It converts "the refactor is behaviour-preserving" from a claim into something a reviewer can check by applying the test file to `main`.

</details>

<details><summary><b>❓ For whoever adopts this, since the author cannot answer</b> · ~1 min</summary>
<br/>

1. **Should the merge policy let an agent-authored change through on an automation approval alone?** Not a question about this PR, which is fine. A question about the class, and this is a good example to decide it on because the change is small enough to reason about completely.
2. **Is the never-escapes property worth a test?** A test that fails if a frame view is retained past a compaction would turn today's correct-by-reading into correct-by-construction.
3. **`tsc --noEmit` reports one error on this branch and one on `main`**, `TS2688` for `yauzl`, naming no file in this diff. Pre-existing and measured against a clean worktree, which is the right way to report it, and it is still an error sitting on `main`.

</details>

> [!NOTE]
> The invariant: **only how many bytes have arrived decides when a frame is classified, never how they were chopped into chunks.** That sentence is in the test file's header, and every test below it is an instance of it. It is also the property the old `splice` implementation happened to satisfy expensively and the new one satisfies cheaply.

PR #929, opened 2026-08-19
feat(javascript): migrate to AI SDK 7

0xdeafcafe:
Moves `@langwatch/scenario` to AI SDK 7.

**This is not a breaking change.** `package.json`'s module shape is byte-identical to `main` — the dual CJS+ESM build stays exactly as it is, and consumers see no difference. The only changed lines are dependency versions and the Node floor.

## Why CommonJS survives

`ai@7` publishes no `require` condition, which looks like it forces an ESM-only package. It doesn't: Node's `require(esm)` loads it, unflagged since **22.12**, and `ai@7` already requires Node >=22 — so the only consumers this could strand are ones pinned to 22.0–22.11. Verified end to end: the CJS bundle builds and `require('./dist/index.js')` loads it. `@langwatch/web`'s production `server.cjs` already does the same thing eleven times over.

The one real risk is **top-level await**. If a future `ai` release introduces it anywhere in its graph, `require()` starts throwing `ERR_REQUIRE_ASYNC_MODULE`, visible only to CommonJS consumers and only at runtime. `smoke:dist` now covers both formats so that arrives as a failed build instead of a failed consumer process.

## What actually broke, and why tsc didn't catch it

v7 kept deprecated aliases for nearly everything it renamed, so broken code typechecks clean. These were found by running 7.0.64 against a 6.0.217 baseline, not by reading the migration guide.

**System messages are rejected inside `messages`** — v7 throws `InvalidPromptError` where v6 only warned. The judge, the user simulator and the composable voice adapter each led their message array with a system turn; all three now pass it as top-level `instructions`.

The red-team attacker is the deliberate exception and keeps `allowSystemInMessages: true`: its `H_attacker` history interleaves system-role `[SCORE]`, `[BACKTRACKED]` and `[INJECTED]` markers with the conversation, so hoisting would change what the attacker model sees and how it behaves.

Hoisting also shifted one index that was not obvious. `forceVerdict` recaps discovery by collapsing the decision-call history and slicing off *"the two messages the verdict phase rebuilds by itself (the system prompt and the criteria block)"*. With the system prompt no longer in that array the slice drops one, not two — otherwise the first recap entry is eaten and the model loses what the discovery tools returned. One test caught this.

**Telemetry is off until it is registered** — v7 moves OpenTelemetry into `@ai-sdk/otel` and emits *nothing* until `registerTelemetry` runs. Measured: 2 spans on v6, 0 on v7, 3 once registered. Without this the judge span collector and the LangWatch exporter would both have gone silent. `setupScenarioTracing` now registers the integration on both paths — fresh setup and attach-to-existing-provider.

**Renames:** `stepCountIs` → `isStepCount`, `experimental_telemetry` → `telemetry`.

## Examples

The examples are how people learn this library, so they teach the shape that works on v7. Fourteen hoisted their system turn to `instructions`. The OpenAI voice helper keeps its system message — it builds `ChatCompletionMessageParam` for a direct OpenAI API call, not an AI SDK one.

Two v7 type changes needed fixing, both caught by tsc: `ToolExecutionOptions` gained a required `context` field, and `GenerateTextResult` takes three type arguments.

## Verification

- 1273 tests pass
- both dist formats build, and `smoke:dist` loads each
- the two pre-existing `@ag-ui/core` type errors in `event-reporter.ts` are unchanged from `main` and unrelated — confirmed against a pristine `origin/main` worktree

## Release note

This is a **minor**. Please keep the squash message as `feat(javascript): migrate to AI SDK 7` — no `!`, no `BREAKING CHANGE` footer — or Release Please will cut an unwarranted 2.0.0.

### DIFF
--- a/javascript/examples/vitest/package.json
+++ b/javascript/examples/vitest/package.json
@@ -21,11 +21,11 @@
     "vitest-mock-extended": "5.1.1"
   },
   "dependencies": {
-    "@ai-sdk/anthropic": "3.0.41",
-    "@ai-sdk/xai": "3.0.50",
+    "@ai-sdk/anthropic": "4.0.38",
+    "@ai-sdk/xai": "4.0.40",
     "@langwatch/scenario": "workspace:*",
     "@openai/agents": "^0.16.0",
-    "ai": "^6.0.0",
+    "ai": "^7.0.64",
     "express": "^5.1.0",
     "openai": "^6.9.1",
     "vite": "^8.2.1",
--- a/javascript/examples/vitest/tests/00-demo-light.test.ts
+++ b/javascript/examples/vitest/tests/00-demo-light.test.ts
@@ -25,15 +25,10 @@ describe("Demo: Lightweight Scenarios", () => {
     call: async (input) => {
       const response = await generateText({
         model: openai("gpt-5-mini"),
+        instructions: "You can analyze images and respond briefly and helpfully.",
         messages: [
-          {
-            role: "system",
-            content:
-              "You can analyze images and respond briefly and helpfully.",
-          },
           ...input.messages,
         ],
-        experimental_telemetry: { isEnabled: true },
       });
       return response.text;
     },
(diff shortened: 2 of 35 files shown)


sergioestebance (changes requested):
This PR has merge conflicts with the base branch. Please rebase and resolve them so we can proceed with the review.

Issue #681, opened 2026-06-16
chore(audio-examples): align agent prompts to judge criteria across all 3 siblings


drewdrewthis:
## Summary

All 3 audio example test files instruct the agent via their message prompt to exhibit behaviors the LLM judge no longer checks for. This creates a coherence gap: the prompt is stricter than the judge, and a reader cannot tell if that is intentional.

## Problem

Each affected test includes this (or equivalent) text in the user message:
```
"After you've guessed, you must repeat the question and say what format the input was in (audio or text)"
```

The old LLM-judge criteria checked for these behaviors verbatim. After #612 and #655, the judge uses generic behavioral criteria that deliberately do NOT check for exact phrasing or that the question was repeated. But the prompt still commands the agent to do these things.

A reader sees:
- **Prompt**: _"you MUST repeat the question and say what format"_  
- **Judge criteria**: _"exact phrasing does not matter"_

These are in direct contradiction. The next engineer can't tell which is the spec.

## Files affected

- `javascript/examples/vitest/tests/multimodal-audio-to-audio.test.ts` (~line 57)
- `javascript/examples/vitest/tests/multimodal-audio-to-text.test.ts` (~line 54)
- `python/examples/test_audio_to_text.py` (~line 201)

## Fix

In one coordinated PR across all 3 siblings, either:
1. **Drop the vestigial prompt clause** ("repeat the question / say what format") — the judge no longer checks it, so the instruction is noise that misleads about what correct behavior looks like
2. **Or add a brief code comment** at the criteria explaining that the prompt is intentionally stricter than the judge, so paraphrasing agents still pass

Option 1 is cleaner. Option 2 is safer if the prompt instruction is serving a different purpose (e.g., producing richer agent output for manual review).

## Context

Surfaced during review of #655 by [principles], [uncle-bob], [metz-beck]. The mismatch is pre-existing across all 3 siblings — fixing in #655 alone would break parity.

Issue #680, opened 2026-06-16
refactor(audio-examples): tighten criteria 1+3 across all 3 audio example siblings


drewdrewthis:
## Summary

Three audio example test files share identical LLM-judge criteria (introduced/aligned via #655 and #612), but two of the criteria have latent coverage gaps that a future test run may surface as false-passes.

## Problem

**Criterion 1 — vacuous-green risk:**
```
"The agent's response demonstrates it processed the audio content (e.g. it addresses what was in the audio, attempts to answer the audio question, or acknowledges what it heard)"
```
The final branch — _"acknowledges what it heard"_ — allows a broken audio pipeline to pass. An agent that receives empty/silent audio and replies _"I received your audio"_ satisfies criterion 1 without any content-specific processing.

**Criterion 3 — polite deflection gap:**
```
"The agent's response indicates it received input in a non-text format, or that the question came via audio rather than text (exact phrasing does not matter)"
```
A polite deflection ("I'm sorry, I can't process audio files") mentions the non-text format and satisfies criterion 3. Criterion 2 ("not an error message, refusal, or unrelated reply") should catch this, but "polite deflection" is ambiguous — the two criteria together don't close the gap.

## Files affected

- `javascript/examples/vitest/tests/multimodal-audio-to-audio.test.ts` (criteria lines 82–86)
- `javascript/examples/vitest/tests/multimodal-audio-to-text.test.ts` (criteria lines 109–113)
- `python/examples/test_audio_to_text.py` (criteria lines 217–221)

## Fix

Must be a single coordinated PR touching all 3 siblings to preserve parity:
1. Tighten criterion 1 to require content-specific processing (remove "acknowledges what it heard" or add "and responds to its specific content")
2. Tighten criterion 2 to explicitly include "polite deflection / inability claim" in its exclusion list, OR scope criterion 3 to only apply to passing (non-deflection) responses

## Context

Surfaced during review of #655 by [test-reviewer] and [uncle-bob]. Fixing in a parity-only PR would break the alignment these files now share.

drewdrewthis:
Delivered in #861 (draft, targeting `main`) — one coordinated PR touching all three siblings, one commit per fix item.

**One finding that changes how this issue should be read**, surfaced while proving the fix:

[...]

Issue #655, opened 2026-06-10
audio example tests retain brittle judge criteria (audio-to-audio + py audio-to-text) — parity with #612's generic rewrite


drewdrewthis:
## For Humans

**Two audio-example test files retain brittle LLM-judge criteria that assert exact model behaviors a correct agent might not exhibit — parity gap left when #612 fixed only the TS audio-to-text sibling.** The TS audio-to-audio and Python audio-to-text tests both still use `"The agent correctly guesses it's a male voice"` / `"The agent repeats the question"` / `"The agent says what format the input was in"` — criteria that fail a correct agent that paraphrases or hedges.

**Ask:** Replace the 3 brittle criteria in `multimodal-audio-to-audio.test.ts` and `test_audio_to_text.py` with the 3 generic behavioral criteria already proven in the fixed TS sibling.

_Status: investigating → needs-plan. Full spec below._

<details><summary>For Agents — full spec</summary>

### Symptoms / context

Two files retain this identical 3-criterion set (confirmed by grep):

- `javascript/examples/vitest/tests/multimodal-audio-to-audio.test.ts` lines 83–85
- `python/examples/test_audio_to_text.py` lines 218–220

```
"The agent correctly guesses it's a male voice"
"The agent repeats the question"
"The agent says what format the input was in (audio or text)"
```

The TS audio-to-text sibling (`multimodal-audio-to-text.test.ts`) was fixed in #612 to use 3 generic behavioral criteria instead.

### Root cause / key findings

**The criteria over-specify model output rather than testing observable pipeline behavior.**

- `"correctly guesses it's a male voice"` — voice-gender classification is non-deterministic and not the capability under test (multimodal audio handling, not gender detection). A correct agent that hedges ("this sounds like it could be male") fails.
- `"repeats the question"` — prescribes exact conversational behavior; a correct agent that summarizes or paraphrases fails.
- `"says what format the input was in"` — prescribes specific phrasing; a correct agent that says "based on the audio clip" rather than "audio" fails.

**The fixed sibling already has the right pattern.** `multimodal-audio-to-text.test.ts` lines 110–114:

```typescript
"The agent's response demonstrates it processed the audio content (e.g. it addresses what was in the audio, attempts to answer the audio question, or acknowledges what it heard)"
"The agent provides a coherent, on-topic response — not an error message, refusal, or unrelated reply"
"The agent's response indicates it received input in a non-text format, or that the question came via audio rather than text (exact phrasing does not matter)"
```

[...]

PR #612, opened 2026-06-04
chore(examples/voice/#486): retire legacy gpt-4o-audio-preview surface, migrate supported audio examples to gpt-audio-mini

drewdrewthis:
## What

Cohesive retirement of the legacy `gpt-4o-audio-preview` voice/audio **example** surface, consolidating three threads into one PR:

- **Folds in #607** (model swap `gpt-4o-audio-preview` → `gpt-audio-mini`) — its commits are cherry-picked here, so #607 is superseded and will be closed.
- **Supersedes #486** (the "unskip the voice/audio tests" campaign) — see "Why we don't unskip" below; #486 will be closed with reasoning.
- **Original #612 scope** (delete the deprecated voice-to-voice test) — retained.

## Why we don't unskip (the key correction)

#486 framed the goal as "remove the `skipif(CI)` markers and restore CI coverage." That premise was never achievable: these audio/voice example tests are **live end-to-end tests** — they call real OpenAI (`gpt-audio-mini`) and the real LangWatch backend, incur cost, and produce non-deterministic audio. They are correctly CI-skipped *regardless of model* (same class as the other live `voice/*` example tests). The dead `gpt-4o-audio-preview` model was only the *historical* reason for the skip; swapping it doesn't make these CI-runnable.

So the right end-state is **migrated + intentionally CI-skipped**, not "unskipped." We migrate the model so the examples work when run live/locally, keep the skip markers, and fix the stale skip *comments* to state the real reason.

## What changed

**Genuinely-dead → retired:**
- Tombstone `docs/docs/pages/examples/multimodal/voice-to-voice.mdx` and `testing-voice-agents.mdx` → a short pointer to [`/voice/getting-started`](/voice/getting-started). The pages' URLs still return 200 (the vocs fork has no redirect layer, so a tombstone avoids 404s on previously-public URLs).
- Delete the now-unused `LegacyVoiceDeprecation.mdx` snippet (zero importers after the edits).
- `test_voice_to_voice_conversation.py` deleted (it was explicitly `DEPRECATED` — the legacy single-call pattern).

**Supported → kept + migrated to `gpt-audio-mini`:**
- `audio-to-text.mdx` / `audio-to-audio.mdx` **kept and updated** (prereq prose now names `gpt-audio-mini`; `LegacyVoiceDeprecation` banner removed) — these document the *current* supported single-call pattern, not a legacy one.
[...]

PR #936, opened 2026-08-21
chore(audio-examples): bring the fourth sibling into the shared judge criteria (#682)

### LINKED ISSUE #682: chore(audio-examples): align python/examples/test_audio_to_audio.py criteria with generic set
## Summary

Three of four audio example test files now use the same 3 generic behavioral LLM-judge criteria (aligned via #612 and #655). The fourth — `python/examples/test_audio_to_audio.py` — still uses the old style.

## Current state of test_audio_to_audio.py

```python
criteria=[
    "The agent identifies or guesses the voice is male",
    "The agent acknowledges the input was audio",
    ...
]
```

These are over-specific (voice-gender classification is non-deterministic and not the capability under test).

## The generic criteria now used by the other 3 siblings

```python
criteria=[
    "The agent's response demonstrates it processed the audio content (e.g. it addresses what was in the audio, attempts to answer the audio question, or acknowledges what it heard)",
    "The agent provides a coherent, on-topic response — not an error message, refusal, or unrelated reply",
    "The agent's response indicates it received input in a non-text format, or that the question came via audio rather than text (exact phrasing does not matter)",
]
```

## Fix

Either:
1. Update `test_audio_to_audio.py` to use the same 3 generic criteria as the siblings
2. Or document explicitly why this file intentionally differs (in a code comment at the criteria block)

Option 1 is preferred unless this test has a genuinely different behavioral contract.

## Context

Surfaced during review of #655 by [fowler] and [hygiene]. Out of scope for #655 (which targeted two specific files). Align in a dedicated PR.

langwatch-agent:
## Ticket

Closes #682. `python/examples/test_audio_to_audio.py` still judged against the pre-alignment criteria, including `"The agent identifies or guesses the voice is male"` - voice-gender classification, which is non-deterministic and is not the capability the example tests.

## Premise, re-checked on current main

Still true, and the interesting part is why it survived. Since the ticket was filed, #680 built real machinery for this: `javascript/examples/vitest/tests/helpers/audio-judge-criteria.ts` holds the criteria, and `audio-judge-criteria-parity.test.ts` keeps the Python copy byte-identical and asserts both JS siblings actually pass the constant to their judge rather than an inline array.

That machinery covers three siblings. `test_audio_to_audio.py` is not one of them, so it drifted for a year with nothing to notice. The ticket's option 1 ("update it to use the same criteria") fixes today's file; putting it inside the parity set is what stops the next one.

## Change

- The Python criteria move out of `test_audio_to_text.py` into `python/examples/helpers/audio_judge_criteria.py` and are exported from the `helpers` package, mirroring the JS helper. One copy per language instead of one per test file.
- Both Python examples import it. `test_audio_to_audio.py` drops its own two criteria.
- The vitest parity guard reads the new module and keeps the one check that genuinely spans both languages: the two constants stay byte-identical.
- The shared criteria are immutable, and the guard checks the object the examples actually import, `helpers/__init__.py` included: the Python copy is a `tuple`, the JS copy is `as const`. Every other guard here watches the *name*, and mutating the object binds no name, so an `AUDIO_JUDGE_CRITERIA.append(...)` in an example would have handed the judge a drifted criterion with all of them green. Making the object immutable removes the move rather than enumerating the ways to spell it, and unlike a name-based check it holds through an alias.
- Whether the Python siblings actually *use* theirs is checked in Python, by `python/tests/test_audio_example_judge_criteria.py`, which walks each example's AST. It inspects every `JudgeAgent(...)` call and its `criteria` argument, and requires the criteria name to be bound exactly once in the file and by an import, with `list` never rebound. Text matching could not see a second judge with an inline list, a parameter or `for` target or walrus that shadows the name, or an import that stopped coming from `helpers`. It parses files, so no keys and no network, and it runs in `python-ci`'s unit step.

The criteria strings themselves are unchanged, which the parity test enforces.

The prompt in `test_audio_to_audio.py` stays as it is and reads coherently against the new criteria: it asks the agent to answer the question in the audio (criterion 1) and to mention it received audio (criterion 3). Prompt/criteria coherence in the other three siblings is #681's scope, not this one.

## Evidence

The strongest check is that the extended guard fails on exactly the state the ticket reports. With `test_audio_to_audio.py` restored to its `origin/main` content:

```
FAIL  audio-judge-criteria-parity.test.ts > keeps test_audio_to_audio.py on the shared constant, not an inline copy
AssertionError: test_audio_to_audio.py must pass AUDIO_JUDGE_CRITERIA to the judge, not an inline list
```

With the fix in place:

```
pnpm -F vitest-examples exec vitest run tests/audio-judge-criteria-parity.test.ts
  Test Files  1 passed (1)
       Tests  4 passed (4)

uv run pytest tests/ -m "not integration"
  1328 passed, 10 skipped         # 12 of them the new guard

uv run pytest examples/test_audio_to_audio.py examples/test_audio_to_text.py --collect-only -q
  2 tests collected

uv run pyright examples/test_audio_to_audio.py examples/test_audio_to_text.py examples/helpers/
  0 errors, 0 warnings, 0 informations

eslint tests/audio-judge-criteria-parity.test.ts     clean
```

Resolved at runtime, not just read off the page: importing `helpers` gives 3 criteria, and parsing the JS constant out of `audio-judge-criteria.ts` and comparing gives `byte-identical to JS: True`.

The guard is deterministic (file read plus parse, no API keys, no network), so it runs in CI unlike the live examples it protects.

### Mutation testing

| Mutation | Result |
| --- | --- |
| python constant drifts one word from JS | CAUGHT |
| a second judge in the file gets an inline list | CAUGHT |
| either sibling pastes an inline criteria list back | CAUGHT |
| the judge is given no criteria at all | CAUGHT |
| the constant is re-declared inside a sibling | CAUGHT |
| the import stops coming from `helpers` | CAUGHT |
| the constant is imported under an alias | CAUGHT |
| a parameter shadows the constant | CAUGHT |
| a `for` target shadows the constant | CAUGHT |
| a walrus shadows the constant | CAUGHT |
| a `with ... as` shadows the constant | CAUGHT |
| an `except ... as` shadows the constant | CAUGHT |
| a lambda parameter shadows the constant | CAUGHT |
| a second import rebinds the constant | CAUGHT |
| the `list` builtin is shadowed | CAUGHT |
| the shared constant reverts to a mutable `list` | CAUGHT |
| an example appends to the shared criteria | CAUGHT (pyright) |
| an example aliases the constant, then appends | CAUGHT (pyright) |
| an example assigns to one criterion by index | CAUGHT (pyright) |
| the `helpers` facade exports a stale list beside the tuple | CAUGHT |
| the facade exports a genuine copy, `(*_shared,)` | CAUGHT |
| the facade exports a copy with one criterion drifted | CAUGHT |

The last twelve came out of review. The first version of this guard matched source text, which proves one occurrence rather than the property; a later version checked only assignment-shaped rebindings; and the version after that still watched only the name, so mutating the object behind it escaped (verified: `10 passed` with the exact wording #680 removed appended back on). CPython's own `symtable` answers the binding cases without an enumeration to get wrong, and an immutable object answers the mutation cases without one either.

One near-miss is worth recording, since it looked like an escape and was not. The first copy mutation tried was `AUDIO_JUDGE_CRITERIA = tuple(_shared)`, which passed the identity check. That is CPython returning its argument unchanged: `tuple(t) is t` holds for an exact tuple, as does `t[:] is t`. There was no copy to catch. `(*_shared,)` does build a new tuple, and the check fails on it.

The four `(pyright)` rows are labelled rather than folded in: `pytest tests/` alone does not catch them, `uv run pyright .` does, and `python-ci` runs both in the same job with `examples/` inside pyright's scope. At runtime the interpreter raises `AttributeError` regardless.

### Pre-existing, not from this branch

`javascript-ci` is red on `main` and red here, with the **same file failing in both**, and it is not one this PR touches:

```
main   run 32476011706 (76ee430e)      this branch  run 32480323868 (ab2ae7d9)
  tests/scenario-expert-realtime.test.ts               (same, and the only one)
```

It is a live realtime example. The rest of that job is `34 passed | 21 skipped`, and `audio-judge-criteria-parity.test.ts` is among the passing four-test files. `python-ci`, which runs both `pytest tests/` and `pyright .`, is **green** on this branch, so every guard this PR adds is verified in CI rather than only locally.

Earlier pushes on this branch saw five live files failing, on main and here alike; that set shrank to one on both sides. These examples call real models, so which of them fail varies run to run. What stays constant is that the set matches main's.

Locally `tsc --noEmit` in `examples/vitest` reports `TS2688: Cannot find type definition file for 'yauzl'`. That is this box's install, not the branch: the same error appears with `origin/main`'s copy of the only TypeScript file this PR touches, and CI's own build step passes.

## Advisory note

---
<sub>🤖 **Filed by `tech-debt-fixer` via the create-issue procedure, on behalf of the fleet.** Authored under the owner's GitHub token for API access — the content is the fleet's, not his. Owner-authored and fleet-authored issues are otherwise indistinguishable in this repo; this footer is the only signal.</sub>

### DIFF
--- a/javascript/examples/vitest/tests/audio-judge-criteria-parity.test.ts
+++ b/javascript/examples/vitest/tests/audio-judge-criteria-parity.test.ts
@@ -29,12 +29,11 @@ const JS_SIBLINGS = [
   "multimodal-audio-to-text.test.ts",
 ];
 
-/**
- * Python siblings, relative to `python/examples/`. They import the shared
- * constant rather than declaring one, so the drift they can suffer is the same
- * one the JS pair can: importing it and then passing something else.
- */
-const PYTHON_SIBLINGS = ["test_audio_to_text.py", "test_audio_to_audio.py"];
+// Whether the Python siblings actually USE their constant is checked where it
+// can be checked properly: python/tests/test_audio_example_judge_criteria.py
+// walks each example's AST and inspects every JudgeAgent call. Text matching
+// from here could not see a second judge with an inline list, a local rebinding
+// that shadows the import, or an import that stopped coming from `helpers`.
 
 /**
  * Extract the `AUDIO_JUDGE_CRITERIA = [...]` literal from the Python module.
@@ -94,43 +93,6 @@ describe("audio example judge criteria — cross-language parity", () => {
       `${sibling} inlines a judge criterion again — import AUDIO_JUDGE_CRITERIA instead`,
     ).not.toContain("The agent's response demonstrates");
   });
-
-  // The Python pair can drift the same way: import the constant and then hand
-  // the judge something else. #682 is the case this catches: audio-to-audio
-  // sat outside the shared set for a year with its own two criteria, one of
-  // them ("identifies or guesses the voice is male") non-deterministic and not
-  // the capability under test.
-  it.each(PYTHON_SIBLINGS)(
-    "keeps %s on the shared constant, not an inline copy",
-    (sibling) => {
-      const source = readFileSync(
-        fileURLToPath(
-          new URL(`../../../../python/examples/${sibling}`, import.meta.url),
-        ),
-        "utf8",
-      );
-      expect(
-        source,
-        `${sibling} must pass AUDIO_JUDGE_CRITERIA to the judge, not an inline list`,
-      ).toMatch(
-        // Same shape as the JS check: assert on what reaches the judge. The
-        // \b stops a drifted `criteria=list(AUDIO_JUDGE_CRITERIA_STALE)` from
-        // satisfying the match.
-        /\bcriteria\s*=\s*(?:list\(\s*AUDIO_JUDGE_CRITERIA\s*\)|AUDIO_JUDGE_CRITERIA\b)/,
-      );
-      expect(
-        source,
-        `${sibling} declares its own criteria list again. Import AUDIO_JUDGE_CRITERIA instead`,
-      ).not.toContain("AUDIO_JUDGE_CRITERIA = [");
-      // A file can hand one judge the constant and another an inline list, so
-      // the match above is not enough on its own. Same second assertion the JS
-      // siblings carry.
-      expect(
-        source,
-        `${sibling} inlines a judge criterion again. Import AUDIO_JUDGE_CRITERIA instead`,
-      ).not.toContain("The agent's response demonstrates");
-    },
-  );
 
   it("still carries the #680 tightenings the siblings were fixed for", () => {
     const [contentSpecific, coherent] = AUDIO_JUDGE_CRITERIA;
--- a/javascript/examples/vitest/tests/audio-judge-criteria-parity.test.ts
+++ b/javascript/examples/vitest/tests/audio-judge-criteria-parity.test.ts
@@ -36,18 +36,21 @@ const JS_SIBLINGS = [
 // that shadows the import, or an import that stopped coming from `helpers`.
 
 /**
- * Extract the `AUDIO_JUDGE_CRITERIA = [...]` literal from the Python module.
+ * Extract the `AUDIO_JUDGE_CRITERIA = (...)` literal from the Python module.
  *
- * The Python list is written as double-quoted string literals precisely so it
- * parses as JSON once the trailing comma is dropped. See the warning in that
- * module's docstring.
+ * The Python tuple is written as double-quoted string literals precisely so it
+ * parses as JSON once the parentheses are swapped for brackets and the trailing
+ * comma is dropped. See the warning in that module's docstring. The annotation
+ * between the name and the `=` is deliberately not pinned here: what this guard
+ * is about is the criteria, and a stricter pattern would fail as a rename of the
+ * type rather than as the drift it exists to catch.
  */
 function readPythonCriteria(): string[] {
   const source = readFileSync(PYTHON_CRITERIA_MODULE, "utf8");
-  const match = /^AUDIO_JUDGE_CRITERIA = \[\n(.*?)^\]$/ms.exec(source);
+  const match = /^AUDIO_JUDGE_CRITERIA[^=\n]*= \(\n(.*?)^\)$/ms.exec(source);
   expect(
     match,
-    `could not find an 'AUDIO_JUDGE_CRITERIA = [' list literal in ${PYTHON_CRITERIA_MODULE}. ` +
+    `could not find an 'AUDIO_JUDGE_CRITERIA = (' tuple literal in ${PYTHON_CRITERIA_MODULE}. ` +
       "if it was renamed or reformatted, update this parity guard rather than deleting it",
   ).not.toBeNull();
 
--- a/javascript/examples/vitest/tests/audio-judge-criteria-parity.test.ts
+++ b/javascript/examples/vitest/tests/audio-judge-criteria-parity.test.ts
@@ -1,12 +1,14 @@
-// Cross-language parity guard for the audio example judge criteria (#680).
+// Cross-language parity guard for the audio example judge criteria (#680, #682).
 //
-// The three audio example siblings (JS audio-to-audio, JS audio-to-text, Python
-// audio-to-text) must judge against byte-identical criteria — #655/#612 aligned
-// them by copy-paste, and #680 found two latent gaps that had to be fixed in all
-// three at once. The JS pair now shares one constant, so only the Python copy
-// can drift; this test is the mechanism that stops it. It is fully deterministic
-// (file read + parse, no API keys, no network), so it runs in CI unlike its
-// live-only siblings.
+// The four audio example siblings (JS audio-to-audio, JS audio-to-text, Python
+// audio-to-text, Python audio-to-audio) must judge against byte-identical
+// criteria. #655/#612 aligned three of them by copy-paste, #680 found two
+// latent gaps that had to be fixed in all three at once, and #682 found the
+// fourth still carrying the pre-alignment criteria a year later. There is now
+// one constant per language and this test holds the two together, so no copy
+// can drift and no sibling can fall out of the set unnoticed. It is fully
+// deterministic (file read + parse, no API keys, no network), so it runs in CI
+// unlike the live-only examples it guards.
 
 import { readFileSync } from "node:fs";
 import { fileURLToPath } from "node:url";
@@ -15,8 +17,11 @@ import { describe, it, expect } from "vitest";
 
 import { AUDIO_JUDGE_CRITERIA } from "./helpers/audio-judge-criteria";
 
-const PYTHON_SIBLING = fileURLToPath(
-  new URL("../../../../python/examples/test_audio_to_text.py", import.meta.url),
+const PYTHON_CRITERIA_MODULE = fileURLToPath(
+  new URL(
+    "../../../../python/examples/helpers/audio_judge_criteria.py",
+    import.meta.url,
+  ),
 );
 
 const JS_SIBLINGS = [
@@ -25,18 +30,25 @@ const JS_SIBLINGS = [
 ];
 
 /**
- * Extract the `AUDIO_JUDGE_CRITERIA = [...]` literal from the Python sibling.
+ * Python siblings, relative to `python/examples/`. They import the shared
+ * constant rather than declaring one, so the drift they can suffer is the same
+ * one the JS pair can: importing it and then passing something else.
+ */
+const PYTHON_SIBLINGS = ["test_audio_to_text.py", "test_audio_to_audio.py"];
+
+/**
+ * Extract the `AUDIO_JUDGE_CRITERIA = [...]` literal from the Python module.
  *
  * The Python list is written as double-quoted string literals precisely so it
- * parses as JSON once the trailing comma is dropped — see the ⚠ note next to the
- * declaration in that file.
+ * parses as JSON once the trailing comma is dropped. See the warning in that
+ * module's docstring.
  */
 function readPythonCriteria(): string[] {
-  const source = readFileSync(PYTHON_SIBLING, "utf8");
+  const source = readFileSync(PYTHON_CRITERIA_MODULE, "utf8");
   const match = /^AUDIO_JUDGE_CRITERIA = \[\n(.*?)^\]$/ms.exec(source);
   expect(
     match,
-    `could not find an 'AUDIO_JUDGE_CRITERIA = [' list literal in ${PYTHON_SIBLING} — ` +
+    `could not find an 'AUDIO_JUDGE_CRITERIA = [' list literal in ${PYTHON_CRITERIA_MODULE}. ` +
       "if it was renamed or reformatted, update this parity guard rather than deleting it",
   ).not.toBeNull();
 
@@ -45,7 +57,7 @@ function readPythonCriteria(): string[] {
     return JSON.parse(`[${body}]`) as string[];
   } catch (error) {
     throw new Error(
-      `AUDIO_JUDGE_CRITERIA in ${PYTHON_SIBLING} is not JSON-parseable — the criteria ` +
+      `AUDIO_JUDGE_CRITERIA in ${PYTHON_CRITERIA_MODULE} is not JSON-parseable. The criteria ` +
         "must be plain double-quoted Python string literals (no single quotes, no " +
         `concatenation, no f-strings) so this parity guard can read them. Cause: ${String(error)}`,
     );
@@ -53,7 +65,7 @@ function readPythonCriteria(): string[] {
 }
 
 describe("audio example judge criteria — cross-language parity", () => {
-  it("keeps the Python sibling byte-identical to the shared JS constants", () => {
+  it("keeps the Python constant byte-identical to the shared JS one", () => {
     expect(readPythonCriteria()).toEqual([...AUDIO_JUDGE_CRITERIA]);
   });
 
@@ -82,6 +94,43 @@ describe("audio example judge criteria — cross-language parity", () => {
       `${sibling} inlines a judge criterion again — import AUDIO_JUDGE_CRITERIA instead`,
     ).not.toContain("The agent's response demonstrates");
   });
+
+  // The Python pair can drift the same way: import the constant and then hand
+  // the judge something else. #682 is the case this catches: audio-to-audio
+  // sat outside the shared set for a year with its own two criteria, one of
+  // them ("identifies or guesses the voice is male") non-deterministic and not
+  // the capability under test.
+  it.each(PYTHON_SIBLINGS)(
+    "keeps %s on the shared constant, not an inline copy",
+    (sibling) => {
+      const source = readFileSync(
+        fileURLToPath(
+          new URL(`../../../../python/examples/${sibling}`, import.meta.url),
+        ),
+        "utf8",
+      );
+      expect(
+        source,
+        `${sibling} must pass AUDIO_JUDGE_CRITERIA to the judge, not an inline list`,
+      ).toMatch(
+        // Same shape as the JS check: assert on what reaches the judge. The
+        // \b stops a drifted `criteria=list(AUDIO_JUDGE_CRITERIA_STALE)` from
+        // satisfying the match.
+        /\bcriteria\s*=\s*(?:list\(\s*AUDIO_JUDGE_CRITERIA\s*\)|AUDIO_JUDGE_CRITERIA\b)/,
+      );
+      expect(
+        source,
+        `${sibling} declares its own criteria list again. Import AUDIO_JUDGE_CRITERIA instead`,
+      ).not.toContain("AUDIO_JUDGE_CRITERIA = [");
+      // A file can hand one judge the constant and another an inline list, so
+      // the match above is not enough on its own. Same second assertion the JS
+      // siblings carry.
+      expect(
+        source,
+        `${sibling} inlines a judge criterion again. Import AUDIO_JUDGE_CRITERIA instead`,
+      ).not.toContain("The agent's response demonstrates");
+    },
+  );
 
   it("still carries the #680 tightenings the siblings were fixed for", () => {
     const [contentSpecific, coherent] = AUDIO_JUDGE_CRITERIA;
--- a/python/tests/test_audio_example_judge_criteria.py
+++ b/python/tests/test_audio_example_judge_criteria.py
@@ -12,14 +12,21 @@ ways a sibling drifts are precisely the ones text matching misses: a second
 shadows the import, or an import that no longer comes from `helpers`. Every
 `JudgeAgent(...)` call in the file has to be checked, not merely one of them.
 
+The one drift no reading of the example can catch is a mutation of the shared
+object itself, which binds no name and leaves the helper file pristine. That
+one is not checked for, it is made impossible: the criteria are a tuple, and
+the last test here holds them that way.
+
 Deterministic: it parses files. No API keys, no network, no model. Runs in
 `python-ci`'s unit step alongside the rest of `tests/`, unlike the live
 examples it guards.
 """
 
 import ast
+import importlib.util
 import symtable
 from pathlib import Path
+from types import ModuleType
 from typing import Final, NamedTuple
 
 import pytest
@@ -29,6 +36,7 @@ HELPERS_MODULE: Final = "helpers"
 LIST_BUILTIN: Final = "list"
 
 EXAMPLES: Final = Path(__file__).resolve().parents[1] / "examples"
+CRITERIA_MODULE: Final = EXAMPLES / "helpers" / "audio_judge_criteria.py"
 AUDIO_EXAMPLES: Final = ["test_audio_to_text.py", "test_audio_to_audio.py"]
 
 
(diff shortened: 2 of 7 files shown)


javascript/examples/vitest/tests/audio-judge-criteria-parity.test.ts
```diff
@@ -83,6 +95,43 @@ describe("audio example judge criteria — cross-language parity", () => {
...
+        source,
+        `${sibling} declares its own criteria list again. Import AUDIO_JUDGE_CRITERIA instead`,
+      ).not.toContain("AUDIO_JUDGE_CRITERIA = [");
+      // A file can hand one judge the constant and another an inline list, so
+      // the match above is not enough on its own. Same second assertion the JS
+      // siblings carry.
+      expect(
+        source,
+        `${sibling} inlines a judge criterion again. Import AUDIO_JUDGE_CRITERIA instead`,
+      ).not.toContain("The agent's response demonstrates");
```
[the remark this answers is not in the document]
langwatch-agent: Valid, and taken in `ddd1cf35`. You are right that the regex proved only that one text occurrence names the constant. Every drift you listed would have passed it.

Checked in the language where it can be checked properly. `python/tests/test_audio_example_judge_criteria.py` walks each example's AST and asserts:

- every `JudgeAgent(...)` call in the file, not the first one found, passes `criteria` (the guard also fails if a file constructs no judge at all, so it cannot pass by finding nothing)
- that argument resolves to `AUDIO_JUDGE_CRITERIA`, directly or through the `list(...)` copy both examples use
- the name arrives via `from helpers import`, from nowhere else, and unaliased
- nothing rebinds it locally, counting `Assign`, `AnnAssign` and `AugAssign`

It parses files, so no keys and no network, and it runs in `python-ci`'s unit step alongside the rest of `tests/`.

The vitest guard keeps the one check that genuinely spans both languages (byte-identity of the two constants) and points at the Python guard for usage. The weaker text checks are gone rather than left sitting next to a stronger one.

Mutation tested against the cases you named:

```
a second judge in the file gets an inline list      1 failed, 7 passed   CAUGHT
an annotated local rebinding shadows the import     1 failed, 7 passed   CAUGHT
the import stops coming from helpers                1 failed, 7 passed   CAUGHT
the constant is imported under an alias             1 failed, 7 passed   CAUGHT
the judge is given no criteria at all               1 failed, 7 passed   CAUGHT
the judge is given an inline list instead           1 failed, 7 passed   CAUGHT
python constant drifts one word from JS             1 failed | 3 passed  CAUGHT (vitest)
```

`uv run pytest tests/ -m "not integration"`: 1324 passed, 10 skipped.

python/tests/test_audio_example_judge_criteria.py
```diff
@@ -0,0 +1,154 @@
...
+    """
+    if isinstance(value, ast.Name):
+        return value.id == CRITERIA_NAME
+    if (
+        isinstance(value, ast.Call)
+        and _callee_name(value.func) == "list"
+        and len(value.args) == 1
+    ):
+        return resolves_to_shared_constant(value.args[0])
+    return False
```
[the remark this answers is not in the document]
langwatch-agent: Valid, and taken in `ffb23e28`. You are right that checking `Assign`/`AnnAssign`/`AugAssign` covered the obvious shadow and missed every other way to bind a name.

I did not track bindings per call site, though. Scope resolution is the part that would have been easy to get subtly wrong, and the guard does not actually need it: what it wants to say is "this name is bound exactly once in the file, by that import". So it collects every binding the module makes, by any means, and asserts the count and the kind.

```python
def bindings(tree): ...  # ast.Name in Store/Del ctx, ast.arg, alias, ExceptHandler.name
```

`ast.Name` in a `Store`/`Del` context covers assignment targets, `for` targets, walrus, `with ... as` and comprehension targets alike; `ast.arg` covers function and lambda parameters; `alias` covers both import forms; `ExceptHandler` covers `except ... as`. Two assertions follow: `AUDIO_JUDGE_CRITERIA` is bound exactly once and that binding is an `ast.alias`, and `list` is never bound at all, since `criteria=list(AUDIO_JUDGE_CRITERIA)` only means what it looks like while `list` is the builtin.

Mutation tested against each shadow you named, plus two you did not:

```
a parameter shadows the constant           1 failed, 9 passed   CAUGHT
a for-target shadows the constant          1 failed, 9 passed   CAUGHT
a walrus shadows the constant              1 failed, 9 passed   CAUGHT
a with-as shadows the constant             1 failed, 9 passed   CAUGHT
an except-as shadows the constant          1 failed, 9 passed   CAUGHT
a second import rebinds the constant       1 failed, 9 passed   CAUGHT
a lambda parameter shadows the constant    1 failed, 9 passed   CAUGHT
the list builtin is shadowed               1 failed, 9 passed   CAUGHT
```

The earlier eight still hold, and `CRITERIA_NAME`, `HELPERS_MODULE`, `LIST_BUILTIN`, `EXAMPLES` and `AUDIO_EXAMPLES` now carry `Final`.

`uv run pytest tests/ -m "not integration"`: 1326 passed, 10 skipped. `pyright`: 0 errors.

python/tests/test_audio_example_judge_criteria.py
```diff
@@ -0,0 +1,199 @@
...
+        if isinstance(node, ast.Name) and isinstance(node.ctx, (ast.Store, ast.Del)):
+            found.append((node.id, node))
+        elif isinstance(node, ast.arg):
+            found.append((node.arg, node))
+        elif isinstance(node, (ast.Import, ast.ImportFrom)):
+            for alias in node.names:
+                found.append((alias.asname or alias.name.split(".")[0], alias))
+        elif isinstance(node, ast.ExceptHandler) and node.name:
+            found.append((node.name, node))
+    return found
```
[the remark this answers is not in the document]
langwatch-agent: Right again, and I checked before fixing: both escaped.

```
a def declaration shadows the constant      10 passed     ESCAPED
a class declaration shadows the constant    10 passed     ESCAPED
```

Rather than add `FunctionDef`, `AsyncFunctionDef` and `ClassDef` to the walk, I stopped enumerating. That list had already been short twice, and `match` patterns would have been the next omission after these two.

`330f5752` uses `symtable`, the binder CPython itself runs. There is no list of forms to get wrong, it reports nested scopes rather than only the module, and it distinguishes the case directly: a name bound purely by an import reports `imported` and not `assigned`, while anything else binding it as well, `def` and `class` included, turns `assigned` on. So the assertion is now what the rule actually is, "bound in exactly one scope, the module's, by an import and nothing else", rather than an inference from the shapes I happened to think of.

Fifteen binding forms, including the two you found and three you did not:

```
a def declaration shadows the constant        1 failed, 9 passed   CAUGHT
a class declaration shadows the constant      1 failed, 9 passed   CAUGHT
a class declaration shadows list              1 failed, 9 passed   CAUGHT
a match pattern binds the constant            1 failed, 9 passed   CAUGHT
a comprehension target shadows the constant   1 failed, 9 passed   CAUGHT
a nested function rebinds the constant        1 failed, 9 passed   CAUGHT
a parameter shadows the constant              1 failed, 9 passed   CAUGHT
a for-target shadows the constant             1 failed, 9 passed   CAUGHT
a walrus shadows the constant                 1 failed, 9 passed   CAUGHT
a with-as shadows the constant                1 failed, 9 passed   CAUGHT
an except-as shadows the constant             1 failed, 9 passed   CAUGHT
a second import rebinds the constant          1 failed, 9 passed   CAUGHT
a lambda parameter shadows the constant       1 failed, 9 passed   CAUGHT
an annotated local rebinding shadows it       1 failed, 9 passed   CAUGHT
the list builtin is shadowed by assignment    1 failed, 9 passed   CAUGHT
```

The six usage mutations still hold. `uv run pytest tests/ -m "not integration"`: 1326 passed, 10 skipped. `pyright`: 0 errors.

python/tests/test_audio_example_judge_criteria.py
```diff
@@ -0,0 +1,223 @@
...
+    return None
+
+
+def resolves_to_shared_constant(value: ast.expr) -> bool:
+    """
+    True for `AUDIO_JUDGE_CRITERIA` and for the `list(...)` copy of it, which
+    is the form both examples use so the judge cannot mutate the shared list.
+    """
+    if isinstance(value, ast.Name):
+        return value.id == CRITERIA_NAME
```
langwatch-agent, line 81: **P2 · Guard against mutation of the imported criteria object, not just rebinding.**

- **What you are looking at.** This accepts `list(AUDIO_JUDGE_CRITERIA)` whenever the name is still the import. The symbol-table check only rejects bindings of that name.

- **Why it is scary.** `AUDIO_JUDGE_CRITERIA.append("old criterion")` before the judge does not bind the name: CPython reports it as `imported=True, assigned=False`, so every new guard still passes, as does the JS/Python source parity test (it reads the unmodified helper file). Yet `list(AUDIO_JUDGE_CRITERIA)` then hands the mutated, drifted criteria to the live example. Make the exported criteria immutable (and adapt the parity reader) or explicitly reject mutations/other uses of the imported object; add this mutation as a regression test.

langwatch-agent (commented):
One P2 finding — see inline comment.

The audio-to-audio prompt is coherent with all three shared criteria: it asks for the audio’s specific answer and for acknowledgement of its audio format. I also confirmed the five failing JavaScript realtime examples are the same five on main; the deterministic parity test passes there. The remaining finding is a bypass in the new Python enforcement guard.

LangWatch-Review: verdict=findings sha=330f5752faa7f8c70a03fd7ef55c3a660f68f991 p0=0 p1=0 p2=1 p3=0

python/tests/test_audio_example_judge_criteria.py
langwatch-agent: Valid, and taken in `ab2ae7d9`. You are right that every guard here watched the name, and that mutating the object binds no name.

**Reproduced first.** Appending the exact wording #680 removed, at module level in `test_audio_to_audio.py`:

```python
AUDIO_JUDGE_CRITERIA.append("The agent acknowledges what it heard")
```

`10 passed` on the symbol-table guard, and the parity test passed too, since it reads the helper file rather than the live object. The judge would have been handed the drifted criterion with CI fully green.

**Fixed by removing the move, not by adding a check for it.** The Python copy is now a tuple:

```python
AUDIO_JUDGE_CRITERIA: Final[tuple[str, ...]] = (
```

I deliberately did not add an AST check for `.append` / `[0] =` / `.clear()` on the imported name. That is the same enumeration that already came up short twice on this PR, and it is blind to the one-line laundering `stale = AUDIO_JUDGE_CRITERIA; stale.append(...)`. A tuple is not blind to any of them.

**What catches what, measured:**

| mutation | `pytest tests/` | `pyright .` |
|---|---|---|
| helper reverts to a mutable `list` | **FAIL** (1 failed, 10 passed) | pass |
| `AUDIO_JUDGE_CRITERIA.append(...)` in an example | pass | **FAIL** (1 error) |
| `stale = AUDIO_JUDGE_CRITERIA` then `stale.append(...)` | pass | **FAIL** (1 error) |
| `AUDIO_JUDGE_CRITERIA[0] = "anything at all"` | pass | **FAIL** (1 error) |

Being explicit about the split rather than claiming the pytest guard covers it: the guard's job is to keep the constant a tuple, and `python-ci` runs `uv run pytest tests/` and `uv run pyright .` in the same job, with `examples/` inside pyright's scope. Both must stay green, so all four are caught in CI. At runtime the interpreter raises `AttributeError` regardless.

**Regression test added** (`test_the_shared_criteria_cannot_be_mutated_in_place`): loads the helper by path, asserts the object is a tuple with the reason in the message, and asserts the mutation raises. It is the row-one row of that table, so the tuple cannot quietly become a list again.

Two follow-ons from the change: the parity reader now takes a `(` tuple literal and no longer pins the annotation, so renaming the type reads as a rename rather than as drift (proved it still binds by editing one criterion word: `FAIL`, byte-identical assertion). And the JS helper's `as const` gets a note saying it is load-bearing for the same reason, plus its cross-reference corrected to the moved Python path.

Verification: `1327 passed, 10 skipped` in `python/tests/`, `pyright .` 0 errors, parity test 4/4.

python/tests/test_audio_example_judge_criteria.py
```diff
@@ -0,0 +1,279 @@
...
+
+def load_criteria_module() -> ModuleType:
+    """
+    The shared module as the examples get it, loaded from its own path.
+
+    `python/examples` is on `sys.path` when pytest collects the examples, not
+    when it collects this file, so the import goes through the loader directly
+    rather than through a path the test would have to arrange.
+    """
+    spec = importlib.util.spec_from_file_location(
```
langwatch-agent: **P2 · Verify the `helpers` export rather than a separately loaded implementation module.**

- **What you are looking at.** This loads `helpers/audio_judge_criteria.py` directly under a synthetic module name, but both examples use `from helpers import AUDIO_JUDGE_CRITERIA`. The test therefore bypasses `helpers/__init__.py`, despite claiming to load the value “as the examples get it.”

- **Why it is scary.** A future change can keep the implementation tuple and parity file pristine while the façade exports a stale list, e.g. import the tuple as `_shared` in `helpers/__init__.py` and assign `AUDIO_JUDGE_CRITERIA = ["old criterion"]`. The example-source, symtable, direct-module tuple, and JS parity checks all pass; pyright also accepts a normally inferred list export, yet the examples receive the drifted value. Import the real `helpers` package with the example path arranged, or assert its exported object is the immutable implementation object, and add this re-export mutation as a regression case.

langwatch-agent (commented):
One P2 finding — see inline comment.

The tuple fixes the prior direct and alias-based in-place mutation path: tuples of strings cannot be changed at runtime, and the required `uv run pyright .` step covers `python/examples/` (it is not excluded), so pyright is an appropriate backstop for the type-invalid mutation variants. I also confirmed the one failing live realtime JavaScript file is the same as main. The remaining issue is that the new runtime test does not inspect the value the examples actually import.

LangWatch-Review: verdict=findings sha=ab2ae7d966ba51d3f610f2b1f0a0744183c4ab7e p0=0 p1=0 p2=1 p3=0

python/tests/test_audio_example_judge_criteria.py
langwatch-agent: Valid, and taken in `7fe4bec3`. The examples import from `helpers`, the test loaded `helpers/audio_judge_criteria.py`, and `helpers/__init__.py` sat on the path between them unchecked.

**Reproduced first**, with your exact construction:

```python
from .audio_judge_criteria import AUDIO_JUDGE_CRITERIA as _shared

AUDIO_JUDGE_CRITERIA = ["The agent acknowledges what it heard"]
```

`11 passed` and `pyright .` clean, with the criterion #680 removed reaching both examples.

**The guard now imports the package the way the examples do** and asserts identity, not equality:

```python
assert getattr(package, CRITERIA_NAME) is getattr(submodule, CRITERIA_NAME)
```

Equality would accept a facade that rebuilds an equal copy, and a second copy is the thing this whole set of guards exists to prevent. `monkeypatch.syspath_prepend` arranges the example path and puts it back.

**Re-measured, this time on the facade:**

| mutation | `pytest tests/` |
|---|---|
| facade exports a stale list | **FAIL** (both new tests) |
| facade exports a genuine copy, `(*_shared,)` | **FAIL** (identity test) |
| facade exports a copy with one criterion drifted | **FAIL** (identity test) |
| criteria revert to a mutable `list` | **FAIL** (tuple test) |

One result worth recording, because it looked like an escape and is not. My first copy mutation was `AUDIO_JUDGE_CRITERIA = tuple(_shared)`, which passed. That is CPython returning the argument unchanged: `tuple(t) is t` is `True` for an exact tuple, as is `t[:] is t`. There was no copy to catch. `(*_shared,)` does build a new tuple, and the identity check fails on it.

**Cost, stated rather than buried:** importing `helpers` pulls the package's real dependencies, so this file goes from 0.03s to 3s and surfaces scenario's existing `audioop` deprecation warning through `voice/adapters/_twilio_shared.py`. Reading `__init__.py` statically would keep it fast and would be another enumeration to get wrong; the import is what the examples actually get.

Verification: `pyright .` 0 errors, `pytest tests/` 1328 passed / 10 skipped, guard 12/12, and the four rows above. The suite total is unchanged at 95 warnings, so the deprecation was already there.

langwatch-agent (commented):
No findings from re-review.

The package façade is now exercised through the same `from helpers import ...` path as the examples and its export is compared by identity to the criteria submodule, so the stale-list and genuine-copy façade mutations fail. The tuple of strings removes direct and alias-based in-place mutation at runtime; `pyright .` is a required step in the green Python CI job and its configuration includes `examples/`, making it an enforced check for the statically invalid mutation variants. The `tuple(_shared)` note is accurate: CPython returns an exact tuple unchanged, so it was not a copy; the spread-tuple case does create a second object and is caught.

I also verified the current JavaScript failure paths match the cited main run; they are unrelated live realtime examples.

LangWatch-Review: verdict=clean sha=7fe4bec3ac395329e1c3c682302b723a5eccec9b p0=0 p1=0 p2=0 p3=0

langwatch-agent:
## Human Review Brief

> [!CAUTION]
> Merging this changes what an audio example is judged against, and pulls a fourth file into a parity check that keeps two language copies byte-identical. The risk is small and it is the parity machinery, not the criteria: once `test_audio_to_audio.py` is inside that set, any future edit to the criteria in one language fails the build until the other language matches. That is the intent. Just be sure you want the fourth file locked, because unlocking it later is a conversation.

| | |
|---|---|
| **Issue** | Closes #682 |
| **Size** | +403/-44 across 7 files |
| **Mode** | Fast skim |
| **Blast radius** | The audio examples suites in both languages |

<details>
<summary><b>What you are looking at, if the audio examples are not your area</b></summary>

`javascript/examples/` and `python/examples/` are runnable example simulations, not unit tests. They call live models, so they need API keys and do not run in ordinary CI. Four of them are siblings that exercise the same thing from different directions: JS audio-to-audio, JS audio-to-text, Python audio-to-text and Python audio-to-audio.

Each of those hands an LLM judge a list of natural-language criteria, and the wording of those criteria *is* the assertion. That has a history worth having before you read the diff.

- **#655 and #612 aligned three of the four by copy-paste.** Copies drift, and this set drifted.
- **#680 found two latent gaps that had to be fixed in three places at once.** Both were cases where a broken pipeline passed the judge. Criterion 1 used to accept "acknowledges what it heard", so empty audio in and "I received your audio file" out was a pass. Criterion 3 is satisfied by a polite deflection, because "sorry, I can't process audio files" *does* name the non-text format, so criterion 2 has to exclude deflection in so many words rather than leaning on the judge's reading of "refusal". That is why the wording is so defensive and so specific.
- **The two JavaScript siblings were then given one shared constant**, and `audio-judge-criteria-parity.test.ts` was written to hold the Python copy byte-identical to it. That guard is deliberately deterministic, a file read and a parse with no keys and no network, so it runs in CI even though everything it protects cannot.

The Python list is written with double-quoted string literals *on purpose*, so that dropping the trailing comma makes it parse as JSON from the JavaScript side. That is a real constraint on how the Python file may be formatted, and there is a note next to the declaration saying so.

What this PR is about is that the parity guard only ever knew about three siblings, so the fourth kept its pre-alignment criteria and nothing noticed. Which explains the shape of the change: a Python-side shared constant to match the JavaScript one, the guard repointed at the constant rather than at one example file, and a new Python test that walks each example's AST and inspects every `JudgeAgent` call, because text matching from the JavaScript side cannot see a second judge with an inline list, a local rebinding that shadows the import, or an import that quietly stopped coming from `helpers`.

</details>

<details>
<summary><b>What was wrong, and the good bit about why</b></summary>

`python/examples/test_audio_to_audio.py` was still judged against pre-alignment criteria, one of which was "The agent identifies or guesses the voice is male". That is voice-gender classification: non-deterministic, and not the capability the example is meant to test. A flaky assertion about something nobody meant to assert.

The interesting part is why it survived. #680 already built the machinery for exactly this problem: shared criteria in `audio-judge-criteria.ts`, and a parity test that keeps the Python copy byte-identical and asserts both JavaScript siblings pass the constant to their judge rather than an inline array. That machinery covered three siblings. This file was not one of them, so it drifted for a year with nothing watching.

The ticket's suggested fix was "update it to use the same criteria", which fixes today's file. Adding it to the parity set is what stops the next one, and that is the choice made here.
</details>

<details>
<summary><b>What to look at</b></summary>

`python/tests/test_audio_example_judge_criteria.py` and the JavaScript parity test are the durable part. Confirm the parity assertion covers all four siblings now and would actually fail if someone edited one copy, since a parity test that silently passes on a missing file is how this drifted in the first place.

The criteria change itself is prose in a constant. Read it once for whether the remaining criteria describe capabilities the example genuinely exercises.
</details>

> [!NOTE]

drewdrewthis:
## Verdict: READY

PR: https://github.com/langwatch/scenario/pull/936
Head SHA: 107a5faa5c69898b9c0407197099b54c0b230188

**Summary:** Fourth audio example added, matches the pattern of its three siblings. 12/12 tests pass. CI green (all gating checks SUCCESS or intentionally SKIPPED).

No blocking findings. No inline `review-thread` comments needed — nothing crossed the Fix/New-AC bar.

<details>
<summary>Evidence</summary>

- CI: python-complete, javascript-complete, ci-checks (24.x), docs-complete, test (3.12), CodeRabbit, Validate PR Title, Analyze (python/js-ts) — all SUCCESS at this head SHA.
- Human Review Brief: https://github.com/langwatch/scenario/pull/936#issuecomment-5412905664
- Attribution: this session did not author any content in this PR's diff.

</details>

drewdrewthis (commented):
Reviewed. No blocking findings — the fourth sibling is brought in consistently with the other three (same judge-criteria shape, same shared source, parity test covers all four, CI green). Full verdict: https://github.com/langwatch/scenario/pull/936#issuecomment-5494702322

Issue #893, opened 2026-08-11
ci: fork PRs changing python/examples/** go green with nothing exercising the change


drewdrewthis:
## Summary

**Fork PRs that change `python/examples/**` go green with no job exercising the changed code.**

`Test (Examples)` is the only job that runs anything under `python/examples/`. PR #739 taught that step to skip on fork PRs, to stop the module-level `AsyncOpenAI()` collection hard-fail described in #858. The guard is correct for the failure it targets, but it leaves external-contributor changes to `examples/` verified by nothing, while the required check still reports green.

## Expected

A green required check on a PR that changes `python/examples/**` means something ran against that change.

## Actual

Nothing runs against it, and the board is green anyway:

| element | behaviour |
|---|---|
| `Test (Examples)` | **skipped** on fork PRs (the #739 guard) |
| `test` job | runs `pytest tests/`, which never imports `examples/` |
| `python-complete` | accepts `skipped` as a non-failure, so the rollup goes **green** |

Live case: **#840** changes `python/examples/voice/_bot/bot.py` (+9/-2) and adds `scripts/test-voice-pipecat-lifecycle.sh` (+78). Its green run at head SHA `f9e3e0418ed1a5ba2f62925742fd1feed32f829e` reports `1212 passed, 25 skipped, 35 deselected` — none of which touch either file. The added shell script is invoked by no workflow at all.

For an external contributor, a green board on an `examples/` change carries no signal. Contributor and reviewer both see a pass that means only "the unrelated unit suite still works".

## To reproduce

1. Open a PR from a fork that modifies any file under `python/examples/`.
2. Wait for `python-ci` to conclude.
3. Observe `Test (Examples)` skipped, `python-complete` green, and no executed test covering the changed file.

## Environment

`.github/workflows/python-ci.yml` — the `Test (Examples)` step guard:

```
if: github.actor != 'dependabot[bot]' && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository)
```

Related: #858 (the collection hard-fail this guard was added for), #739 (the guard), #867 (one example dir type-checked and tested by nothing), #654 (voice-integration step permanently skipped — closed, same shape).

---
*Classification: Bug*
*Status: stub — investigation pending. Issue is not-ready until the investigation step completes.*

[...]

PR #944, opened 2026-08-25
fix(ci): a shared budget outage stops blocking unrelated pull requests (#892)

### LINKED ISSUE #892: Required CI gate blocked by examples suite exhausting the LLM daily budget (402 budget_exceeded)
## What this is

A required merge gate can fail for reasons unrelated to the pull request: the `Test (Examples)` step makes real LLM calls through the LangWatch gateway on a virtual key with a **daily spend budget**, and when that budget is exhausted every example test returns HTTP `402 budget_exceeded` and the whole suite goes red. Because `javascript-complete` / `python-complete` (the required checks) depend on the jobs that run this step, an unrelated PR is blocked from merging even though its own tests pass. This is a structural CI problem, not a bug in any single test (distinct from #882, which is one specific voice test failing its own assertion).

## Evidence

Observed on PR #889 at head `77ce1dbd` (a judge-logic change that touches no example and makes no gateway calls in its own tests):

| Job | Step | Result |
|---|---|---|
| `ci-checks (24.x)` | core `Tests` | **1088 passed**, 4 skipped |
| `ci-checks (24.x)` | `Test (Examples)` | 26 of 55 files fail, 36 tests |
| `test (3.12)` | core `Run tests` | **1219 passed**, 0 failed |
| `test (3.12)` | `Test (Examples)` | 26 fail |

Every example failure traces to one upstream response:

```
HTTP 402  {"error":{"type":"budget_exceeded","fault":"customer",
  "meta":{"budget_scope":"virtual_key","budget_window":"day"},
  "message":"The virtual key spending limit (per day) for this request has been reached."}}
```

~200 occurrences of that message in a single JS job log. The failures span agents with no relation to each other or to the PR (`weather-agent`, `vegetarian-recipe-agent`, `database-tool-mocking`, `audio-judge-*`, `voice/*`, `test_azure_api_gateway`), which is the signature of an account-level cap rather than a code regression. (A secondary, smaller batch — `Remote agent returned HTTP 500` — comes from a demo remote endpoint being down; same class of external dependency.)

## Root cause

`Test (Examples)` is an end-to-end suite that exercises real models through the LangWatch gateway using a shared virtual key that carries a per-day budget. The step is part of `ci-checks (24.x)` / `test (3.12)`, and those feed the **required** `javascript-complete` / `python-complete` gates. So merge-readiness is coupled to a consumable, time-windowed account resource that any PR (or repeated CI runs) can exhaust. Once exhausted, no code change on a branch can turn the gate green — only a budget reset or raise.

## Why it matters

- A green-tested PR cannot merge until an unrelated external budget refills — readiness is hostage to account spend, not to the diff.
- It is indistinguishable at a glance from a real failure, so every contributor has to open logs to confirm "not my change."
- It compounds: each re-run to "try again" spends more budget, making the window harder to clear.

## Potential solutions (not yet chosen)

1. **Make `Test (Examples)` non-required.** Keep it running and visible, but drop it (and only it) from the required set so the real unit/execution suites gate merges. Lowest effort; the examples become an informational signal.
2. **Give examples their own uncapped or separately-budgeted key** (or a higher daily limit) so normal PR traffic can't exhaust it.
3. **Run examples on demand / nightly**, not per-PR — a scheduled workflow or a label-triggered run, so per-PR CI depends only on deterministic suites.
4. **Mock the model boundary for most examples** (cheap/stub model), reserving real-gateway runs for a small curated set that runs less often.
5. **Fail-soft on `402`**: detect `budget_exceeded` and mark the step skipped/neutral rather than failed, so a spend cap doesn't read as a code failure.

Options 1 and 2 are the smallest changes that fix the merge-blocking behavior; 3/4 reduce spend pressure structurally.

## Acceptance criteria

- A PR whose unit/execution suites pass can merge without being blocked by an exhausted example-suite budget.
- A genuine example-test regression is still surfaced (not silently hidden).
- Whatever path is chosen is documented so contributors know a red `Test (Examples)` on a budget cap is not their change.

langwatch-agent:
## Ticket

Closes #892. The examples suite makes real model calls through the LangWatch AI Gateway on one virtual key with a daily spend cap. When the cap is hit the gateway answers HTTP 402 `budget_exceeded` and every test in the suite fails at once. That suite feeds a required check, so an unrelated pull request is blocked from merging by a shared resource it never touched.

The ticket's evidence, PR #889 at `77ce1dbd`: 1088 JavaScript tests passed, 1219 Python tests passed, and 26 example files failed on 402.

## Change

Both examples suites now run behind `scripts/ci/examples-suite-gate.sh`, which separates "the suite could not run" from "the suite ran and disagreed":

| situation | verdict |
|---|---|
| the suite passed | pass |
| failed, nothing names the budget | **fail**, a real defect |
| failed on budget, PR changes the examples | **fail**, the change is unverified |
| failed on budget, PR changes nothing the suite covers | warn and pass |

## The part I want argued with

**A suite that cannot fail is the disease, not the cure.** #893 in this repo records fork PRs reaching a green required check with nothing having read the changed file, and the `Collect (Examples)` step now on `main` is the fix for it. Loosening this gate risks recreating exactly that.

Three things hold the line, and they are the ones to attack:

1. **A budget outage never reports as a pass.** It emits `::warning::` and writes `$GITHUB_STEP_SUMMARY`, both saying the suite did **not** run. There is a test asserting the summary says "not run"/"unverified" rather than implying success.
2. **A PR that changes the examples stays red through an outage.** This is the rule that stops a day-long budget exhaustion becoming a day-long free pass on example changes. It is why the gate needs to know what the PR touched, and why it refuses to run without being told (`EXAMPLES_TOUCHED:?`) rather than guessing a default.
3. **Only the budget is tolerated.** The classification requires the gateway's own wording, `budget_exceeded` or `budget_scope`. Any failure without it exits with the suite's status.

If you think case 4 should stay red too, that is a legitimate position and the change is one line; but then #892 is only closed by raising the cap.

## The trap this surfaced

Telling "touches the examples" apart needed a new `examples` path filter. `.github/actions/detect-changes/action.yml` documents what happens if you add one carelessly:

> every filter key the caller wants to consume must be declared as an `outputs.<key>` entry below AND echoed into the force step ... silently missing entries land as empty strings.

Empty is worse than either `true` or `false`: the filter looks wired and answers nothing. Nothing enforced that comment, so `scripts/validate-examples-gate.sh` now checks it for **every** filter key in every workflow, not just the one this PR adds. It currently covers 5 key/workflow pairs across `docs-ci`, `javascript-ci` and `python-ci`.

Both new scripts run in `meta-ci`, alongside the existing validators and with the same `!cancelled()` treatment so one push surfaces every verdict.

## Evidence

- **10 gate-decision checks**, driving the real script with synthetic suite outcomes: `bash scripts/test-examples-budget-gate.sh`.
- **8 wiring checks**: `bash scripts/validate-examples-gate.sh`.
- The pre-existing `validate-aggregator-workflows.sh` still passes.
- `shellcheck` clean on all three new scripts; `actionlint` reports the same single pre-existing SC2086 on `python-ci.yml` before and after.

### Mutations, all 8 caught

| mutation | caught by |
|---|---|
| tolerate a budget outage even when the PR changes examples | gate tests |
| tolerate any failure (drop the budget classification) | gate tests |
| stop writing the run summary | gate tests |
| default `EXAMPLES_TOUCHED` instead of requiring it | gate tests |
| action stops declaring the `examples` output | wiring validator |
| action stops forcing it on non-PR events | wiring validator |
| a workflow goes back to running the suite raw | wiring validator |
| a gate call drops `EXAMPLES_TOUCHED` | wiring validator |

Both harnesses assert how many checks they ran. That is not decoration: the first version of the gate harness returned a temp path on stdout, and capturing it with `$(...)` swallowed every PASS line along with it. Six cases ran and printed nothing while the closing line still said all checks passed.

## Review round

Three findings, all taken.

**The JavaScript fork gap (P1).** A fork pull request that changes `javascript/examples/**` went green with nothing having read the changed file. The examples suite is the only JavaScript step that reads them, it is skipped without secrets, and `javascript-complete` accepts a job whose every examples step was skipped. That is #893, which was closed on the Python side alone. This PR did not create the gap, but it is the same required check, so leaving it meant leaving a known hole in the thing being repaired.

`List (Examples)` now runs `pnpm -F vitest-examples exec vitest list` on the exact inverse condition of `Test (Examples)`, mirroring Python's `Collect (Examples)`. It imports every example to enumerate its tests and makes no provider call. Verified locally with every model key unset: a clean tree exits 0, and an unresolvable import in one example exits 1 naming the file. It is weaker than running the suite and the comment says so, catching an import that does not resolve, a syntax error or a renamed export, but not a wrong assertion.

The validator now **proves** the pairing rather than checking that a fallback exists, since two conditions can both be false and leave the identical hole. It parses both `if:` expressions and evaluates them over every assignment of the contexts they name, requiring exactly one to be true in each.

**`EXAMPLES_TOUCHED` was fail-open.** Everything that was not exactly `true` took the lenient branch, so `tru` from a mis-wired filter key would tolerate an outage on a pull request that does change the examples. Only `true` and `false` are accepted now.

**The report is read defensively.** It is written while pull-request-controlled examples run. A `<!DOCTYPE` is now refused, which no test runner writes and which is how internal entity expansion is smuggled in, and the read is capped. `defusedxml` is deliberately not used: the gate invokes this script with the runner's bare `python3` rather than `uv`, so the import would not resolve. Whoever can write that file already runs the suite on the runner, so this is depth rather than a boundary, and both refusals are fail-closed.

### Updated evidence

- **24 gate-decision checks**: `bash scripts/test-examples-budget-gate.sh`.
- **10 wiring checks**: `bash scripts/validate-examples-gate.sh`.

Six further mutations, all caught:

| mutation | caught by |
|---|---|
| accept `tru` / `True` / `yes` / `1` as `EXAMPLES_TOUCHED` | gate tests |
| drop the doctype refusal | gate tests |
| drop the size cap | gate tests, via the diagnostic |
| remove the JavaScript fork fallback | wiring validator |
| give the fallback a condition that is not the complement | wiring validator |
| make the fallback read secrets | wiring validator |

The size-cap mutation escaped at first and the fix was to assert the diagnostic rather than to claim the check mattered more than it does. The read is already bounded, so removing the cap leaves a truncated document that fails to parse and is refused anyway. What it costs is knowing why, and "larger than it can be" and "malformed" send a maintainer to different places.

## Advisory note

### DIFF
--- a/.github/workflows/javascript-ci.yml
+++ b/.github/workflows/javascript-ci.yml
@@ -44,6 +44,10 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       relevant: ${{ steps.detect.outputs.relevant }}
+      # Whether this PR changes anything the examples suite covers. The gate
+      # in front of that suite needs it to decide whether a gateway budget
+      # outage is collateral or a genuine loss of coverage.
+      examples: ${{ steps.detect.outputs.examples }}
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v3
       - uses: ./.github/actions/detect-changes
@@ -53,6 +57,8 @@ jobs:
             relevant:
               - 'javascript/**'
               - '.github/workflows/javascript-ci.yml'
+            examples:
+              - 'javascript/examples/**'
 
   ci-checks:
     needs: changes
@@ -126,8 +132,11 @@ jobs:
       # Examples — skipped for Dependabot PRs and fork PRs since they don't have access to repo secrets
       - name: Test (Examples)
         if: github.actor != 'dependabot[bot]' && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository)
-        run: pnpm -F vitest-examples test
+        # Wrapped so a shared-budget outage is told apart from a broken
+        # example. See scripts/ci/examples-suite-gate.sh.
+        run: ../scripts/ci/examples-suite-gate.sh pnpm -F vitest-examples test
         env:
+          EXAMPLES_TOUCHED: ${{ needs.changes.outputs.examples }}
           LANGWATCH_API_KEY: ${{ secrets.LANGWATCH_API_KEY }}
           OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
         working-directory: javascript
--- a/.github/workflows/javascript-ci.yml
+++ b/.github/workflows/javascript-ci.yml
@@ -144,6 +144,25 @@ jobs:
           OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
         working-directory: javascript
 
+      # The inverse condition of the step above, so exactly one of the two runs.
+      # Without it a fork pull request that changes javascript/examples/** goes
+      # green with nothing having read the file it changed: the examples suite
+      # is the only JavaScript step that reads them, it is skipped for forks,
+      # and javascript-complete accepts a job whose steps were all skipped.
+      # That is #893, which was closed on the Python side alone by its
+      # `Collect (Examples)` fallback. This is the JavaScript half.
+      #
+      # `vitest list` imports every example file to enumerate its tests and
+      # makes no provider call, so it needs no secrets. It catches what a fork
+      # can break without any key: an import that does not resolve, a syntax
+      # error, a renamed export. It does not catch a wrong assertion, so this
+      # is weaker than running the suite, and is deliberately not presented as
+      # equivalent.
+      - name: List (Examples)
+        if: github.actor == 'dependabot[bot]' || (github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name != github.repository)
+        run: pnpm -F vitest-examples exec vitest list
+        working-directory: javascript
+
   # One required check for the whole workflow. Reports success if every
   # upstream job either passed or was legitimately skipped (paths didn't
   # match). Reports failure if any upstream job failed.
--- a/.github/workflows/javascript-ci.yml
+++ b/.github/workflows/javascript-ci.yml
@@ -134,9 +134,12 @@ jobs:
         if: github.actor != 'dependabot[bot]' && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository)
         # Wrapped so a shared-budget outage is told apart from a broken
         # example. See scripts/ci/examples-suite-gate.sh.
-        run: ../scripts/ci/examples-suite-gate.sh pnpm -F vitest-examples test
+        # The junit reporter is what lets the gate decide per failed test
+        # rather than by grepping the log. See the python-ci counterpart.
+        run: ../scripts/ci/examples-suite-gate.sh pnpm -F vitest-examples test --reporter=junit --outputFile="$EXAMPLES_REPORT"
         env:
           EXAMPLES_TOUCHED: ${{ needs.changes.outputs.examples }}
+          EXAMPLES_REPORT: ${{ runner.temp }}/examples-junit.xml
           LANGWATCH_API_KEY: ${{ secrets.LANGWATCH_API_KEY }}
           OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
         working-directory: javascript
--- a/scripts/ci/classify-examples-failures.py
+++ b/scripts/ci/classify-examples-failures.py
@@ -16,16 +16,51 @@ suites either, since pytest puts the reason on its summary line and vitest does
 not.
 """
 
+import re
 import sys
 import xml.etree.ElementTree as ET
 
 BUDGET_MARKERS = ("budget_exceeded", "budget_scope")
 
+# A JUnit report from pytest or vitest is kilobytes. Anything past this is not
+# one, and reading it would be the denial of service rather than the defence.
+MAX_REPORT_BYTES = 16 * 1024 * 1024
+
+DOCTYPE = re.compile(r"<!DOCTYPE", re.IGNORECASE)
+
 
 def classify(path: str) -> str:
+    """Read the report defensively, then decide.
+
+    `xml.etree` resolves no external entities, but it does expand internal
+    ones, so a report carrying a `<!DOCTYPE>` with nested entity definitions
+    can be made to expand to gigabytes. Neither pytest nor vitest ever writes a
+    doctype into a JUnit report, so refusing one closes that off entirely and
+    costs nothing. That is preferred here over `defusedxml`: this script runs
+    under the runner's bare `python3` rather than under `uv`, so a third-party
+    import would not resolve at all.
+
+    The threat is modest either way: whoever can write this file already runs
+    the test suite on this runner. This is depth rather than the only barrier,
+    and refusing is fail-closed, since the gate reads `unreadable` and keeps
+    the suite's own red exit status.
+    """
     try:
-        root = ET.parse(path).getroot()
-    except (OSError, ET.ParseError) as error:
+        with open(path, "rb") as handle:
+            raw = handle.read(MAX_REPORT_BYTES + 1)
+    except OSError as error:
+        return f"unreadable {error}"
+
+    if len(raw) > MAX_REPORT_BYTES:
+        return f"unreadable report is larger than {MAX_REPORT_BYTES} bytes"
+
+    report_text = raw.decode("utf-8", errors="replace")
+    if DOCTYPE.search(report_text):
+        return "unreadable report carries a doctype, which no test runner writes"
+
+    try:
+        root = ET.fromstring(report_text)
+    except ET.ParseError as error:
         return f"unreadable {error}"
 
     budget = 0
--- a/scripts/ci/classify-examples-failures.py
+++ b/scripts/ci/classify-examples-failures.py
@@ -1,0 +1,57 @@
+"""Classify every failed test in a JUnit report as a gateway budget refusal or not.
+
+Prints one word plus detail, for the shell gate to branch on:
+
+    budget <n>     every failed test was refused on the gateway budget
+    mixed <names>  at least one failure was NOT a budget refusal
+    nofailures     the report lists no failed test at all
+    unreadable <e> the report could not be read
+
+EVERY failure, not any failure. A budget outage and a genuinely broken example
+can land in the same run: the tests that reach the gateway get 402 while one
+that fails on its own assertion fails for its own reason. Deciding on "the
+output mentions the budget somewhere" would hide that real failure, which is
+why this reads a report rather than the log. Grepping cannot do it across both
+suites either, since pytest puts the reason on its summary line and vitest does
+not.
+"""
+
+import sys
+import xml.etree.ElementTree as ET
+
+BUDGET_MARKERS = ("budget_exceeded", "budget_scope")
+
+
+def classify(path: str) -> str:
+    try:
+        root = ET.parse(path).getroot()
+    except (OSError, ET.ParseError) as error:
+        return f"unreadable {error}"
+
+    budget = 0
+    other: list[str] = []
+    for case in root.iter("testcase"):
+        problems = list(case.findall("failure")) + list(case.findall("error"))
+        if not problems:
+            continue
+        text = " ".join(
+            f"{p.get('message') or ''} {p.text or ''}" for p in problems
+        ).lower()
+        if any(marker in text for marker in BUDGET_MARKERS):
+            budget += 1
+        else:
+            name = f"{case.get('classname') or ''}::{case.get('name') or '?'}"
+            other.append(name.strip(":"))
+
+    if other:
+        return "mixed " + "; ".join(other[:5])
+    if budget:
+        return f"budget {budget}"
+    return "nofailures"
+
+
+if __name__ == "__main__":
+    if len(sys.argv) != 2:
+        print("usage: classify-examples-failures.py <junit.xml>", file=sys.stderr)
+        raise SystemExit(2)
+    print(classify(sys.argv[1]))
--- a/scripts/ci/examples-suite-gate.sh
+++ b/scripts/ci/examples-suite-gate.sh
@@ -1,0 +1,92 @@
+#!/usr/bin/env bash
+# Run the examples suite, and tell a shared-budget outage apart from a broken
+# example.
+#
+# THE PROBLEM. The examples suite makes real model calls through the LangWatch
+# AI Gateway on one virtual key with a daily spend budget. When that budget is
+# exhausted the gateway answers HTTP 402 `budget_exceeded` and every test in
+# the suite fails at once. The suite feeds a required check, so an unrelated
+# pull request whose own tests all pass is blocked from merging by a shared
+# resource it never touched. Observed on #889, whose 1088 JavaScript and 1219
+# Python tests passed while 26 example files failed on 402.
+#
+# WHAT THIS DOES NOT DO. It does not turn the suite green and walk away. A
+# suite that cannot fail is the failure mode this repository has already been
+# bitten by: #893 records fork pull requests reaching a green required check
+# with nothing having run against the changed file. So the budget case is
+# reported loudly, in the log and in the run summary, and it is tolerated only
+# where tolerating it costs no coverage.
+#
+# THE RULE. A budget outage means the suite could not be run at all, not that
+# it ran and disagreed:
+#
+#   * the suite passed                          -> pass
+#   * failed, no budget signal                  -> FAIL, this is a real defect
+#   * failed on budget, PR changes the examples -> FAIL, the change is
+#                                                  unverified and saying
+#                                                  otherwise would be a lie
+#   * failed on budget, PR changes nothing the
+#     suite covers                              -> warn and pass, the suite
+#                                                  was never evidence about
+#                                                  this PR
+#
+# The third case is the one that keeps this honest. Budget exhaustion lasts
+# until the daily window resets, so without it an examples change could merge
+# on a day-long green that ran nothing.
+#
+# Required environment:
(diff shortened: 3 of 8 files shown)


scripts/ci/examples-suite-gate.sh
```diff
@@ -0,0 +1,92 @@
...
+  exit 0
+fi
+
+# The gateway's own wording for a hard budget breach. `budget_scope` appears in
+# `error.meta` on the same response, so either is proof the refusal came from
+# the budget rather than from the example.
+if ! grep -qiE 'budget_exceeded|budget_scope' "$output"; then
+  echo "::error::The examples suite failed and nothing in its output names the gateway budget, so this is a real failure."
+  exit "$status"
+fi
```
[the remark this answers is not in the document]
langwatch-agent: Valid, and it is the sharpest thing anyone could have said about this PR. Fixed in a78ad682.

You are right about the mechanism and about the consequence. `budget_exceeded` anywhere in the output classified the whole run, so a run where the gateway refused most tests on budget AND one example was genuinely broken exited 0 under `EXAMPLES_TOUCHED=false` and buried the real failure. That is worse than the blocking this PR set out to fix, and I reasoned past it while writing the thing: I told myself budget exhaustion fails everything anyway, which is true of the tests that reach the gateway and says nothing about a test that fails before it gets there.

I did not take the suggested diff as written, because grepping cannot carry the rule across both suites: pytest puts the reason on its `FAILED` summary line and vitest does not, so "every reported failure names the budget" would be true by construction for pytest and false by construction for vitest. Attributing per failure needs a report.

So both suites now write JUnit XML (`--junitxml` for pytest, `--reporter=junit` for vitest), and `scripts/ci/classify-examples-failures.py` reads it. Every failed `<testcase>` must carry a budget refusal for the run to be tolerated; if any does not, the gate exits red and the message names the offending tests.

Three cases now fail closed that the text version would have waved through or guessed at:

- **mixed**, the case you raised, red whatever `EXAMPLES_TOUCHED` says
- **an unreadable or missing report**, red, since the claim cannot be checked
- **a report listing no failure while the suite exited non-zero**, red, since nothing there can be attributed to the budget

And the gate now *requires* `EXAMPLES_REPORT` rather than falling back to the old leniency when it is absent.

Your regression case is in, in both directions:

```
=== budget refusals mixed with a real failure ===
  PASS: a mixed run stays red even when the PR touches nothing (exit 1)
  PASS: a mixed run stays red when the PR touches examples (exit 1)
```

17 gate checks now, up from 10, and 8 mutations each caught. One is worth reporting because it needed a new assertion to become reachable at all: deleting the `case` arm for the mixed verdict does not change the verdict, since the statement is fail-closed and the default arm also exits red. What it costs is the diagnostic naming the broken test, which is the actionable half, so that is now asserted rather than assumed.

.github/workflows/javascript-ci.yml
```diff
@@ -126,8 +132,14 @@ jobs:
       # Examples — skipped for Dependabot PRs and fork PRs since they don't have access to repo secrets
       - name: Test (Examples)
```
langwatch-agent, line 133: **P1 Ensure JavaScript example changes cannot pass unexamined from a fork**

This step is still skipped for every fork pull request. A fork can change `javascript/examples/**`: `changes` reports `examples=true` and `ci-checks` runs, but this is the only JavaScript examples step, so nothing reads the changed examples and `javascript-complete` still accepts the successful job. Python has the inverse-condition `Collect (Examples)` fallback for exactly this no-secrets path, but JavaScript has none. That leaves the fork-green-with-nothing-run failure mode intact for JavaScript examples, including during a budget outage. Add an equivalent secret-free JavaScript discovery or collection fallback, or fail example-changing fork PRs explicitly.

langwatch-agent (commented):
Found one blocking coverage gap.

The per-test gate is correctly fail-closed for absent or unreadable reports, and its supported Vitest JUnit output path places failure data in the elements the classifier reads. The supplied gate and wiring harnesses pass locally.

LangWatch-Review: verdict=findings sha=a78ad682e340ad179a60e3346b8a392cb409b97b p0=0 p1=1 p2=0 p3=0

.github/workflows/javascript-ci.yml
```diff
@@ -126,8 +132,14 @@ jobs:
...
         if: github.actor != 'dependabot[bot]' && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository)
-        run: pnpm -F vitest-examples test
+        # Wrapped so a shared-budget outage is told apart from a broken
+        # example. See scripts/ci/examples-suite-gate.sh.
+        # The junit reporter is what lets the gate decide per failed test
+        # rather than by grepping the log. See the python-ci counterpart.
+        run: ../scripts/ci/examples-suite-gate.sh pnpm -F vitest-examples test --reporter=junit --outputFile="$EXAMPLES_REPORT"
```
langwatch-agent, line 139: **P1 Ensure JavaScript example changes cannot pass unexamined from a fork**

This step is still skipped for every fork pull request. A fork can change `javascript/examples/**`: `changes` reports `examples=true` and `ci-checks` runs, but this is the only JavaScript examples step, so nothing reads the changed examples and `javascript-complete` still accepts the successful job. Python has the inverse-condition `Collect (Examples)` fallback for exactly this no-secrets path, but JavaScript has none. That leaves the fork-green-with-nothing-run failure mode intact for JavaScript examples, including during a budget outage. Add an equivalent secret-free JavaScript discovery or collection fallback, or fail example-changing fork PRs explicitly.

langwatch-agent (commented):
Finding location correction: the P1 from the preceding review is attached below to the changed line it concerns.

LangWatch-Review: verdict=findings sha=a78ad682e340ad179a60e3346b8a392cb409b97b p0=0 p1=1 p2=0 p3=0

scripts/ci/examples-suite-gate.sh
```diff
@@ -0,0 +1,111 @@
...
+# Required environment:
+#   EXAMPLES_TOUCHED   "true" when this PR changes files the suite covers
+#   EXAMPLES_REPORT    path to the JUnit XML the suite writes
+# Optional:
+#   GITHUB_STEP_SUMMARY  written to when set, so the reason survives the log
+#
+# Usage: examples-suite-gate.sh <command> [args...]
+set -uo pipefail
+
+: "${EXAMPLES_TOUCHED:?}" "${EXAMPLES_REPORT:?}"
```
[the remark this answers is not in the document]
langwatch-agent: Valid, and it is the fail-open direction: everything that is not exactly `true` took the lenient branch, so `tru` from a mis-wired filter key would tolerate an outage on a pull request that does change the examples. Fixed in b7401969. Only `true` and `false` are accepted; anything else exits 2 before the suite runs. The harness pins `tru`, `True`, `yes` and `1`, and a mutation that adds them back to the accepted set is caught.

scripts/ci/classify-examples-failures.py
```diff
@@ -0,0 +1,57 @@
...
+import xml.etree.ElementTree as ET
+
+BUDGET_MARKERS = ("budget_exceeded", "budget_scope")
+
+
+def classify(path: str) -> str:
+    try:
+        root = ET.parse(path).getroot()
+    except (OSError, ET.ParseError) as error:
+        return f"unreadable {error}"
```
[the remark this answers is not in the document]
langwatch-agent: Hardened in b7401969, without the dependency.

`defusedxml` is not usable here: the gate invokes this script as `python3 $SCRIPT_DIR/classify-examples-failures.py`, the runner's interpreter, not `uv run`. Adding it to python/pyproject.toml would not put it on that interpreter's path, so the import would fail and the gate would lose its classifier entirely.

What the rule is actually pointing at is internal entity expansion, since `xml.etree` resolves no external entities. That needs a `<!DOCTYPE`, and neither pytest nor vitest ever writes one into a JUnit report, so the report is now refused outright if it carries one, and the read is capped at 16 MiB. That closes the vector rather than suppressing the warning.

Worth stating plainly: whoever can write this file already runs the test suite on this runner, so this is depth rather than a boundary. Both refusals are fail-closed, the gate reads `unreadable` and keeps the suite's own red exit status. The harness covers a doctype report and an oversized one, and mutations removing either check are caught.

.github/workflows/javascript-ci.yml
langwatch-agent: Confirmed and fixed in b7401969. You are right that it was left intact, and it is worth being precise that this pull request did not create the gap: before it, the JavaScript examples step was skipped for forks in exactly the same way. But it is the same required check, so leaving it was leaving a known hole in the thing being repaired.

`List (Examples)` now runs on the exact inverse condition of `Test (Examples)`, mirroring Python's `Collect (Examples)`. It runs `pnpm -F vitest-examples exec vitest list`, which imports every example file to enumerate its tests and makes no provider call. Verified locally with every model key unset: clean tree exits 0, and an unresolvable import in one example exits 1 naming the file.

It is weaker than running the suite, and the comment says so: it catches an import that does not resolve, a syntax error, a renamed export, and not a wrong assertion.

The wiring validator now proves the pairing instead of asserting a fallback exists, since two conditions can both be false and leave the identical hole. It parses both `if:` expressions and evaluates them over every assignment of the contexts they name, requiring exactly one to be true in each. Both workflows pass. Three mutations are caught: removing the step, giving it a condition that is not the complement, and making it read secrets, which would skip it in precisely the case it exists to cover.

.github/workflows/javascript-ci.yml
langwatch-agent: Same finding as the thread the reviewer reattached to the changed line, answered there. Fixed in b7401969: `List (Examples)` runs `vitest list` on the exact inverse condition of `Test (Examples)`, and the wiring validator now proves the two partition every context.

langwatch-agent:
## Human Review Brief

> [!CAUTION]
> Merging this gives a required check a way to go green while its tests are red. That is the entire point of the change and it is defensible, but it means the review question is narrow and important: exactly when does "warn and pass" fire, and can a genuine example break slip through that door? If the "PR changes nothing the suite covers" test is even slightly too generous, real breakage merges on a budget outage.

| | |
|---|---|
| **Issue** | Closes #892 |
| **Size** | +764/-2 across 8 files |
| **Mode** | Targeted, on the classifier |
| **Blast radius** | The required examples check on every PR in this repo |

<details>
<summary><b>The problem, which is real</b></summary>

The examples suite makes live model calls through the AI Gateway on one shared virtual key with a daily spend cap. When the cap is hit the gateway returns HTTP 402 `budget_exceeded` and every test in the suite fails at once. Because that suite feeds a required check, one exhausted shared budget blocks every open pull request, including ones that touch nothing the suite covers.

The cited evidence: PR #889 at `77ce1dbd` had 1088 JavaScript tests pass, 1219 Python tests pass, and 26 example files fail on 402.
</details>

<details>
<summary><b>The four-way decision table, and where it can be wrong</b></summary>

| situation | verdict |
|---|---|
| suite passed | pass |
| failed, nothing names the budget | fail |
| failed on budget, PR changes the examples | fail |
| failed on budget, PR changes nothing the suite covers | warn and pass |

Row 2 is the safety net: an ordinary failure still fails. Row 3 is the good instinct: if you touched the examples, a budget outage does not excuse you from proving them. Row 4 is the hole, and it is intentional.

So read two files against each other. `scripts/ci/classify-examples-failures.py` decides whether a failure "names the budget", and the `detect-changes` action decides whether the PR "changes nothing the suite covers". A false positive in the first plus a false negative in the second is the path by which a real defect merges.

On the classifier specifically: matching on 402 status is much safer than matching on message text, because a model provider is free to put the word "budget" in an unrelated error. Check which one it does.
</details>

<details>
<summary><b>The author asked to be argued with</b></summary>

The description has a section headed "The part I want argued with", which is the right way to open this kind of PR. Give it a real answer rather than an approval; a reviewer agreeing out loud with the deliberate hole is what makes it a team decision instead of one person's call.

Worth adding to that conversation: a warn-and-pass that is invisible becomes permanent. If the budget stays exhausted for a week, this check passes all week and nobody notices the suite has not actually run since Tuesday. Ask whether the warn surfaces anywhere a human will see it, and whether there is a ceiling on how long it can keep passing.
</details>

<details>
<summary><b>The other half of the fix, out of scope here</b></summary>

Everything above treats the symptom. The cause is that a shared daily spend cap on one virtual key is a single point of failure for the whole repo's CI. Raising the cap, giving CI its own key, or caching example responses are all conversations this PR correctly does not try to have.
</details>

> [!NOTE]

drewdrewthis:
@/tmp/verdict_body.md

drewdrewthis (commented):
Two-pass review complete (line-by-line correctness + structural). No blocking findings — verdict: READY. Full verdict with verification detail posted as a separate comment: https://github.com/langwatch/scenario/pull/944#issuecomment-5423368717

drewdrewthis (approved):
this seems okay to me, however, it should post in the PR that this is what happened. if it's skipping tests, we should know

langwatch-agent:
Followed up on your condition here: #947.

The gap was real. The tolerated case turns a required check green having run nothing, and both places it reported that (the `::warning::` and the run summary) live in the Actions UI, behind a click, on a check that reads as passing. Nothing reached the pull request.

#947 posts the verdict on the PR itself, on three branches: the tolerated outage, the red one where the budget is out and the PR changes examples (because "the shared budget is out" and "your example is broken" send an author to different places), and the passing one, where it retracts a stale outage comment from an earlier attempt rather than leaving a false "did not run" on a run that did. A genuinely broken example gets no comment; the suite output already says it.

Reporting never decides. The gate reaches its verdict first, and every failure to post logs and exits 0, so a read-only token cannot turn a tolerated outage red. One comment per suite, edited in place, so a day-long budget window leaves one comment rather than a dozen.

Both jobs needed `pull-requests: write`, and the wiring validator now checks it: the reporter is fail-open by design, so without that scope the comment silently never appears, which is the same "looks wired, answers nothing" shape the rest of that validator exists to catch.

PR #957, opened 2026-08-31
feat(ci): comment what the coding agents spent on each pull request

rogeriochaves:
Brings the `pr-token-usage` workflow to this repository. It already runs on `langwatch/langwatch` (langwatch/langwatch#7676), where it posts and refreshes its own comment on itself.

## What it does

On every pull request open, push and reopen, it reads the LangWatch pull-request usage API for this repository and this pull request number, and keeps **one** comment in step with the answer:

| Contributor | Agent | Sessions | Total tokens | Estimated cost |
| --- | --- | --: | --: | --: |
| 🏭 Background Agents |  Claude Code | 1 | 268 million | $201.12 |
| **Total** |  | **1** | **268 million** | **$201.12** |

A details section carries the input / output / cache split per contributor and a per-model breakdown. Counts read as words, so a nine-digit number stays legible.

The comment is found again by a hidden marker and rewritten in place, so a pull request never collects a trail of stale tables.

## It never gates a merge

This job reports; it does not judge.

- A LangWatch outage or an error answer logs a `::warning` annotation and the job **stays green**. Do not add it to branch protection.
- A pull request with no recorded usage and no existing comment is left untouched, rather than stamped with an empty table. An existing comment is still refreshed, so numbers never go stale.
- Fork pull requests are skipped. They get a read-only token and no secrets, so both the read and the comment would fail; a quiet skip beats a red X on a first contribution.

## Shape

- `.github/scripts/pr-token-usage.ts` builds and upserts the comment. Dependency-free, run under `node --experimental-strip-types`, so there is no install step on the runner.
- `.github/scripts/pr-token-usage.test.ts` covers the builder. The workflow runs it **first**, in the same job, from the same checkout, so the logic that is about to write a comment proves itself on every run.
- `.github/scripts/package.json` marks those two files as ES modules. Without it node infers the module type from the repository root, which declares none, and reparses on every run with a `MODULE_TYPELESS_PACKAGE_JSON` warning. Nothing installs or publishes it.
- `specs/pr-token-usage.feature` holds the scenarios; six of them are bound by the test suite through `@scenario` annotations.

## Verified before pushing

- 13/13 tests pass locally.
- The API answers for this repository with the configured key: `langwatch/scenario#954` returns 200, `#956` returns `github_pr_not_mapped`, which the script reads as "no usage recorded" rather than an error.
- A dry run against both renders the "No coding agent sessions recorded" body, and a dry run against a pull request that does have sessions renders the table above.

## One thing to know

Every mapped pull request in this repository currently reports **zero sessions**. The repository is known to LangWatch; no coding agent session has been attributed to it yet. So this workflow will stay quiet here until sessions start landing, which is the correct behaviour and not a failure. Investigating that attribution gap is separate work.

## Setup

`LANGWATCH_PR_USAGE_API_KEY` is already set as a repository secret. No other configuration.

### DIFF
--- a/.github/scripts/package.json
+++ b/.github/scripts/package.json
@@ -1,0 +1,6 @@
+{
+  "name": "pr-token-usage-scripts",
+  "private": true,
+  "type": "module",
+  "description": "Marks the CI scripts beside it as ES modules. Without it node infers the module type from the repository root, which declares none, and reparses the script on every run with a MODULE_TYPELESS_PACKAGE_JSON warning. This is not a package: nothing installs, builds or publishes it, and it declares no dependencies because the scripts have none."
+}
--- a/.github/scripts/pr-token-usage.test.ts
+++ b/.github/scripts/pr-token-usage.test.ts
@@ -10,6 +10,8 @@ import {
   formatCount,
   humanizeCount,
   interpretUsageResponse,
+  nextPageUrl,
+  readPullRequestHead,
   type PullRequestUsage,
   type UsageRow,
 } from "./pr-token-usage.ts";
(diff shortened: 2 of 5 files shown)


rogeriochaves:
Both findings were real and both are fixed in `fd5aaf8`.

**Manual refresh named the wrong commit.** Correct: `github.event.pull_request` is absent on a dispatch, so the job-level `if` let every dispatched number through, and `github.sha` put the dispatch branch in the footer instead of the pull request's head. The pull request answers both questions, so the script asks it: the workflow passes a head sha only when the event carries one, and an empty value sends the script to the API for `head.sha` and `head.repo.full_name`. A head branch in another repository is refused, and so is a head repository that is gone because the fork was deleted. Verified against the live API: with `PR_HEAD_SHA` unset, the script resolves the pull request's real head sha where the old code printed the default branch.

**Comment pagination.** Correct: the page-10 cap could hide the marker and turn an update into a duplicate POST. The walk now follows the `Link` header's `rel="next"` and stops when there is no next, rather than at a fixed page count.

Tests went from 13 to 16. The three new ones cover the resolved head sha, both fork readings, and `rel="next"` parsing including its absence.

On the merge-risk note about credentials: the job takes no untrusted input. Fork pull requests are refused on both paths now, and same-repository branches already carry write access, so the secret is not reachable by anyone who could not already use it.

PR #958, opened 2026-08-31
fix(cursor): quote the globs so the rule frontmatter parses

dennishavermans:
## Why

Five `.cursor/rules/*.mdc` files have frontmatter that is not valid YAML, so none of it is read — not just the glob, but `description` and `alwaysApply` with it. A bare `*` opens an alias in YAML, so `globs: *.tsx` is a reference to an anchor named `.tsx` that nothing defines, and the parse fails at that line. It is invisible because a rule that fails to load fails silently.

No linked issue — I found this from outside while running a static analyser over public repos' agent configuration. Happy to open one if your change-management process wants the ticket for the audit trail.

## What changed

- Quoted the value in the five affected files, rather than rewriting the globs. Quoting is the minimum change that makes the frontmatter parse, and it leaves the matching behaviour exactly as the author wrote it.
- Left the glob strings themselves untouched (`*.tsx` stays `*.tsx`). Widening them to `**/*.tsx` would make the rules match more files, and that is a scope decision for you, not a syntax fix — see *Anything surprising?*.
- Changed nothing else: five files, one line each, `+1/-1`, and every other frontmatter key parses to the same value it was meant to have.

## Test plan

From a checkout of this branch, comparing `main` against `HEAD` with stdlib PyYAML — no dependencies beyond `git`, `awk` and `python3`:

```bash
RULES=(.cursor/rules/react-tsx.mdc .cursor/rules/typescript.mdc \
       docs/.cursor/rules/seo.mdc docs/.cursor/rules/writing-style.mdc \
       python/.cursor/rules/python.mdc)

check() {
  for f in "${RULES[@]}"; do
    printf '  %-38s ' "$f"
    git show "$1:$f" | awk '/^---$/{n++; next} n==1' | python3 -c '
import sys, yaml
try:
    yaml.safe_load(sys.stdin.read()); print("parses")
except Exception as e:
    print("FAILS:", str(e).splitlines()[0])
'
  done
}

echo "on main:"; check main
echo "on this branch:"; check HEAD
```

No test file is added: these are configuration files, and the repo has no harness that parses them.

Output of the command above, run on this branch:

```
on main:
  .cursor/rules/react-tsx.mdc            FAILS: while scanning an alias
  .cursor/rules/typescript.mdc           FAILS: while scanning an alias
  docs/.cursor/rules/seo.mdc             FAILS: while scanning an alias
  docs/.cursor/rules/writing-style.mdc   FAILS: while scanning an alias
  python/.cursor/rules/python.mdc        FAILS: while scanning an alias

on this branch:
  .cursor/rules/react-tsx.mdc            parses
  .cursor/rules/typescript.mdc           parses
  docs/.cursor/rules/seo.mdc             parses
  docs/.cursor/rules/writing-style.mdc   parses
  python/.cursor/rules/python.mdc        parses
```

I checked the same five files against four independent YAML parsers, so this does not rest on one implementation's strictness: **`yaml` (JS), `js-yaml`, PyYAML 6.0.3, and Ruby Psych 3.1.0 all reject every file before and accept every file after.** The parsed `globs` value is byte-identical to the string that was already there in all five.

## Anything surprising?

Two things worth your judgement, neither of which this PR acts on.

**The globs may not match what they look like they match.** Once the frontmatter parses, `*.tsx` is a path pattern anchored at the rule's directory, and every `.tsx` file in the repo is nested rather than sitting at that level. Same for the other two:

| rule | glob | files it matches | files of that type |
|---|---|---|---|
| `.cursor/rules/react-tsx.mdc` | `*.tsx` | 0 | 75 |
| `.cursor/rules/typescript.mdc` | `*.tsx` | 0 | 75 |
| `docs/.cursor/rules/seo.mdc` | `*.mdx` | 0 | 88 |
| `docs/.cursor/rules/writing-style.mdc` | `*.mdx` | 0 | 88 |
| `python/.cursor/rules/python.mdc` | `*.py` | 2 | 552 |

The obvious change is `**/*.tsx`, which Cursor's docs describe as "all `.ts` files in any directory". I deliberately did not make it, because it is not free: `**/*.tsx` from the repo root also picks up the 53 vendored `.tsx` files under `python/examples/lovable_clone/template/`, 48 of them shadcn/ui components, and `**/*.py` under `python/` picks up the 207 generated files in `python/scenario/_generated/`. Whether you want those rules attaching to vendored and generated code is a call for whoever owns the rules.

**`.cursor/rules/typescript.mdc` has an empty `description:` and scopes to `*.tsx` only.** For a rule named "typescript" I would have guessed `.ts` was meant to be in there too, but I have left it alone since that is intent rather than syntax.

### DIFF
--- a/.cursor/rules/react-tsx.mdc
+++ b/.cursor/rules/react-tsx.mdc
@@ -1,6 +1,6 @@
 ---
 description: React and TSX component development guidelines
-globs: *.tsx
+globs: "*.tsx"
 alwaysApply: false
 ---
 
--- a/.cursor/rules/typescript.mdc
+++ b/.cursor/rules/typescript.mdc
@@ -1,6 +1,6 @@
 ---
 description:
-globs: *.tsx
+globs: "*.tsx"
 alwaysApply: false
 ---
 
--- a/docs/.cursor/rules/seo.mdc
+++ b/docs/.cursor/rules/seo.mdc
@@ -1,5 +1,5 @@
 ---
-globs: *.mdx
+globs: "*.mdx"
 alwaysApply: false
 ---
 
--- a/docs/.cursor/rules/writing-style.mdc
+++ b/docs/.cursor/rules/writing-style.mdc
@@ -1,5 +1,5 @@
 ---
-globs: *.mdx
+globs: "*.mdx"
 alwaysApply: false
 ---
 # Writing Style
--- a/python/.cursor/rules/python.mdc
+++ b/python/.cursor/rules/python.mdc
@@ -1,5 +1,5 @@
 ---
-globs: *.py
+globs: "*.py"
 alwaysApply: false
 ---
 


langwatch-agent (commented):
Static review found no blocking correctness or integrity issue. Quoting the YAML scalar preserves each glob value while making the frontmatter parseable. No PR code was executed.

LangWatch-Review: verdict=clean sha=3303cedc0a4cb7e85293193ecad0eb58f6f4ada1 p0=0 p1=0 p2=0 p3=0

langwatch-agent:
## Human Review Brief

> [!NOTE]
> `Fast Skim`. Five files, one line each, `+1/-1`. Quoting five YAML scalars. The change is correct and the reasoning in the description is right; the only judgement to make is the scope question the author raises themselves and deliberately did not act on.

| | |
| --- | --- |
| **Issue** | None. Outside contributor, found by static analysis over public repos |
| **Size** | +5/-5 across 5 files |
| **Mode** | Fast Skim |
| **Blast radius** | Cursor rule loading, editor only. No runtime surface |
| **Author** | dennishavermans (outside contributor) |
| **Reviewer** | `rogeriochaves` |

<details>
<summary><b>What you are looking at, if <code>.cursor/rules</code> is not something you use</b></summary>

`.mdc` files are Cursor's rule format: YAML frontmatter followed by markdown. The frontmatter carries `globs` (which files the rule applies to), `description`, and `alwaysApply`.

The bug is a YAML one and it is worth knowing because it will recur. **A bare `*` opens an alias in YAML.** So `globs: *.tsx` does not mean "files ending in .tsx", it is a reference to an anchor named `.tsx` that nothing defines, and the document fails to parse at that line.

The consequence is larger than the glob. When the frontmatter fails to parse, **nothing in it is read** — not `globs`, not `description`, not `alwaysApply`. The rule silently does not load. There is no error surface anywhere: an editor that cannot parse a rule just does not apply it, and no one notices a rule that was never enforced.

Five files are affected, across three trees: the repo root, `docs/`, and `python/`.

</details>

<details>
<summary><b>The one decision to make</b></summary>

The author quoted the values and deliberately left the glob strings alone: `*.tsx` stays `*.tsx` rather than becoming `**/*.tsx`. That is the correct instinct for a syntax fix, and they say so explicitly.

But it does mean the rules will now load and match **only files at the same directory level as the rule**, which is almost certainly not what the original authors intended and is not what anyone has been getting either way. So the practical question is not whether to merge this, it is whether to follow it with the widening. Worth answering in the same breath, because a rule that loads and matches nothing looks fixed and is not.

Nothing else to check. No runtime code, no dependency, no CI surface.
</details>

> [!NOTE]

dennishavermans:
Closing this. I got it wrong, and merging it would have made things worse rather than better.

The premise was that `globs: *.tsx` is invalid YAML, which it is: a bare `*` opens an alias, and four independent parsers reject the frontmatter. What I did not establish is whether Cursor reads these files as YAML at all. It does not. Cursor takes the raw text after `globs:` as a comma-separated pattern list, which is why its own UI writes the value unquoted and why its documentation shows it unquoted.

So quoting does not fix a broken rule here. It breaks a working one. After this change the pattern would be the six-character string `"*.tsx"` including the quote marks, which matches no file, and the rule would silently stop attaching. That is precisely the failure mode I opened the PR to prevent.

Credit where it is due: a reviewer on an unrelated PR of mine at PostHog caught this and tested it properly, which is more than I did. https://github.com/PostHog/posthog/pull/91631#issuecomment-5484615411

What is left standing is much smaller. The frontmatter genuinely is invalid YAML, so any tool other than Cursor that reads these files gets nothing from them, including `description` and `alwaysApply`. But there is no fix for that which keeps Cursor working, since any YAML-valid spelling of a leading-`*` glob is a spelling Cursor will not match on. That is a Cursor issue, not a langwatch one, and not something to paper over in your repo.

Sorry for the review time. The one thing from the PR still worth a look, entirely independently of the YAML question, is that `.cursor/rules/typescript.mdc` has an empty `description:` and scopes to `*.tsx` only, which for a rule named "typescript" may not be what was intended.

PR #964, opened 2026-09-01
fix(execution): clear the sampled barge-in delay on every skip path (#578)

### LINKED ISSUE #578: refactor(voice): break `maybeScheduleInterruptedAgentTurn` into named substeps
## Context
ScenarioExecution.maybeScheduleInterruptedAgentTurn at javascript/src/execution/scenario-execution.ts:1422 is ~115 lines. It does, sequentially:

1. Resolve interruption config + bail conditions (next-role check, voice-adapter check, voice-user-sim check, in-flight-task check, RNG gate).
2. Locate the AGENT to interrupt + reorder pendingAgentsOnTurn / pendingRolesOnTurn.
3. Build the non-blocking task entry that runs the AGENT (duplicates structure from agentNonBlocking).
4. Sleep for the configured delayRange then schedule + record the barge-in.

Each step is independently understandable; together they are a wall of code that is hard to review.

**Effort**: S/M — follow-up from PR #561. Not blocking.

## Suggested fix shape
Extract to private methods:
-  — the bail/lookup phase.
-  — the queue surgery.
-  (consider folding into makeTaskEntry from the shared-helpers follow-up issue).
-  (or returns a Promise — match existing style).

maybeScheduleInterruptedAgentTurn becomes ~20 lines of orchestration. Each substep can grow its own unit test.

## Why it matters
Long private methods on long classes hide the seams; an outside reader cannot tell where the config gate ends and the side-effect begins. Extraction also unblocks targeted unit tests (e.g. for the rewirePendingForInlineBarge invariant — currently exercised only end-to-end).

## Acceptance
- maybeScheduleInterruptedAgentTurn is <= 30 lines.
- Each extracted substep has at least one targeted unit test exercising its branches.
- All existing tests still pass.

langwatch-agent:
## Ticket

Closes #578.

## Premise, re-checked on current main: most of it already landed

Worth stating up front, because it makes this PR much smaller than the ticket reads.

| acceptance criterion | state on `main` |
|---|---|
| `maybeScheduleInterruptedAgentTurn` is <= 30 lines | **done.** It is 17 lines, and reads as orchestration: five named substeps and the bail conditions |
| each extracted substep has at least one targeted unit test | **three of four done.** `resolveNextAgentForInlineBarge`, `consumePendingRolesUntilAgent` and `dispatchAgentBackground` each have a `describe(... substep N)` block in `proceed-interruptions.test.ts` |
| all existing tests still pass | unchanged here |

So what is left is the fourth substep. `prepareAndFireBargeIn` was extracted with the others and is the only one that never got its block.

## Why that gap is the interesting one

It is the substep that decides whether a barge-in happens at all, and it is 43 lines with four exits. Three of the four return `true` **without** firing one:

- the sampled delay is recorded, TTS succeeds, the bot is still talking, the barge-in fires;
- TTS throws, so the barge-in is skipped;
- the bot finished while TTS was running, so there is nothing to interrupt;
- and the happy path.

`true` therefore does not mean "the user interrupted", it means "the AGENT was dispatched". A test that asserts the return value cannot tell any of these apart. So these assert the side effects: whether the interrupt fired, whether the voiced turn reached the conversation, and what the sampled delay was left as.

The stale-delay case is the one worth having. The delay is sampled *before* the TTS attempt, so the failure path has to clear it; if it does not, a later successful barge-in silently inherits the failed attempt's value. The code comments call that a P2 fix and nothing pinned it.

## Change

Tests in one file, following the `substep N` convention already in it, **plus a production fix that review turned up**.

### The bug review found

This PR started test-only. Writing the coverage exposed a real defect, so it now carries the fix too.

The barge-in delay is sampled *before* the TTS attempt, so any exit that fires no barge-in owns clearing it: nothing downstream consumes it, because `fireUserInterrupt` is what consumes it and it never runs on a skip path. Only the TTS failure cleared it. Two things were wrong:

1. the bot-finished bail (`entry.done`) returned with the value still set;
2. the write itself was conditional on a positive sample, so a later turn sampling zero kept the earlier turn's value.

Either alone is a leak, and together they mean a barge-in can run on a delay sampled by a turn that never interrupted. It is reachable: `delayRange` is caller-settable (`init.delayRange ?? [0.5, 3.0]`), so a range starting at zero produces a zero sample. The stock range is always positive, which always overwrites, and that is what hid it.

Both bail paths now clear the field, and the write is unconditional.

I also deleted one of my own tests rather than keep it. It asserted the field was unset after a zero sample on a freshly built executor, where the field starts unset anyway, so it could not distinguish "never written" from "cleared" and passed whatever production did. The stale-value test supersedes it strictly.

## Evidence

- `vitest run src/execution/__tests__/proceed-interruptions.test.ts`: **16 passed** (10 before).
- `vitest run src/execution src/voice src/domain`: **631 passed, 1 skipped** across 70 files.
- `eslint` clean on the changed file, and `lint:lib` clean.
- Typecheck: **0 errors naming this file**. Total diagnostics are identical with my change and with `main`'s version of the file (8 either way, all pre-existing and in files this PR does not touch), and the run was confirmed live by injecting a deliberate type error and seeing it reported.

### Mutations, all 8 caught

Re-derived against the current code rather than carried over, because the fix changed what the anchors are:

| mutation | result |
|---|---|
| the TTS bail stops clearing the sampled delay | 1 failed |
| TTS failure fires the barge-in anyway | 1 failed |
| the bot-finished check is removed entirely | 2 failed |
| the barge-in fires but the turn is never recorded on the conversation | 1 failed |
| a different phrase is voiced than the one the config picked | 1 failed |
| the bot-finished bail stops clearing the sampled delay | 1 failed |
| the delay write becomes conditional on a positive sample again | 1 failed |
| the bot-finished bail is moved to before the TTS call | 1 failed |

Two notes on how that table was produced, since both are ways a mutation table lies.

**A mutation has to land where you meant.** `this.state.addMessage(voicedMessage)` appears twice in `scenario-execution.ts` and only the second is in this substep, so a first-occurrence replacement patches the wrong call site, the suite stays green, and it reads as an escaped mutation. That row is line-targeted.

**Moving a guard means deleting it from the old place too.** The last row initially came back green because the edit added the check above the TTS call while leaving the original below it, which preserves behaviour exactly. A guard that is only added is not moved.

## Advisory note

### DIFF
--- a/javascript/src/execution/__tests__/proceed-interruptions.test.ts
+++ b/javascript/src/execution/__tests__/proceed-interruptions.test.ts
@@ -267,8 +267,17 @@ type BargeInInternals = {
  * A live entry is the precondition the substep runs under: its sole caller
  * dispatches the AGENT first, so `done` is false unless the bot beat the TTS.
  */
-function liveEntry(done = false) {
-  return { promise: Promise.resolve(), done, error: null };
+function liveEntry() {
+  return { promise: Promise.resolve(), done: false, error: null };
+}
+
+/** A promise whose settling the test controls, to pin an ordering. */
+function deferred<T>() {
+  let resolve!: (value: T) => void;
+  const promise = new Promise<T>((r) => {
+    resolve = r;
+  });
+  return { promise, resolve };
 }
 
 /** A voice user simulator whose TTS outcome the test decides. */
@@ -362,21 +371,36 @@ describe("prepareAndFireBargeIn (substep 4)", () => {
     expect(inner.interruptBargeInDelayMs).toBeUndefined();
   });
 
-  it("skips the barge-in when the bot finished before the phrase was voiced", async () => {
+  it("skips the barge-in when the bot finishes while the phrase is being voiced", async () => {
     const { exec, inner, fired } = makeBargeInExec();
     exec.interruptOverrides = { rng: () => 0 };
     const config = new InterruptionConfig({
       strategy: "random_phrase",
       delayRange: [2, 2],
     });
+    // Live when the substep starts, which is what its sole caller guarantees.
+    const entry = liveEntry();
 
-    const result = await inner.prepareAndFireBargeIn(
-      config,
-      simThat({ voiced }),
-      liveEntry(true),
-    );
+    const ttsStarted = deferred<void>();
+    const ttsFinishes = deferred<void>();
+    const sim = {
+      voice: "alloy",
+      async voiceifyText(): Promise<ModelMessage> {
+        ttsStarted.resolve();
+        await ttsFinishes.promise;
+        return voiced;
+      },
+    };
 
-    expect(result).toBe(true);
+    const pending = inner.prepareAndFireBargeIn(config, sim, entry);
+    await ttsStarted.promise;
+    // The bot finishing DURING the TTS call is the situation the check exists
+    // for, and the ordering is the whole of it: a check that ran before the
+    // phrase was voiced would have seen a live entry and barged in anyway.
+    entry.done = true;
+    ttsFinishes.resolve();
+
+    await expect(pending).resolves.toBe(true);
     expect(fired).toEqual([]);
   });
 
--- a/javascript/src/execution/__tests__/proceed-interruptions.test.ts
+++ b/javascript/src/execution/__tests__/proceed-interruptions.test.ts
@@ -9,6 +9,7 @@
  * determinism; no network, no real keys.
  */
 
+import type { ModelMessage } from "ai";
 import { describe, it, expect, vi } from "vitest";
 
 import {
@@ -18,8 +19,8 @@ import {
   type AgentReturnTypes,
   UserSimulatorAgentAdapter,
 } from "../../domain";
-import { ScenarioExecution } from "../scenario-execution";
 import { InterruptionConfig } from "../../voice/interruption";
+import { ScenarioExecution } from "../scenario-execution";
 
 class MockAgent extends AgentAdapter {
   role = AgentRole.AGENT;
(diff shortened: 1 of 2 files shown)


javascript/src/execution/__tests__/proceed-interruptions.test.ts
```diff
@@ -251,6 +252,169 @@ describe("consumePendingRolesUntilAgent (substep 2)", () => {
...
+    exec.interruptOverrides = { rng: () => 0 };
+    const config = new InterruptionConfig({
+      strategy: "random_phrase",
+      delayRange: [2, 2],
+    });
+
+    const result = await inner.prepareAndFireBargeIn(
+      config,
+      simThat({ voiced }),
+      liveEntry(true),
```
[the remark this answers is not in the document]
langwatch-agent: Valid, and fixed in `1c9c82be`. I checked it with a mutation rather than taking it on faith, and the check agreed with you.

**Why the original test was weak.** Setting `done` before the call does reach the skip branch, so the branch was covered, but it pins nothing about *when* the check runs. Hoisting the check above the `voiceifyText` call keeps the old test green while destroying the protection it stands for, because the real situation is the bot finishing while TTS is in flight, and at the moment a hoisted check runs the entry is still live.

Measured, moving the check from after the TTS call to before it:

| test version | moved-check mutation |
|---|---|
| original (`done` set upfront) | **16 passed, escapes** |
| this fix (`done` set during TTS) | **1 failed** |

**What changed.** The TTS call now blocks on a deferred. The test waits for it to start, marks the entry done while it is pending, then releases it, so the assertion rests on the ordering:

```ts
const pending = inner.prepareAndFireBargeIn(config, sim, entry);
await ttsStarted.promise;
entry.done = true;
ttsFinishes.resolve();

await expect(pending).resolves.toBe(true);
expect(fired).toEqual([]);
```

I replaced the upfront-`done` case rather than keeping both. Its input is not reachable in production anyway: the substep's sole caller dispatches the AGENT immediately before, so the entry is always live on entry, which is what the docblock on `liveEntry` now says. Keeping a weaker duplicate would have added a test that detects nothing the new one misses.

Also dropped the now-unused `done` parameter on the `liveEntry` helper.

**Re-verified:** 16 passed in the file, `618 passed / 1 skipped` across 68 files in `src/execution` and `src/voice`, eslint clean, 0 typecheck errors naming the file. The other five mutations in the PR body still fail as before.

javascript/src/execution/__tests__/proceed-interruptions.test.ts
```diff
@@ -251,6 +252,193 @@ describe("consumePendingRolesUntilAgent (substep 2)", () => {
...
+    const pending = inner.prepareAndFireBargeIn(config, sim, entry);
+    await ttsStarted.promise;
+    // The bot finishing DURING the TTS call is the situation the check exists
+    // for, and the ordering is the whole of it: a check that ran before the
+    // phrase was voiced would have seen a live entry and barged in anyway.
+    entry.done = true;
+    ttsFinishes.resolve();
+
+    await expect(pending).resolves.toBe(true);
+    expect(fired).toEqual([]);
```
langwatch-agent, line 389: **P2 · Assert the delay is cleared on this other skip path**

- **What you are looking at.** This test reaches the `entry.done` return after `prepareAndFireBargeIn` has sampled and stored a nonzero delay, then only asserts that no interrupt fired.

- **Why it is scary.** The production `entry.done` branch returns without clearing `interruptBargeInDelayMs`; unlike the TTS catch branch, no `fireUserInterrupt` runs to consume it. A later successful barge-in whose sampled delay is zero will therefore inherit this prior delay. Add an assertion that the field is unset here, then clear it in the corresponding production branch.

langwatch-agent (commented):
One P2 found. See the inline comment.

LangWatch-Review: verdict=findings sha=1c9c82bed56438f5deb9aed00b6b5ab9441de5cc p0=0 p1=0 p2=1 p3=0

langwatch-agent:
## Human Review Brief

> [!NOTE]
> `Fast Skim`. One new test file, no production code. The author re-checked the ticket against `main` and found three of its four acceptance criteria already satisfied, so this is the remaining quarter rather than the whole thing. That re-check is the most useful part of the description.

| | |
| --- | --- |
| **Issue** | `Closes #578`. Most of it already landed; this closes the gap |
| **Size** | One file, `javascript/src/execution/__tests__/proceed-interruptions.test.ts` |
| **Mode** | Fast Skim |
| **Blast radius** | Test coverage only |
| **Reviewer** | `rogeriochaves` |

<details>
<summary><b>What you are looking at, if the Scenario executor is not your area</b></summary>

A scenario is a scripted conversation: the executor advances it turn by turn, deciding which agent speaks next and dispatching it.

**Barge-in** is the case where the simulated user interrupts mid-turn. It is the hardest part of the executor because it is the one place the normal turn order does not hold: an agent turn is already in flight, a user turn arrives on top of it, and the executor has to decide whether the interrupted agent gets rescheduled, which role speaks next, and what happens to work already dispatched.

`maybeScheduleInterruptedAgentTurn` is the function that answers that. It was refactored into named substeps to get it under a length bound, and the acceptance criteria for that refactor required each extracted substep to carry its own targeted test. Three got one. This is the fourth.

The point of testing an extracted substep directly, rather than relying on the orchestrator's tests, is that extraction moves behaviour without changing it, so the orchestrator's existing tests keep passing whether or not the substep is correct in isolation. A substep with no test of its own is behaviour that only the code path currently reaching it can prove, which stops being true the moment someone reuses it.

</details>

<details>
<summary><b>The one useful check</b></summary>

Whether the new test would fail if the substep were broken. A test written after the fact, against code that already works, is easy to write in a way that passes for the wrong reason. The cheapest verification is to break the substep locally and confirm the test goes red, which is the same experiment the author ran on the neighbouring #7719 and reported.

Nothing else. No production code, no dependency, no CI surface.
</details>

> [!NOTE]

javascript/src/execution/__tests__/proceed-interruptions.test.ts
drewdrewthis, line 389: **New AC** — the stale-delay guard this PR's own description calls out ("the failure path has to clear it; if it does not, a later successful barge-in silently inherits the failed attempt's value") is only pinned for the TTS-throw path (test at line ~367). It is not pinned here, and the asymmetry looks real, not incidental:

- In `prepareAndFireBargeIn`, the delay field is only ever explicitly cleared in the `catch` block (TTS failure). The `entry.done` bail (this test, `scenario-execution.ts` around the `if (entry.done) { return true; }` branch) returns early **without** clearing `interruptBargeInDelayMs` — yet it is exactly the "skip the barge-in" case the PR's rationale says needs clearing.
- Separately, the top-of-function assignment is `if (delaySeconds > 0) { this.interruptBargeInDelayMs = ... }` — when a fresh sample is `0`/negative, a **previously-set** value is left untouched (not reset to `undefined`). Test 2 ("leaves the barge-in field unset when the sampled delay is zero", ~line 291) only proves this on a freshly-constructed exec where the field starts `undefined` — it can't distinguish "field was never set" from "a stale prior value got cleared", so it passes trivially and doesn't actually guard the regression it's named for.

Concretely: neither test exercises the sequence *(call 1 sets a nonzero delay via entry.done-bail or a real fire) -> (call 2 samples 0)* and asserts the field ends up `undefined`. As written, if that clearing is in fact missing in production, these tests would still pass.

Ask: either (a) add a test that seeds a stale non-zero `interruptBargeInDelayMs` before invoking `prepareAndFireBargeIn` again and asserts it's cleared on both bail paths, confirming current behavior is correct, or (b) if it does not clear (worth checking against `scenario-execution.ts`'s `entry.done` branch), file the production fix and land the regression test in the same change. This is precisely the class of bug the PR text calls a P2 worth pinning — it's just pinned on only one of the two bail paths.

drewdrewthis (commented):
Test-quality focused review of the new `prepareAndFireBargeIn (substep 4)` block. One finding below is blocking (New AC); the rest of the suite is solid — side-effect assertions instead of return-value checks, correct unit-level placement, deterministic RNG/deferred-promise seams.

drewdrewthis (dismissed):
Formal review-state event for the NOT-READY verdict below — see the verdict comment and inline thread https://github.com/langwatch/scenario/pull/964#discussion_r3904558414 for the finding.

drewdrewthis:
## Review verdict: READY

**Reviewed at:** `0f3a0723` · **Run:** pr-review (reviewer: @drewdrewthis)

### Drew's prior CHANGES_REQUESTED review — addressed

Drew's inline finding (https://github.com/langwatch/scenario/pull/964#discussion_r3904558414, at commit `1c9c82b`) flagged an asymmetry: the stale-`interruptBargeInDelayMs` guard was only pinned for the TTS-throw bail, not for the `entry.done` (bot-finished) bail, and the zero-sample test couldn't distinguish "never set" from "cleared."

Both asks are now satisfied in `javascript/src/execution/scenario-execution.ts` and `javascript/src/execution/__tests__/proceed-interruptions.test.ts`:

- **Production fix:** the `entry.done` bail now explicitly clears `this.interruptBargeInDelayMs = undefined` before returning, matching the TTS-failure bail. The top-of-function write is now unconditional (`delaySeconds > 0 ? ... : undefined`) instead of `if (delaySeconds > 0) { ... }`, so a zero/negative sample overwrites rather than preserving a stale prior value.
- **Regression tests:** `describe("when a prior call left a sampled delay behind", ...)` adds exactly the two cases Drew asked for — "clears the delay on the bot-finished bail, as the TTS bail does" (seeds a real stale value via the production entry-done path, then asserts a second call clears it) and "overwrites a stale delay when the next sample is zero" (pre-sets a nonzero value, samples 0, asserts `undefined`). Both would fail against the pre-fix behavior.

This closes the gap Drew identified — verified by reading the current diff against the specific claims in the prior review comment, not by re-running the test suite myself.

### Other review notes (non-blocking)
- CodeRabbit's earlier suggestion on this PR (ordering of `entry.done` vs. TTS completion in the "bot finishes while phrase is voiced" test) is already implemented correctly in the current test — the test drives `entry.done = true` only after `voiceifyText` has started, via a deferred promise.
- CI: `ci-checks (24.x)`, `javascript-complete`, `python-complete`, `docs-complete`, `preflight`, `changes`, `Validate PR Title`, CodeQL analyze, and CodeRabbit are all SUCCESS at this SHA. The one FAILURE row on this SHA (`evaluate`, run 34195741109) belongs to the "Auto-approve for validated Low-Risk or Firefighting Check" workflow, not the test/build pipeline — this PR carries a production fix so it doesn't qualify as low-risk auto-approve, which is the expected/correct outcome, not a CI regression.

---
_Verdict is prose, not a GitHub approval. Scope: review findings only — READY means no unresolved blocking review threads at this SHA. It is **not** a merge-readiness signal._

javascript/src/execution/__tests__/proceed-interruptions.test.ts
langwatch-agent: You were right on both counts, and it was option (b): the clearing is genuinely missing, so this is now a production fix plus the regression tests, in `0f3a0723`.

**What I checked first.** Rather than argue either way, I wrote the two tests you described and ran them against unmodified production code. Both failed:

```
× clears the delay on the bot-finished bail, as the TTS bail does
× overwrites a stale delay when the next sample is zero
Tests  2 failed | 16 passed (18)
```

So the asymmetry is real, and your read of why my existing test could not see it is exactly right: it built a fresh executor whose field already starts `undefined`, so "never written" and "cleared" are indistinguishable and it passed whatever production did.

**Reachability, since a defect that cannot happen deserves a different answer.** It can happen. `delayRange` is caller-settable (`init.delayRange ?? [0.5, 3.0]`), so a range starting at zero yields a zero sample. With the stock range the fresh sample is always positive and always overwrites, which is what hid this. The sequence is: a turn bails at `entry.done` leaving the value set, nothing consumes it because `fireUserInterrupt` never runs, then a later turn samples zero, the conditional write is skipped, and the barge-in runs on a delay from a turn that never interrupted. `maybeScheduleInterruptedAgentTurn` runs every proceed step, so there is no shortage of later turns.

**The fix.** Both halves, since each is independently wrong:

- the bot-finished bail now clears the field, for the same reason the TTS bail does;
- the write is unconditional, `delaySeconds > 0 ? ... : undefined`, so a turn that sampled nothing says so instead of inheriting.

**I also deleted the weak test rather than leaving it.** Your point that it "passes trivially and doesn't actually guard the regression it's named for" is the whole objection to it, and the new stale-value test subsumes it strictly. A test that cannot fail for the reason its name gives is worse than no test, so it is gone rather than renamed.

**Mutations, each half reverted in turn:**

| reverted | result |
|---|---|
| bot-finished bail stops clearing | 1 failed |
| write becomes conditional again | 1 failed |
| TTS bail stops clearing (the guard that already existed) | 1 failed |

**Verification:** 17 passed in the file, `632 passed / 1 skipped` across 70 files in `src/execution`, `src/voice`, `src/domain`, eslint clean, and 0 typecheck errors naming either changed file with the total unchanged from baseline. The typecheck was confirmed live with an injected control error rather than trusted.

Thanks for pushing on this one. The PR was originally test-only, and the finding turned it into a real bug fix.

langwatch-agent: Confirmed and fixed in `0f3a0723`. You called it correctly: the `entry.done` branch returned without clearing `interruptBargeInDelayMs`, and unlike the TTS catch there is no `fireUserInterrupt` on that path to consume it.

I wrote the assertion you asked for before touching production, and it failed against the unmodified code, so the behaviour was wrong rather than merely unpinned.

One thing worth adding to your description of it: clearing the bail alone would not have closed the hole. The write was also conditional (`if (delaySeconds > 0)`), so a later turn sampling zero kept the earlier value regardless. Both are fixed, and each is pinned by its own mutation.

Detail, evidence and the reachability analysis are in the sibling thread rather than duplicated here.

drewdrewthis (commented):
READY — Drew's prior CHANGES_REQUESTED (stale barge-in delay guard) is addressed by the new commit. See verdict: https://github.com/langwatch/scenario/pull/964#issuecomment-5494675434

drewdrewthis (approved):
lgtm

PR #973, opened 2026-09-06
feat(python): resolve voice providers per run

### LINKED ISSUE #896: feat(voice): Python voice provider state is a module global, not per-run (ADR-002/003 follow-up)
## The gap

Python has no per-run voice configuration. TypeScript resolves the STT provider from the per-run carrier. Python falls through to a process-wide module global.

Same job, two mechanisms:

`javascript/src/agents/judge/judge-agent.ts:482`

```ts
const resolved = resolveVoiceConfig(undefined, input.scenarioConfig.voice);
```

`python/scenario/judge_agent.py:529`

```python
await transcribe_segments(recording)
```

No provider argument. `transcribe_segments` falls back to `get_stt_provider()`, which reads the module global `_provider` in `python/scenario/voice/stt.py`.

Concretely missing on the Python side:

- no `voice` field on `ScenarioConfig` (`python/scenario/config/scenario.py` has zero references to it)
- no `voice=` parameter on `run()` (`python/scenario/scenario_executor.py:2008`)
- no `resolve_voice_config` equivalent to `javascript/src/voice/config.ts`
- no `resolve_stt_provider` router equivalent to `javascript/src/voice/stt/stt-provider.ts`

## Why it matters

Two `scenario.run()` calls in one Python process with different STT providers clobber each other. The last `set_stt_provider()` wins for both. ADR-002 names this directly:

> process-wide mutable provider state is not safe under concurrent runs

That is the failure ADR-001 was written to eliminate, and the platform runs batches of scenarios in parallel.

## Why it is only being filed now

ADR-003 (`docs/adr/003-voice-internal-design.md`, Accepted) records the obligation under "TS/Python divergence, a Python parity follow-up is OWED":

> **Obligation:** file a follow-up issue to bring the Python SDK to this corrected design, parity on the right design, not on the bug. Do not close #372's epic without it.

ADR-002 (`docs/adr/002-voice-provider-state.md`) says the same in its consequences.

That issue was never filed. #740 carried the work in its gap table as "Per-run STT provider isolation, Python `stt.py` is a module-level singleton global, concurrent runs race", and closed on 2026-07-10 with the slice unlanded. There is still no `resolve_stt_provider` anywhere in the Python tree.

#743 sat on top of this and is now closed by #841, which corrected the docs to name `scenario.set_stt_provider()` as the Python entry point rather than adding the `configure(stt=)` argument both ADRs removed. #841 deliberately left the global in place, because replacing it is this issue.

## Target design

Per ADR-002 the carrier is the per-run config object that reaches `call()`, because the consumers (the judge transcription pass and the user simulator TTS pass) run inside `call()`:

```
cfg.voice?.stt  ??  <per-run OpenAI default>
```

`javascript/src/voice/config.ts` is the reference implementation. Build the Python equivalent, do not invent a second shape.

## Scope

- add a voice config to `ScenarioConfig` and a `voice=` seam at the `run()` boundary
- resolve the provider per run, constructing a fresh OpenAI default when unset
- thread the resolved config into the judge and user simulator paths
- decide the fate of `scenario.set_stt_provider()`: keep it working as a convenience, or deprecate it on a stated timeline

This touches the judge, the user simulator and the executor, so it is not a small change. It is also a public API addition, so the shape should match TypeScript rather than diverge again.

## Definition of done

- two concurrent `scenario.run()` calls with different STT providers each use their own, proven by a test
- the judge reads the provider from the per-run config, not from a module global
- `python/scenario/voice/stt.py` no longer needs `_provider` for the judge path

## Refs

ADR-002, ADR-003, #372, #740, #743, #841

atirna:
## Description

I added per-run Python voice configuration. The judge now resolves STT from `voice=VoiceConfig(stt=...)` carried by `ScenarioConfig`, and the user simulator reads the same run config for TTS routing.

`set_stt_provider()` remains available for direct transcription helpers, but is deprecated and no longer controls the judge path.

Fixes #896

## Verification

- Before, current `main` has no `voice` argument on `scenario.arun()`, and the judge falls through to the module-global STT provider.
- After: `python -m pytest tests/test_arun_voice_config.py tests/test_user_simulator_agent.py tests/voice/test_stt.py tests/voice/test_transcribe.py tests/test_judge_agent.py -q`
- After: `python -m pytest tests -k 'stt and not stt_swap' -q`


python/scenario/judge_agent.py
```diff
@@ -764,7 +765,12 @@ async def _build_conversation_view(self, input: AgentInput) -> _ConversationView
         if conversation_has_audio and not self.effective_include_audio(conversation_has_audio):
             recording = self._extract_recording(input)
             if recording is not None:
-                await transcribe_segments(recording)
+                config = input.scenario_state.config
+                if isinstance(config, ScenarioConfig):
+                    resolved_voice = resolve_voice_config(scenario_level=config.voice)
```
langwatch-agent, line 770: **P1 · Do not silently replace an existing STT provider**

- **What you're looking at.** Every normal scenario has a ScenarioConfig, so this branch resolves a new OpenAI STT default even when the caller supplies no voice configuration. Before this change, transcribe_segments used the provider registered through set_stt_provider.

- **Why it's scary.** An existing run configured for custom or local transcription now bypasses that provider. With OPENAI_API_KEY present it sends the recording to OpenAI instead; without it, transcription can degrade rather than use the working configured provider. A normally hidden DeprecationWarning is not a safe migration boundary for recorded audio. Snapshot the legacy provider into the per-run config when no explicit provider is supplied during the compatibility period, or fail visibly and require migration rather than silently changing destinations. Add a regression test that registers a custom provider, calls arun without voice, and proves no default OpenAI transcription occurs.

python/scenario/voice/config.py
```diff
@@ -0,0 +1,101 @@
...
+    stt: Any
+    tts: Optional[TtsConfig] = None
+
+
+def resolve_stt_provider(config: SttConfig) -> STTProvider:
+    """Resolve a ``provider/model`` descriptor to its concrete STT provider."""
+    provider, _, model = config.model.partition("/")
+    provider = provider.lower()
+    if provider == "openai":
+        return OpenAISTTProvider(model or OPENAI_STT_MODEL)
```
langwatch-agent: **P2 · Honor the credentials accepted by the new descriptor**

- **What you're looking at.** SttConfig accepts api_key and language, but the OpenAI resolver forwards only model. OpenAISTTProvider still constructs AsyncOpenAI() without the supplied key. Likewise TtsConfig.api_key is accepted but the simulator forwards only the voice string to synthesis.

- **Why it's scary.** A per-run OpenAI credential is silently ignored: the run uses the process account instead, or fails when only the descriptor key was supplied. That defeats per-run credential isolation. Thread the supported descriptor options into the actual clients and requests, and reject unsupported options rather than accepting no-ops. Cover descriptor-only credentials with the environment key absent and distinct concurrent run keys.

langwatch-agent (commented):
2 findings (1 P1, 1 P2); details and verification steps are inline. Static external-contribution review using the review skill; existing CI inspected, no PR code executed.

LangWatch-Review: verdict=findings sha=de396b6464cdd7958539f738021e9b9802cb65e2 p0=0 p1=1 p2=1 p3=0

langwatch-agent:
## Human Review Brief

> [!CAUTION]
> **Per-run voice provider resolution means credentials and model choice become run-scoped, so the question is what a run inherits when it names nothing.**
>
> Getting that default wrong is not a crash. It is a run that silently used a different provider than the author expected, which shows up as a cost line or a quality difference rather than an error.

| | |
| --- | --- |
| **Mode** | `Targeted`. |
| **Issue** | `#896`. |
| **State** | +328 / -27 across 11 files. Ready. `scenario` main CI is currently red, so read this PR's green as its own. |
| **Where to look** | The fallback chain, and whether Python and TypeScript now differ. |

<details>
<summary><b>What you are looking at</b></summary>

Scenario runs simulated conversations, and a voice provider turns text into audio for scenarios that exercise a voice agent. Until now the provider was presumably resolved once, globally. Resolving per run lets one suite mix providers, which is the right capability for comparing them.

Anything resolved per run needs an explicit precedence, and there are usually three layers: what the run says, what the suite or config says, and what the environment says. The failure that follows is not exotic: a run that names nothing picks up an environment variable set for a different purpose, and nobody notices because the run still succeeds.

The SDK ships in Python and TypeScript, and this PR is Python only. Any resolution rule added on one side is a divergence until the other follows.

</details>

<details>
<summary><b>What to check</b></summary>

**The precedence chain, written down.** Run, then config, then environment, and what happens when none of them answer. A clear error beats a silent default for something that costs money per call.

**Credentials.** A provider resolved per run means a key looked up per run. Confirm keys are not carried in the run object where they could reach a report artifact or a log line, since scenario reports are written to disk and rendered in a dashboard.

**Validation timing.** Resolving at run start and failing fast is much better than failing at the first synthesis call, halfway through a suite.

**Parity.** Worth a note in the PR saying the TypeScript side is following or deliberately not, so the next person does not read the absence as an oversight.

</details>

> [!NOTE]
> **Sensible capability. Pin down the fallback order and keep credentials out of the run object.**

python/scenario/judge_agent.py
atirna: Pushed a compatibility fix: each executor now snapshots the legacy STT provider before the run starts, so an unset voice path keeps the existing transcription provider. The regression swaps the global after arun() starts and still gets the original provider.

python/scenario/voice/config.py
atirna: Wired the per-run OpenAI STT and TTS keys to their clients, and the STT language to the transcription request. Unsupported TTS formats and per-run keys on providers without a matching client path now fail instead of falling back.

langwatch-agent (commented):
1 actionable finding; details inline.

Reviewed the changed code and relevant base callers/tests, with existing CI logs where available. The findings are not based on executing the PR.

External-contribution safety: reused the existing worktree; no PR code, dependency installation, package scripts, or live secret-bearing tests were executed. This is a COMMENT review, not approval or a claim that all CI passed.

LangWatch-Review: verdict=findings sha=b7103dddcf5fb3b627b22f5122c24b654ced5f1a p0=0 p1=1 p2=0 p3=0

python/scenario/judge_agent.py
```diff
@@ -764,7 +765,12 @@ async def _build_conversation_view(self, input: AgentInput) -> _ConversationView
...
+                    await transcribe_segments(recording, provider=resolved_voice.stt)
```
langwatch-agent, line 771: **P1 · Apply the run's STT provider before adapter transcription**

- **What you're looking at.** The per-run provider is passed only to the judge's missing-transcript backfill. VoiceAgentAdapter still calls voice.stt.transcribe(merged) during each turn, which selects the process-global provider.

- **Why it's scary.** For an audio-only adapter, a run configured with VoiceConfig(stt=...) first sends the audio through the global provider and its credentials. A successful transcript is recorded, so transcribe_segments(only_missing=True) then skips it and the requested provider is never used. This defeats per-run provider isolation and can send audio to a provider/account the caller did not select. Thread the run-resolved provider through the adapter's per-turn STT path too, and test a real audio-only VoiceAgentAdapter with a global provider that must not be called.

atirna:
pushed 7b158eca. per-turn STT resolves through the run carrier now, same resolution as the judge backfill, so the global default is never the one transcribing an audio-only turn.

the regression runs a real base-call drain with a global provider that must not be called: on b7103ddd it transcribes the turn, at this head 0 calls.

python/scenario/scenario_executor.py
```diff
@@ -275,8 +277,16 @@ def __init__(
...
             trace_wait_timeout=trace_wait_timeout,
+            voice=voice,
         )
         self.config = (ScenarioConfig.default_config or ScenarioConfig()).merge(config)
+        from .voice.config import VoiceConfig
+        from .voice.stt import get_stt_provider
+
+        voice_config = VoiceConfig.model_validate(self.config.voice or {})
+        if voice_config.stt is None:
+            voice_config.stt = get_stt_provider()
```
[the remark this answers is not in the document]
atirna: deliberate: set_stt_provider() users keep their provider inside a run instead of silently switching accounts. the executor snapshots the global once at run start so concurrent runs each carry a stable provider, which is what test_arun_without_voice_snapshots_the_legacy_stt_provider pins.

python/tests/test_arun_voice_config.py
```diff
@@ -0,0 +1,156 @@
...
+async def test_arun_without_voice_snapshots_the_legacy_stt_provider() -> None:
+    """Legacy configuration is sampled at the run boundary, not by the judge."""
+    previous = get_stt_provider()
+    provider = _STT("legacy", _MeetingPoint(parties=1))
+    replacement = _STT("replacement", _MeetingPoint(parties=1))
+    started = asyncio.Event()
+    release = asyncio.Event()
+
+    class _PausingAgent(_Agent):
+        async def call(self, input: AgentInput) -> AgentReturnTypes:
```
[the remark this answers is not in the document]
atirna: left this one: the parameter is typed as input throughout the codebase (agent_adapter, the adapters, the judge), so renaming it only in these tests would diverge from the code under test. nothing enforces A002 in the python package.

atirna:
pushed 918a4406 and 13acef95 for the open threads:

- `voice` is keyword-only at the end of `arun`/`run`/`ScenarioExecutor.__init__`, so positional callers keep their old bindings
- unsupported `voice` values raise a TypeError instead of silently running a fresh OpenAI default, and `TtsConfig` rejects an `api_key` for providers that dont take one at construction
- the type-ignores carry their reason, the voice exports are sorted, and the two simulator tests restore the previous `default_config`

one thread left open on purpose (the input rename): the whole codebase types that parameter as `input`, renaming only the tests would diverge from the code under test.

langwatch-agent (commented):
No new actionable findings. The earlier provider-isolation findings are addressed: executor construction snapshots the legacy provider for compatibility, descriptor credentials reach the clients, and adapter per-turn transcription now resolves the same run provider as the judge. Reviewed carrier copying, config merge behavior, explicit-key TTS cache partitioning, and the audio-only adapter regression test. Residual: no live-provider or concurrent-run tests were executed locally. Static external-PR review only: reused existing worktrees; no PR code, package scripts, dependency installs, or live credential-bearing tests were executed. Existing CI was inspected; this is not approval.

LangWatch-Review: verdict=clean sha=13acef9506bb79cb84279bd200f1f479b8016126 p0=0 p1=0 p2=0 p3=0

atirna:
fixed in 9be31d4: call() resolves the run voice through _effective_tts before _generate_text, so the spoken prompt applies when the voice comes from the run config. regression drives the real prompt selection and fails on the previous head.

langwatch-agent (commented):
No new actionable findings after the rebase. Compared the run-provider snapshot and credential plumbing with the previously reviewed head; the added prompt selection now follows the resolved per-run TTS voice. Earlier provider-isolation fixes remain present. Existing checks report no failures, but no live-provider, concurrent-run, or credential-bearing tests were executed locally. Static external-PR review only; no contributor code, dependencies or package scripts executed.

LangWatch-Review: verdict=clean sha=9be31d47df80d1bd65380cf8aa04bf50bdc74027 p0=0 p1=0 p2=0 p3=0

PR #982, opened 2026-09-08
feat(voice/twilio): a-leg external-number dialing for TwilioAgentAdapter

### LINKED ISSUE #762: TwilioAgentAdapter cannot test agents on external phone numbers (no A-leg Connect/Stream support)
## For Humans

**You cannot use Scenario to test a voice agent by phoning its published number.** The Twilio adapter's outbound call only works when the number you dial is one you own in your *own* Twilio account — because it attaches the audio stream by rewriting that number's voice webhook. Dialing an external production agent fails before it rings, and if the number somehow resolved, the adapter would overwrite the target's live webhook config.

**Ask:** let `placeCall` attach the Media Stream to the *caller* leg (`<Connect><Stream>` on our own outbound call), so a tester can run a scenario against a real deployed agent by number without owning it or touching its config.

- Blocker is a hardcoded A-leg TwiML (`<Say>…<Pause>`), not the tunnel — `publicBaseUrl` + a fixed port already works with cloudflared.
- Three safety guarantees the current B-leg flow gets for free must be rebuilt for the A-leg path: the media-stream WebSocket is currently unauthenticated, the `<Pause length="120">` call-duration backstop disappears under `<Connect>`, and there is no allowlist on outbound destinations (this becomes a dial-anyone-and-play-audio primitive).

_Status: enhancement, needs-plan, P2, not-ready. Full spec below._

## Investigation

Verified against `@langwatch/scenario` `origin/main` @ `c8eeb5e`, source under `javascript/src/voice/adapters/`. Reproduced live on 0.5.2 while attempting to run a scenario against a production voice agent reachable only by phone number (+447360542704, external to our account).

### Symptoms / context

`placeCall({ to })` is documented as "outbound — originates a call via Twilio REST" (`twilio.ts:10`), which reads as "dial any number." In practice it dials only numbers the caller **owns**. The real-world use case — phone a production voice agent to test it — fails before the call is placed.

### Root cause / key findings

**The outbound stream attaches on the B-leg (callee) by rewriting the callee's voice webhook — an account-internal loopback shape.** `placeCall` with `attachStreamToSelf` (default `true`, `twilio.ts:217`) does:
- `resolvePhoneNumberSid(args.to)` — looks up the CALLEE's `IncomingPhoneNumber` SID (`twilio.ts:221`).
- `writeVoiceUrl(calleeSid, ${publicBaseUrl}/twilio/voice)` — overwrites the callee number's voice webhook (`twilio.ts:225`).
- Then originates the A-leg with **hardcoded holding TwiML**: `<Say voice="Polly.Joanna">Thank you for calling. I will hold the line while you complete your scenario.</Say><Pause length="120"/>` (`twilio.ts:41`, `228-233`).

The Media Stream therefore rides the **callee's** leg, via the webhook that returns `<Connect><Stream url=wss…/twilio/stream/>` (`twilio-server.ts:231`).

`resolvePhoneNumberSid` throws on any number not in the account (`twilio-shared.ts:418-429` — non-ok response or empty result → generic `Error`), so an external `to` fails **before dialing**. If it somehow resolved, `writeVoiceUrl` would **overwrite the target's production voice webhook** — a destructive side effect on someone else's config.

**`attachStreamToSelf: false` is not an escape hatch.** No Media Streams socket ever opens, so `sendAudio`/`receiveAudio` throw (`_assertStreamLive`, `twilio.ts:366-370`), and the A-leg still plays the hardcoded Say+Pause into the callee while capturing nothing.

**There is no A-leg `<Connect><Stream>` path.** `placeCall` accepts only `{ to, timeoutMs, attachStreamToSelf }` (`twilio.ts:208-212`); the REST helper accepts only inline TwiML, deliberately no `twimlUrl` (SSRF concern, `twilio-shared.ts:448-454`). The standard Twilio shape for streaming an outbound call to your own socket — putting `<Connect><Stream>` on the *originated* leg's TwiML — is unreachable.

**Tunnel is NOT the blocker.** `publicBaseUrl` + a fixed `httpPort` already lets `cloudflared tunnel --url http://localhost:<port>` front the webhook server; ngrok/`NGROK_AUTHTOKEN` is optional. The gap is purely the hardcoded A-leg TwiML.

### Challenge findings (devils-advocate, run on the A-leg-attach approach)

Proposed approach — when calling an external number, make the originated call's inline TwiML `<Connect><Stream url="wss://<publicBaseUrl>/twilio/stream"/>` and touch no callee webhook — is **sound at its core but must ship with three safety items the B-leg flow got for free**:

1. **Mechanism is correct.** `<Connect><Stream>` (not `<Start><Stream>`, which is listen-only/fork) gives a **bidirectional** socket; Twilio plays back `media` frames we send, and `sendAudio` already emits them (`twilio-shared.ts:339`). AC must assert user-sim audio sent via `sendAudio` is actually heard on the call, not just that inbound audio arrives.
2. **The `<Pause length="120">` call-duration backstop disappears.** Under `<Connect>` the call lives exactly as long as the WS — a hung executor keeps a **live PSTN call to an external party open and billing indefinitely**. Needs an explicit max-call-duration guard (server-side WS-close timer or Twilio `timeLimit`), distinct from the connect timeout (`timeoutMs` default 120s, `twilio.ts:218`).
3. **The media-stream WebSocket is UNAUTHENTICATED — and external mode removes the only auth checkpoint.** Signature validation lives only on `POST /twilio/voice` (`twilio-server.ts:188`). The WS upgrade at `/twilio/stream` validates nothing (`twilio-server.ts:90-99`) and `_signalStreamConnected()` fires on the first `start` frame regardless of which call it belongs to (`twilio-server.ts:274-277`). In B-leg flow the authenticated voice POST precedes the socket; in A-leg flow the WS connect is the ONLY inbound signal, so anyone who learns the public tunnel URL can open `wss://…/twilio/stream`, send a `start` frame, and hijack/inject audio into the live external call. Needs a per-call nonce in the `<Stream url>` query (or `<Parameter>`) matched on the `start` frame.
4. **No callSid correlation → first-socket-wins race.** `_signalStreamConnected` ignores `frame.callSid` (`twilio-server.ts:274-277`; `twilio.ts:342-344` ignores subsequent SIDs). Once the A-leg stream is the only signal, a stale/probe socket connecting first resolves `placeCall`'s `promiseWithTimeout` against the wrong stream. `mediaStreamLoop` must match `frame.callSid` against the `placeCall` SID before signalling connected.
5. **Auto-detecting external mode on a `resolvePhoneNumberSid` 404 is unsafe.** That function throws a generic `Error` on 404 AND on 429/500/outage (`twilio-shared.ts:424-429`) — a transient blip would silently convert an intended-internal call into an external one. External mode must be an **explicit flag**, never inferred from a resolve failure.
6. **Unguarded robocall/vishing primitive.** `allowedCallers` gates inbound only (`twilio.ts:123`); there is no allowlist on `placeCall`'s `to`. A-leg external attach + leaked tunnel URL = dial-anyone-and-play-synthesized-audio. Should mirror `allowedCallers` with an `allowedCallees`/destination allowlist or explicit opt-in.
7. **`disconnect()` stays a safe no-op for external mode** (the `call`-mode `writeVoiceUrl` restore at `twilio.ts:177` is guarded by an `if` on `_calleePhoneNumberSid`, which external mode never sets) — but this must be asserted, not assumed.
8. **`sendDtmf` breaks in external A-leg mode.** `sendDtmfOnCall` POSTs `<Response><Play digits/></Response>` to `Calls/{sid}.json` (`twilio-shared.ts:468-472`), which **replaces** the call's TwiML — tearing down the active `<Connect><Stream>` (our own leg now). Must be documented unsupported in external mode or redesigned to in-band tones.
9. **Tunnel-readiness before origination.** In external mode Twilio dials the A-leg then opens the WS from Twilio's network; a not-yet-ready `publicBaseUrl` leaves a **live, silent call already ringing the external party** before `placeCall` times out. Connectivity/readiness should be asserted before originating.

### Strategies tested / considered

| Approach | Result |
|---|---|
| `placeCall({ to: external })`, default `attachStreamToSelf: true` | Fails pre-dial: `resolvePhoneNumberSid` 404s on a number we don't own (`twilio-shared.ts:424`). Would overwrite target webhook if it resolved. |
| `placeCall({ to: external, attachStreamToSelf: false })` | No stream opens; `sendAudio`/`receiveAudio` throw (`twilio.ts:366`). A-leg plays Say+Pause into callee, captures nothing. |
| cloudflared tunnel via `publicBaseUrl` + fixed `httpPort` | **Works** for the transport layer — not the blocker. |
| `rest`-helper TwiML override (inject `<Connect><Stream>` via the test seam) | Ruled out: "hacking around" an unsupported path against a production target; also no auth/duration/allowlist guards. |
| **A-leg `<Connect><Stream>` (proposed feature)** | Correct mechanism; requires the 3 load-bearing safety items above to ship with it. |

- The webhook server already decodes µ-law, validates `X-Twilio-Signature` (on the POST only), and signals stream-connected — the A-leg path reuses all of it; the new work is (a) an A-leg TwiML branch, (b) WS auth via per-call nonce, (c) a call-duration guard, (d) a destination allowlist.
- Ruled out: tunnel/ngrok is not the gap; `attachStreamToSelf: false` is not a workaround; auto-detect-on-404 is a trap.
- `<Connect><Stream>` is bidirectional (verified against Twilio Media Streams semantics); `<Start><Stream>` would silently break `sendAudio`.

### Caveats

- Validate `<Connect><Stream>` bidirectional playback empirically against a real Twilio call — the AC below requires observing sim audio on the callee leg, not just inbound capture.
- Confirm whether Twilio's `timeLimit` REST param or a server-side WS-close timer is the right duration backstop; both are viable, pick one and make it configurable.
- Confirm the exact frame that carries a custom `<Parameter>`/query nonce (`start` frame `customParameters`) before designing the WS-auth match.

### Core capability

- Given an adapter connected with a valid `publicBaseUrl` and an **explicit external-mode flag** (e.g. `placeCall({ to, attachStream: "a-leg" })` or equivalent), when `placeCall` dials an E.164 number **not** in the caller's Twilio account, then the call is originated with `<Connect><Stream url="wss://<publicBaseUrl>/twilio/stream"/>` inline TwiML and **zero** `writeVoiceUrl`/`resolvePhoneNumberSid` REST calls are made against the callee. Evidence: unit test asserting the originated TwiML string + a spy on the REST helper showing no callee-webhook reads/writes.
- Given external mode is active and the Media Stream connects, when the user simulator emits audio via `sendAudio`, then that audio is played into the live call (bidirectional `<Connect><Stream>`), and inbound callee audio is delivered to `receiveAudio`. Evidence: live integration test (or recorded Twilio call) confirming both directions; NOT inbound-only.
- Given external mode, when the callee number is not in the account, then `placeCall` does **not** call `resolvePhoneNumberSid` on it and does **not** attempt to read or write its voice webhook. Evidence: REST-helper spy asserts zero calls touching the callee SID.

- **Failure mode — WS hijack:** Given external mode, when a WebSocket connects to `/twilio/stream` whose `start` frame lacks (or mismatches) the per-call nonce minted by this `placeCall`, then the socket is rejected/closed and `_signalStreamConnected` does **not** fire for it. Evidence: unit test opening a nonce-less/wrong-nonce socket and asserting rejection + that a subsequent correct-nonce socket is the one that connects.
- **Failure mode — wrong-call socket race:** Given external mode with a live `placeCall`, when a `start` frame arrives whose `callSid` differs from the SID `placeCall` originated, then it is ignored and the connect deferred resolves only for the matching `callSid`. Evidence: unit test feeding a mismatched-then-matching `start` frame and asserting connect resolves once, on the right SID.
- **Failure mode — runaway call duration:** Given an external call whose stream never terminates (executor hangs), when the configured max-call-duration elapses, then the adapter closes the WS / ends the call so no PSTN call outlives the guard. Evidence: test with a fake clock asserting the call is torn down at the limit; the connect timeout and max-call-duration are two distinct, independently-configurable values.
- **Failure mode — transient resolve error must not silently go external:** Given internal (default) mode, when `resolvePhoneNumberSid` throws (404 or 429/500), then the error surfaces and the call is **not** silently re-routed to external A-leg mode. Evidence: test injecting a thrown resolve error and asserting no external-mode fallback occurs.
- **Failure mode — tunnel not ready:** Given external mode with an unreachable `publicBaseUrl`, when `placeCall` is invoked, then readiness is checked **before** the Twilio REST origination call and, on failure, `placeCall` rejects with a `TunnelNotReadyError` (or equivalent) having issued **zero** `Calls.create` REST calls. Evidence: unit test stubbing an unreachable `publicBaseUrl`, asserting the origination REST helper is never called and `placeCall` rejects.
- **Abuse surface — destination allowlist:** Given an `allowedCallees` (destination) allowlist is configured, when `placeCall` targets a number not on it in external mode, then the call is rejected before origination (zero `Calls.create` REST calls). Evidence: test asserting rejection for a disallowed destination with no origination REST call.
- **Abuse surface — external mode opt-in:** Given no `allowedCallees` allowlist is set, when external mode is requested, then `placeCall` refuses unless an explicit, typed opt-in is present (the opt-in mechanism — flag, config field, or env var — is pinned in the plan step). Evidence: test asserting external mode without the opt-in throws, and with it proceeds.
- **Ripple — `sendDtmf` in external mode:** Given external A-leg mode, when `sendDtmf` is called, then it either raises a clear "unsupported in external mode" error or injects tones in-band without replacing the call TwiML (it must NOT tear down the active `<Connect><Stream>`). Evidence: test asserting the active stream survives (no `Calls/{sid}.json` TwiML-replace) or a clear throw.
- **Regression — internal loopback unchanged:** Given the existing default flow (callee number owned by the account, `attachStreamToSelf: true`), when `placeCall` runs, then behavior is byte-identical to today (B-leg webhook rewrite, Say+Pause A-leg, webhook restored on `disconnect`). Evidence: existing `twilio.test.ts` suite passes unchanged.
- **Regression — disconnect no-op for external:** Given an external call has completed, when `disconnect()` runs, then it issues **zero** `writeVoiceUrl` calls (no callee webhook was touched; `_calleePhoneNumberSid` stays unset). Evidence: REST-helper spy asserts zero restore writes after an external call.
- **Rollback — no persistent external state:** Given any external call (success or failure), when it ends, then the adapter leaves no callee-side webhook mutation to roll back (external mode never writes one). Evidence: assertion that `_calleePhoneNumberSid`/`_priorCalleeVoiceUrl` remain undefined throughout an external call.

## Plan

_Not started — plan step deferred per investigate-only instruction. Issue remains not-ready._

</details>

drewdrewthis:
Closes #762.

Lets a scenario test a real deployed voice agent on a phone number the Twilio account does **not** own, by attaching the Media Stream to the originated (A) leg via inline `<Connect><Stream>` instead of rewriting the callee's `voice_url`. Python and JS in lockstep.

Shipped in 5 slices, one commit each, plus a review-fix commit and a live use-proof against real Twilio (below). The three guardrails ship **with** the capability, not after it.

## Progress

| Slice | Scope | ACs | Status |
|---|---|---|---|
| 1 | A-leg core: typed `attach_stream`, inline `<Connect><Stream>`, zero callee REST | AC1–AC4 | ✅ committed |
| 2 | Guardrail (a): WS auth via per-call nonce + callSid correlation | AC5, AC6, AC13 | ✅ committed |
| 3 | Guardrail (b): max call duration — Twilio `TimeLimit` + adapter timer | AC7, AC12 | ✅ committed |
| 4 | Guardrail (c): `allowed_callees` default-deny + tunnel readiness probe | AC8, AC9 | ✅ committed |
| 5 | `sendDtmf` unsupported in external mode + frame-loop tripwire | AC10, AC11, AC14 | ✅ committed |

## Slice 1 — what landed

- `place_call`/`placeCall` gain `attach_stream: "a-leg" | "b-leg"`. A-leg originates with `<?xml …?><Response><Connect><Stream url="wss://<host>/twilio/stream"/></Connect></Response>` — the same TwiML the inbound webhook already returns, now on our own leg.
- **A-leg touches nothing on the callee:** zero `resolve_phone_number_sid` / `read_voice_url` / `write_voice_url`. `disconnect()` is consequently a no-op, and the callee-state fields stay unset on both the success and the stream-timeout path.
- **External mode is never inferred.** A b-leg resolve failure (404- or 500-shaped) surfaces to the caller; there is no fallback into a-leg.
- Legacy `attach_stream_to_self` still honoured — `True`/unset → b-leg, `False` → originator-only. Passing both with disagreeing intent raises before any REST call.
- The `https→wss` derivation is extracted into one shared helper per language, used by both the webhook server and origination — no duplicated string-munging.

### Backwards compatibility

B-leg is the default and is byte-identical. A golden regression test pins both the emitted B-leg TwiML string *and* the ordered callee REST sequence (`resolve` → `read` → `write(webhook)`, then `write(prior)` on disconnect), so any drift fails with a message pointing at "B-leg behaviour changed".

### Verification

| | Before | After |
|---|---|---|
| `pytest tests/voice -k twilio` | 24 (adapter file) | **109 passed, 2 skipped** |
| `vitest run src/voice/` | 23 (adapter file) | **559 passed, 1 skipped** |
| `pnpm typecheck` / `pnpm lint:lib` | — | clean |

The 2 skips are the env-gated live e2e. **No real Twilio calls were placed at any point in this work.**

## Acceptance criteria

The AC set was reviewed before implementation and amended — see [the amendment comment](https://github.com/langwatch/scenario/issues/762#issuecomment-5583131033). Two Must-Fix gaps were found in the guardrails:

- **AC12** — §3(b) mandates *two* duration protections but only the adapter-side timer had an AC. The timer is exactly the half that cannot fire if the adapter process hangs, which is the runaway-billing case. AC12 pins `TimeLimit` in the `Calls.create` body.
- **AC13** — a hardcoded constant nonce would pass AC5 unchanged while fully defeating the WS auth. AC13 pins per-call uniqueness and CSPRNG length.

Also added **AC14**, an ungated mock-level frame-loop tripwire, because AC11 is env-gated and would never trip in CI.

The full contract is now in `specs/voice-twilio-a-leg-external.feature` (14 scenarios, one per AC).

## Slice 2 — what landed

In b-leg mode the signed `POST /twilio/voice` precedes the socket, so the socket inherits that trust. **In a-leg mode the socket is the only inbound signal**, and the WS upgrade authenticated nothing — it matched on path alone. A leaked tunnel URL was audio injection into a live PSTN call.

- `place_call`/`placeCall` mint a **16-byte CSPRNG nonce per call** (128 bits, lowercase hex — survives Twilio's `<Parameter>` round-trip with no alphabet, padding, or case ambiguity) and ship it as `<Parameter name="nonce">` inside `<Stream>`. The frame parser now surfaces `start.customParameters`.
- The media loop **closes** any socket whose `start` frame cannot present the nonce (timing-safe compare), and **ignores** frames whose `callSid` is not the originated call, so a stale or probe socket cannot win the race to the connect signal.
- **Enforcement is armed by the mode we originated in**, not by whether the inbound frame happens to carry a nonce — so omitting the `<Parameter>` is a rejection, not a bypass.
- **An unauthenticated socket now gets no adapter state at all.** Previously the socket became `_stream_ws` — the target of `send_audio` — and purged the per-call inbound queue *before* any check ran. That was the actual injection hole; adoption is now deferred until a `start` frame authenticates.
- b-leg mints no nonce, stays byte-identical and un-gated; the Slice 1 golden test passes untouched. The Python/JS TwiML attribute-escaping asymmetry flagged in Slice 1 is closed.

### Verification

Python **114 passed / 2 skipped** (was 109/2) · JS **564 passed / 1 skipped** (was 559/1) · typecheck + lint clean.

Falsification checked rather than assumed: replacing the mint with a **constant** nonce — the failure mode AC13 exists to catch, which would pass every other test while fully defeating the auth — fails exactly one test, `test_a_leg_mints_a_fresh_nonce_per_call`. Disarming enforcement fails both AC5 cases and AC6.

## Slice 3 — what landed

Under a-leg's `<Connect>` the call lives exactly as long as the WebSocket. The implicit ceiling `<Pause length="120"/>` gave b-leg is gone, so a hung or killed executor keeps a **billing PSTN call open indefinitely**. Two independent mechanisms restore it:

- **Twilio-side (load-bearing):** `TimeLimit` on `Calls.create` — a-leg only, absent in b-leg. This is the half that still fires when the adapter process hangs, is SIGKILLed, or loses the network, which is the actual runaway-billing case.
- **Adapter-side:** a wall-clock timer that hangs up via `POST /Calls/{sid}.json` `Status=completed` and closes the socket, covering a misconfigured or dropped `TimeLimit`.

`max_call_duration_seconds` / `maxCallDurationSeconds` is per-call, **defaults to 300s, capped at 300s**. A larger request **raises rather than clamping** — a silent clamp hands the caller a shorter call than they asked for and they debug the wrong thing. Naming it outside a-leg mode is also rejected rather than ignored, so no caller believes a call is bounded when it isn't. Both are distinct from `timeout`/`timeoutMs`, which bounds waiting for the *stream to connect*, not the call's length.

The timer is disarmed on `disconnect()`, at stream end, and on re-arm, so it can never hang up an already-ended call or a **later** call's SID — pinned by a dedicated teardown test.

### Verification

Python **126 passed / 2 skipped** (was 114/2) · JS **576 passed / 1 skipped** (was 564/1) · typecheck + lint clean.

Falsification found a real hole in the first draft of the tests: the AC12 assertion ran against an adapter-level spy that **replaces the REST helper entirely**, so dropping `TimeLimit` from the wire did not fail it. Replaced with true wire-level tests (fake `fetch` in JS, fake Twilio client in Python) that observe the actual request. All three mutations — drop `TimeLimit`, never arm the timer, no-op the cancel — now fail exactly their own tests.

## Slice 4 — what landed

`allowed_callers` gated inbound calls; `place_call`'s `to` was **completely unguarded**. In a-leg mode that is an unbounded dialer, because a-leg dials numbers this account does not own.

- New constructor field `allowed_callees` / `allowedCallees`, **deny-by-default**: with it unset, a-leg `place_call` refuses. b-leg is deliberately **not** gated — it can only reach numbers on this account, which is its own guardrail.
- Entries are normalised and **E.164-validated at construction**, so a typo fails at setup rather than when a PSTN call is about to be billed.
- Membership is an exact set lookup on the normalised number — never `in`/`includes`/`startsWith` — so a prefix or suffix of an allowed entry is refused. Three near-miss cases per language pin this.
- Refusal happens **before origination**: zero REST calls, no nonce minted, asserted by spy.

Also in the same pre-origination slot: an optional **edge-readiness probe**, satisfied structurally by the existing `CloudflareTunnel.wait_until_edge_reachable` and wired in by `TwilioHarness`. An unreachable public URL now raises `TunnelNotReadyError` instead of letting Twilio dial out against a dead tunnel — which the caller pays for and sees only as a 120s stream-connect timeout.

Check order is cheapest-first: mode → allowlist → duration → probe → origination.

### Known gap (JS)

The JS side has the `TunnelReadiness` interface, the `TunnelNotReadyError` and the `placeCall` call site, but **no concrete probe is wired in** — no JS tunnel implementation (`twilio-tunnel.ts`, ngrok/localtunnel) exposes an edge-reachability check, and there is no JS `TwilioHarness`. AC9 protection is therefore Python-only in practice. This is a missing implementation, not a behavioural divergence in `placeCall`.

### Also noticed, deliberately not changed

`allowed_callers` validates nothing today — entries are not checked as E.164. `allowed_callees` is validated. Bringing the older field up to the same standard is out of scope for this PR.

### Verification

Python **138 passed / 2 skipped** (was 126/2) · JS **588 passed / 1 skipped** (was 576/1) · typecheck + lint clean.

Seven falsification mutations, each caught. Both default-deny mutations (`default-deny` → `default-allow`) failed **exactly one test each** — the right one. Moving the tunnel probe to *after* origination was caught by the ordering assertion, not just by skipping it.

## Slice 5 — what landed

**AC10 — guard what a-leg cannot support.** `send_dtmf`/`sendDtmf` works by replacing the live call's TwiML. Under b-leg that was harmless; under a-leg it replaces the `<Connect><Stream>` verb **the scenario is running on** and tears the media stream down mid-call. It now refuses with an error that explains that, armed off the per-call stream nonce so b-leg is untouched.

A survey of every REST call site found **no other operation** needing the guard: `end_call` is the duration watchdog's deliberate teardown (guarding it would disarm guardrail (b)), and the callee `voice_url` writes are already unreachable in a-leg because `_callee_phone_number_sid` is never set there.

**AC14 — ungated tripwire.** The live e2e is env-gated and never runs in CI, so nothing ungated proved the a-leg media loop actually carries audio. Now a test drives the **production** media loop over a scripted in-memory socket and asserts on real bytes both ways: inbound µ-law decodes to expected PCM16, outbound PCM16 comes back as correctly-framed 160-byte µ-law. Expected values come from an **independently written G.711 reference decoder** in the test file, so the shipped codec cannot agree with the assertion by construction, and inbound fixture frames are built by the test rather than by the adapter's own frame builder — an outbound-framing bug cannot cancel an inbound-parsing bug.

> ⚠️ **This claim was false, and review caught it.** The fixture was a *constant* µ-law byte repeated, so the assertion reduced to set membership — invariant under any permutation. Reversing every sample in the shipped decoder still passed. It also exercised 1 of 256 µ-law code points. Rebuilt in `cfde4454`; see the review section below.

**AC11 — live smoke: written, gated, and NOT run.** Behind the existing Twilio env gate plus an explicit `SCENARIO_TWILIO_EXTERNAL_TO` destination. Confirmed reporting `skipped` in this environment. **No Twilio or network call was made at any point in this PR.**

### Verification

Python **141 passed / 3 skipped** (was 138/2) · JS **591 passed / 1 skipped** (was 588/1) · typecheck + lint clean.

---

## Review pass — `cfde4454`

Four reviewers (security, principles, hygiene, tests) audited the five slices. **7 merge blockers and 12 follow-ups** found and fixed in one commit. Every blocker was independently reproduced before the fix and falsified after it.

### The three that mattered

**An unauthenticated socket could kill a live call.** The nonce check correctly denied an attacker socket all adapter state — but the wrapper *around* it nulled `_stream_ws`/`_stream_sid` in a `finally` on **every** socket close. Anyone who knew the public tunnel URL could connect and immediately hang up, killing the genuine call's media while the PSTN leg kept billing to the 300s cap. No nonce needed, repeatable at will. The reset is now identity-guarded to the socket that *is* the transport.

The tests missed it because every a-leg socket test drove `media_stream_loop` directly and never the production wrapper `run_stream_session` — **the exact shortcut PR #697 was already blocked for**. Now covered through the wrapper.

**Python's nonce check raised instead of rejecting.** `secrets.compare_digest` on `str` raises `TypeError` on non-ASCII input. A socket sending `{"nonce":"é"}` turned the auth primitive into an exception that unwound through the live call and stamped its reason as `"error"`. Now compares UTF-8 bytes, like the JS twin always did.

**Python leaked a nonce across calls.** `_stream_nonce` was assigned only inside the a-leg branch, so an a-leg dial followed by a b-leg dial on the same adapter left the nonce armed — the b-leg `start` frame carries no `<Parameter>`, so **the b-leg call never connected**. JS assigned unconditionally and was already correct. Neither suite had a two-sequential-`place_call` test; both do now.

### Three tests that were passing for the wrong reason

This is the theme of the review, and it recurred at every slice.

| Test | Passed despite | Now |
|---|---|---|
| **AC14** frame-loop tripwire | reversing every sample in the shipped decoder | 256-code-point waveform, asserted **per-sample and positionally** against an independent reference codec, both directions |
| **AC13** nonce entropy | dropping `secrets.token_hex` → `random.getrandbits`, or `randomBytes` → `Math.random()` | asserts the CSPRNG call itself |
| **AC8/AC9** harness wiring | deleting both guardrail lines from `TwilioHarness` — its only production construction — while all 12 guard tests stayed green | construction test asserting identity, not truthiness |

Same shape as the AC12 spy bug caught during slice 3: the assertion sat on a seam that the regression routes around.

### Also fixed

- **Billing leak on the ordinary failure path** — an a-leg stream-connect timeout (default 120s) left the originated call running to the 300s cap. It now hangs up before re-raising.
- **Unadopted sockets could rewrite a live call's verdict** via the close, error and — a path the reviewers missed — bare `stop` frame.
- **A live, assignable UK mobile** (`+447911123456`) was used ~25 times across both suites, including as an `allowedCallees` entry and an a-leg destination. A copy-paste into live config would dial a stranger. Now Ofcom's drama range.
- Python's inbound webhook interpolated the stream URL into TwiML **unescaped** while JS escaped it — and this branch added the helper one function away.
- Raw phone numbers leaked through `validate_e164` error messages, bypassing `_redact_e164`.
- JS test fixtures had drifted to **two different values for `ORIGINATED_CALL_SID`**, which silently steers a test into the media loop's "ignore" branch — passing for the wrong reason. Collapsed onto one harness per language.
- Sticky `max_duration` verdict, orphaned timer task, `a-leg` + `attach_stream_to_self=False` conflict now raising, `TunnelReadiness` exported, fixed-clock test sleeps replaced with waits on an observable signal.

**One reviewer finding was wrong** and was not acted on: JS `_setCallSid` was reported as unguarded, but it already checks `if (!this._callSid)`, matching Python. Verified directly.

### Falsification

Mandatory for all four critical fixes, in both languages: each reintroduced defect made the new test fail, each restore went green. The AC14 sample-reversal mutation was additionally re-run by hand after commit — it now **fails** the tripwire, where before it passed.

### Verification

Python **150 passed / 3 skipped** (was 141/3) · JS **599 passed / 1 skipped** (was 591/1) · typecheck + lint clean.

### Spec

`specs/voice-twilio-a-leg-external.feature` AC13 and AC14 Then-clauses were sharpened, because the old wording provably did not describe what the tests now assert. `specs/voice-agents.feature` and its pinned contract test are untouched.

### Deliberately out of scope

Filed rather than fixed: comment de-duplication, the `PLACE_CALL_A_LEG_SAY_TEXT` misnomer, connection/rate limits on `/twilio/stream`, a JS `TwilioHarness`, and unit tests for `CloudflareTunnel.wait_until_edge_reachable` itself.

Falsification: broke the µ-law decode, broke the outbound framing, and removed the DTMF guard. Each mutation failed the intended test; the decode and guard mutations failed **only** that test. All restored and diff-verified before the final run.

### Post-undraft hardening — `600811d8`, `a33c6256`

Undrafting ran the full CI job and the bots for the first time; two real auth gaps came out of it, both fixed in Python and JS with a regression test each:

- **Nonce read at the `start` frame, not socket entry** (`600811d8`). Both servers snapshotted the auth state at loop entry, so a socket that connected before `place_call` armed the nonce was adopted without auth. CodeRabbit finding, confirmed by an independent security review.
- **Every frame from an un-adopted socket is dropped** (`a33c6256`). `media`/`dtmf`/`stop` were processed before the socket had passed `start`-frame auth, so audio could be injected into a live a-leg call. Found by the security review of `600811d8`, then independently by CodeRabbit.

Declined with reasons on the thread: wiring `TunnelReadiness` into the JS adapter by default (opt-in by design; JS has no tunnel-owning harness). Also in this stretch: `81115be9` exports `TunnelNotReadyError` from the `scenario` root (caught by the ready_for_review run), pyright Optional narrowings, and CodeQL hygiene.

### Reviewer findings P1/P2 — `9ae45ef5`, `c3a8efc3`, `24a98990`

- **P1** (sergioestebance): a socket with no nonce could arrive after `connect()` but before `place_call()`, get adopted, and keep ownership of the call after the nonce was minted. Fixed in `9ae45ef5`: each adapter now tracks a call generation, bumped on every dial or wait; dialing resets the connection state; the media loop captures the generation at adoption time and checks it again before every side effect; a stale socket is closed on its next frame. Regression tests: `test_a_leg_socket_adopted_before_dialing_is_evicted` / `"#762 P1: a socket adopted before dialing is evicted and changes nothing"`.
- **P2** (sergioestebance): the max-duration handler checked the call generation before awaiting `endCall`, then closed whatever socket was live when the await returned — closing the wrong call's socket if a new call had started in between. Fixed in `c3a8efc3`: the generation is rechecked after the REST await, so only the expired call's socket is closed. A follow-up fix in `24a98990`, found in our own adversarial re-review: the guard is now keyed on the shared call generation (a b-leg or `waitForCall` takeover was invisible to the old re-arm counter), and the `max_duration` ended-reason is gated the same way. Tests: `test_expired_watchdog_in_flight_does_not_close_a_newer_calls_socket`, `test_stale_watchdog_after_takeover_never_relabels_the_newer_call`, with JS twins in `twilio-call-duration.test.ts`.
- **Test-idiom migration**: `place_call` now blocks until a socket authenticates, so tests moved from pre-setting `_stream_connected` to concurrent-dial helpers (`_dial_then_connect`, `dialAndConnect`).
- Counts at `24a98990`: Python 1549 passed / 10 skipped, pyright 0 errors; JS 319 passed / 2 skipped, typecheck and lint clean.
- Runs 8b/9 at `24a98990`: Python `1 passed`, frames_received 1501, judge success, call `CA934c6a7a3a8d5e8760e9dfb786a22588` completed 30s, 0 child legs (a-leg). JS `1 passed`, bidirectional audio asserted, call `CA8ee0ca04ff116bd8ffe648a9e4db9bf2` completed 31s, 0 child legs. Run 8 at `c3a8efc3` failed with Twilio 31900 (quick tunnel not routable at dial time), re-run clean; not a code regression.

### AC11 bound in JavaScript — `499438eb`

- New env-gated `twilio-a-leg-external.e2e.test.ts` (JS twin of the Python live test; `describe.skip` unless `TWILIO_ACCOUNT_SID`, `TWILIO_AUTH_TOKEN`, `TWILIO_PHONE_NUMBER`, `SCENARIO_TWILIO_EXTERNAL_TO` are all set — skipped in CI, 1 skipped locally without env).
- `openTwilioTunnel` gains an explicit `"cloudflared"` provider (JS peer of Python's `CloudflareTunnel`); default ngrok/localtunnel selection unchanged.
- Every AC in `specs/voice-twilio-a-leg-external.feature` (AC1–AC14) now has a binding test in both languages.

## Live use-proof — real Twilio call, a-leg mode

Recorded 2026-09-09 03:50Z from the branch head with the env-gated e2e `python/tests/voice/test_twilio_a_leg_external_e2e.py`:

- Twilio call `CA03274362692a43f124bf7fa190dc915f`: `outbound-api`, `completed`, **30s**, $0.024, **zero child legs** — the Media Stream rode the originating leg via `<Connect><Stream>`, no REST call ever touched the callee.
- Server log: `placed call … from ***3520 to ***3578` → `WS connection accepted` ~0.9s later (nonce-authenticated socket).
- OTel spans in the run: `voice.audio.send` 161,280 bytes out; `voice.audio.receive` 300 chunks / 1,437,840 bytes in; `voice.stt.transcribe` 141 chars.
- Callee (a TwiML echo line) came back transcribed verbatim: *"Hello, this is the test line. I can hear you loud and clear. Please go ahead. Thank you. I heard what you said. This line is working. Goodbye"*.
- Judge (`gemini-2.5-flash`): `verdict: success`, all three criteria `true`.
- Result: `1 passed in 63.54s`.

Providers used for the run: ElevenLabs TTS+STT and Gemini via litellm (no OpenAI key was available), through the new `SCENARIO_TWILIO_E2E_VOICE/_STT/_MODEL` overrides in `08776850`; defaults remain the OpenAI stack. Earlier attempts (`CAe7c0…`, `CAb1f3…`) dialled the account's second number and came back `busy` in 0s — that number has never received an inbound leg at the carrier, so the roles were swapped for the proof. The callee number's `voice_url` was restored to its prior value afterwards (verified via REST).

Proof captures (redacted):

live a-leg e2e — pytest pass with transcript, spans and judge verdict

Twilio REST record of the proof call — completed, 30s, no child legs, numbers restored

Proof review (`proof-reviewer`): ACCEPT-WITH-NOTES — evidence is real and mutually corroborating; the `<Connect><Stream>` verb is inferred from "WS accepted + zero child legs" rather than quoted from the log.

### Re-proof at post-hardening heads

| Run | Head | Language | Test | Result | Call SID | Twilio |
|---|---|---|---|---|---|---|
| 6 | `a33c6256` | Python | `tests/voice/test_twilio_a_leg_external_e2e.py` | 1 passed (47.5s), frames_received=1500, judge success | `CAc042f68bf8544f50ed84abe36f8eaf1a` | completed, 30s, $0.024, 0 child legs |
| 7 | `499438eb` | JavaScript | `src/voice/adapters/__tests__/twilio-a-leg-external.e2e.test.ts` | 1 passed (47.8s), frames_received=1500, judge success | `CAb256cfe04bb34d5334af734c40df5766` | completed, 30s, $0.024, 0 child legs |
| 8b | `24a98990` | Python | `tests/voice/test_twilio_a_leg_external_e2e.py` | 1 passed, frames_received=1501, judge success | `CA934c6a7a3a8d5e8760e9dfb786a22588` | completed, 30s, 0 child legs |
| 9 | `24a98990` | JavaScript | `src/voice/adapters/__tests__/twilio-a-leg-external.e2e.test.ts` | 1 passed, bidirectional audio asserted | `CA8ee0ca04ff116bd8ffe648a9e4db9bf2` | completed, 31s, 0 child legs |

Both runs dialled Twilio number ending 3578 from number ending 3520 with the callee pointed at a scripted echo TwiML for the run and restored to its exact prior voice_url afterwards (verified by re-fetch). 0 child legs = the stream rode our own leg; the callee was never touched over REST.

## Human verification

1. Open the Twilio console → Monitor → Calls → `CA03274362692a43f124bf7fa190dc915f`: one outbound leg, 30s, no child call.
2. From `python/` with Twilio env + `SCENARIO_TWILIO_EXTERNAL_TO=<a number that answers>`: `uv run pytest tests/voice/test_twilio_a_leg_external_e2e.py -q` — expect `1 passed`, a `WS connection accepted` line, and a transcribed assistant turn.
3. Without `SCENARIO_TWILIO_EXTERNAL_TO` the same test **skips** — no accidental billing.
4. `uv run pytest tests/voice -q -k twilio` → 150 passed / 3 skipped; `cd javascript && ./node_modules/.bin/vitest run src/voice/` → 599 passed / 1 skipped.

- AC1–AC10, AC12–AC14: unit/golden/wire tests listed per slice above, all green at head.
- AC11 (real external call, audio both ways): the live run above — Twilio SID, span byte counts, verbatim transcript, `1 passed`.
- Run 6 (Python, `a33c6256`): `CAc042f68bf8544f50ed84abe36f8eaf1a`.
- Run 7 (JavaScript, `499438eb`): `CAb256cfe04bb34d5334af734c40df5766`.
- Guardrails hold on the wire: per-call nonce accepted the real socket; `TimeLimit=180` was on the `Calls.create` body; the callee was the only entry in `allowed_callees`.

Follow-ups filed from what the proof surfaced: #983 (inbound webhook 403 through the tunnel), #984 (fail fast on `busy`/`no-answer`), #985 (`pytest.ini` 60s timeout vs `pyproject.toml`).

https://claude.ai/code/session_01Tc6K2QfkvcWiasy8ane6EG


drewdrewthis:
## Review verdict: READY
Clean at `0522b847`.

Reviewer findings P1/P2 fixed in 9ae45ef5, c3a8efc3, 24a98990; adversarial re-review at 24a98990: clean (residual risks noted in PR body are pre-existing).

- Four-reviewer pass (hygiene, principles, security, tests) on the feature commits → 7 must-fix + 12 should-fix, all resolved in `cfde4454` (details in the PR body's Review section).
- Commits after that review: `9b4577ce` (pytest-timeout marker on the live e2e), `08776850` (provider overrides + in-script frame sampling for the live e2e), `44b85f90` (CodeQL hygiene, no behavior change), `81115be9` (export `TunnelNotReadyError` from the `scenario` root — caught by the ready_for_review CI run), `600811d8` (CodeRabbit major: read the stream nonce at the `start` frame instead of socket entry, Python + JS, with a regression test in each; plus two pyright Optional narrowings and an unused import), `a33c6256` (security reviewer + CodeRabbit: drop every non-start frame from a socket that has not been adopted — media/dtmf/stop could previously be injected without auth; Python + JS, regression test in each), `499438eb` — test(voice/twilio): bind AC11 live a-leg e2e in JavaScript (env-gated twin of the Python live test; `openTwilioTunnel` gains an explicit `cloudflared` provider), `26d206cd` — test(voice/twilio): race the tunnel DNS probes instead of waiting for all three (CodeRabbit nitpick on `499438eb`, applied).
- CodeRabbit second major (JS `TunnelReadiness` wiring) declined with reasons on the thread: opt-in by design, JS has no tunnel-owning harness.
- Independent security review of `600811d8` (arming-race fix): CLEAN for the race; surfaced the un-adopted-frame injection gap, fixed in `a33c6256`. Remaining note, not blocking: an un-adopted socket has no per-socket idle timeout and lives until `disconnect()` tears the server down.
- Live use-proof recorded in the PR body: real Twilio a-leg call `CA03274362692a43f124bf7fa190dc915f`, 30s, `1 passed`, judge `success` on all three criteria, zero child legs.
- Live re-proof at post-hardening heads: run 6 Python on `a33c6256` (`CAc042f68bf8544f50ed84abe36f8eaf1a`) and run 7 JavaScript on `499438eb` (`CAb256cfe04bb34d5334af734c40df5766`), both completed/30s/0 child legs; CI green on `499438eb` (python-complete rerun after an unrelated live-LLM example flake in `examples/test_vegetarian_recipe_agent.py`); 0 unresolved threads.
- Verification at head: Python `tests/voice -k twilio` 152 passed / 3 skipped, `pyright .` 0 errors, ruff clean on changed files; JS `vitest run src/voice/adapters/__tests__` 315 passed.

https://claude.ai/code/session_01Tc6K2QfkvcWiasy8ane6EG


aefca167: 10 CodeQL py/ineffectual-statement findings in tests fixed (`_ = await …`), no runtime code changed.


0522b847: last CodeQL ineffectual-statement nit in a span test.

javascript/src/voice/adapters/twilio-server.ts
```diff
@@ -279,14 +292,63 @@ export class TwilioWebhookServer {
...
-    adapter._resetCallState();
+
+    // A-leg WS auth (#762 guardrail (a)): `_streamNonceForServer` is set ONLY by
+    // an "a-leg" placeCall, so its presence is what arms enforcement — b-leg and
+    // inbound sockets keep today's un-gated behaviour byte for byte. Until an
+    // armed socket has authenticated, it gets NO adapter state: not `_streamWs`
+    // (which `sendAudio` writes to — adopting an unauthenticated socket would
+    // hand the attacker our outbound audio) and not the per-call queue purge
+    // (which would drop the live call's audio).
+    const expectedNonce = adapter._streamNonceForServer;
```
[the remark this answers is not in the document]
drewdrewthis: Confirmed and fixed in both languages. The nonce/enforce flag is now read inside the `start` branch when the frame is processed, and adoption happens on the first `start` frame after auth (b-leg un-gated path unchanged). Regression tests: `test_a_leg_socket_connected_before_arming_is_still_gated` (Python, `tests/voice/test_twilio_stream_auth.py`) and `AC5: a socket connected before the nonce was armed is still gated on it` (JS, `twilio-stream-auth.test.ts`). Both fail on the old snapshot code and pass on the fix. Lands in the next push on this branch.

javascript/src/voice/adapters/twilio.ts
```diff
@@ -377,6 +601,77 @@ export class TwilioAgentAdapter extends VoiceAgentAdapter {
...
+          `reachable from the edge yet, so Twilio's media stream would ` +
+          `connect to nothing. Not originating.`,
+        { cause: err },
+      );
+    }
+  }
+
   async waitForCall(timeoutMs = 120_000): Promise<void> {
     this._assertConnected();
     const rest = this._rest;
```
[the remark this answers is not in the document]
drewdrewthis: Wont fix, by design. `tunnelReadiness` is a documented opt-in option (`twilio.ts:151-156`) and the no-probe path is intentional: no probe means the caller owns a stable public URL (`twilio.ts:658-661`). JS has no tunnel-owning harness to wire it in -- `openTwilioTunnel` returns `{ url, provider, close }` and does not construct the adapter, and the ngrok/localtunnel abstraction has no edge-reachability probe. Python wires it only because `CloudflareTunnel` is itself the probe and `TwilioHarness` owns both. A JS harness + probe is a separate feature, out of scope for this PR.

python/scenario/voice/adapters/_twilio_server.py
```diff
@@ -329,6 +397,35 @@ async def _enqueue(pcm: bytes) -> None:
...
+                            # rule the terminal sentinel below already follows.
+                            with suppress(Exception):
+                                await ws.close()
+                            return
+                        if verdict == "ignore":
+                            continue
+                    # Adopt on the first start frame — after auth in a-leg mode,
+                    # always in un-gated (b-leg/inbound) mode. Idempotent: a
+                    # resend must not clobber the live call's state.
+                    if not adopted:
```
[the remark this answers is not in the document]
drewdrewthis: Confirmed and fixed. A single guard ahead of the frame branches now drops every non-`start` frame from a socket that has not been adopted (`media`, `dtmf`, `stop`), in both `_twilio_server.py` and `twilio-server.ts`. Regression tests: `test_a_leg_media_from_socket_that_never_authenticated_is_not_enqueued` (Python) and `media from a socket that never authenticated is dropped, not enqueued` (JS); both assert the inbound frame counter stays at 0 and the socket is never adopted. Lands in the next push on this branch.

sergioestebance (commented):
Two reproduced findings from the merge-readiness review. Both need addressing before merge; details and reproduction steps are inline.

python/scenario/voice/adapters/_twilio_server.py
```diff
@@ -328,7 +396,47 @@ async def _enqueue(pcm: bytes) -> None:
...
+                if frame.event != "start" and not adopted:
+                    continue
+
                 if frame.event == "start":
+                    # Read the adapter's CURRENT nonce here, when the start frame
+                    # is processed — never a snapshot from loop entry. This is
+                    # what closes the arming race: a socket that connected before
+                    # ``place_call`` armed the nonce is still gated on it the
+                    # moment its start frame arrives.
+                    enforce_auth = adapter._stream_nonce is not None
```
sergioestebance, line 442: **P1 — A socket adopted before dialing bypasses a-leg authentication (Python and JavaScript)**

A nonceless `start` received after `connect()` but before `place_call()` is adopted because the nonce is still unset. Creating the a-leg nonce does not invalidate that socket's `adopted` flag, transport ownership, or already-set stream-connected signal. Consequently, `place_call()` returns before any socket authenticates, and `send_audio()` writes to the unauthenticated socket. The same issue exists in [the JavaScript start handler](https://github.com/langwatch/scenario/blob/26d206cd700d2a451fb4e93b59f09fbd848fef53/javascript/src/voice/adapters/twilio-server.ts#L426).

Reproduced through the production session wrapper with mocked REST: connect the adapter, deliver a nonceless `start`, then originate an a-leg call and send one audio chunk. Both languages delivered one outbound frame to the unauthenticated socket. In Python, after adopting the genuine nonce-authenticated socket, the early socket also injected five media frames; its subsequent `stop` marked the stream ended and cancelled the live call's watchdog. Python signature validation was enabled in this reproduction.

**Fix direction:** bind socket authorization to the current call generation, reset connection state when dialing, and prevent stale sockets from sending/receiving audio or changing the current call's terminal state and watchdog. Add a regression where the nonceless `start` itself arrives **before** nonce creation; the existing arming-race test delays that frame until after arming.

javascript/src/voice/adapters/twilio.ts
```diff
@@ -439,6 +734,95 @@ export class TwilioAgentAdapter extends VoiceAgentAdapter {
...
+    const rest = this._rest;
+    if (rest) {
+      try {
+        await rest.endCall(callSid);
+      } catch {
+        // Best-effort: Twilio's own TimeLimit is the backstop for this backstop.
+      }
+    }
+    try {
+      this._streamWs?.close();
```
sergioestebance, line 874: **P2 — An expired watchdog can close a newer call's socket**

The generation check runs before `_onMaxDurationExpired()`, but this handler awaits `rest.endCall(callSid)` and then closes whichever socket is currently in `this._streamWs`. If another call starts while the previous hangup's HTTP response is pending, re-arming the timer does not stop the already-running handler: the old response resumes it and closes the new call's authenticated socket.

Reproduced with the existing controlled-timer seam and mocked REST: expire call A's watchdog, hold its `endCall` response, originate call B with a different SID and authenticate B's socket, then release A's response. B's socket is closed; the regression assertion expecting it to remain open fails.

**Fix direction:** carry the timer generation into the expiry handler and recheck it after the REST await. Close only the socket belonging to the expired call. Add coverage for cancellation/re-arming while the hangup request is already in flight; the existing cancellation tests only cover expiry before the handler starts.

python/scenario/voice/adapters/_twilio_server.py
drewdrewthis: Fixed in 9ae45ef (P1 commit). Socket authorization is now bound to a per-adapter call generation. connect()/place_call()/wait_for_call() bump `_call_generation` and dialing resets the per-dial connection state — drops the adopted socket, clears the stream-connected signal, drains the inbound queue, zeroes the frame counter, clears the terminal flag/reason — so a dial can no longer resolve on a stale socket's signal. The media loop captures the generation at adoption and re-checks it on every side effect; a socket adopted under an older generation is closed on its next frame and cannot enqueue audio, connect, or change the current call's terminal state/watchdog.

- Python: `python/scenario/voice/adapters/twilio.py:507` (`_reset_connection_state_for_dial`) + `python/scenario/voice/adapters/_twilio_server.py:420` (generation guard).
- JS: `javascript/src/voice/adapters/twilio.ts:1058` (`_resetConnectionStateForDial`) + `javascript/src/voice/adapters/twilio-server.ts:429` (generation guard).

Regression (your exact sequence — nonceless start BEFORE nonce creation, then dial + one chunk): `test_a_leg_socket_adopted_before_dialing_is_evicted` (Python) and `"#762 P1: a socket adopted before dialing is evicted and changes nothing"` (JS). Both fail without the fix. The pre-set-`_stream_connected` idiom is replaced by a concurrent-dial harness since place_call now blocks until a socket authenticates.

javascript/src/voice/adapters/twilio.ts
drewdrewthis: Fixed in c3a8efc (P2 commit). The timer generation is now carried into the expiry handler and re-checked AFTER the REST await. Ending the originated call SID stays unconditional (it targets the expired call), but the socket close only runs when no newer call has taken over the generation — so a hang-up whose HTTP response resolves after call B started can no longer close B's authenticated socket.

- Python: `python/scenario/voice/adapters/twilio.py:1039` (recheck in `_run_max_duration_timer`, generation captured at arm time).
- JS: `javascript/src/voice/adapters/twilio.ts:848` (recheck in `_onMaxDurationExpired`).

Regression (your controlled-timer + mocked REST sequence — expire A, hold A's endCall, start B with a different SID and authenticate its socket, release A): `"#762 P2: an expired watchdog in flight does not close a newer call's socket"` (JS, also asserts B's own watchdog still hangs up B) and `test_expired_watchdog_in_flight_does_not_close_a_newer_calls_socket` (Python). Both fail without the fix. Note: the Python twin uses a b-leg B because a second a-leg dial cancels A's asyncio task outright; b-leg bumps the generation without cancelling, reproducing the JS un-cancellable-handler path.

drewdrewthis: Follow-up in 24a98990 after a security re-review of c3a8efc3 found two remaining holes in this path:

1. (JS, critical) The guard keyed on the re-arm counter (`_maxDurationGeneration`), which only `_cancelMaxDurationTimer()` bumps. A b-leg / `waitForCall` / no-cap takeover bumps the *call* generation (`_resetConnectionStateForDial`) without arming, so it never moved the re-arm counter — a stale watchdog stayed able to close the newer call's live socket. `_onMaxDurationExpired` now gates both the ended-reason stamp and the socket close on the shared `_callGeneration` (captured at arm time), while the REST endCall of A's own SID stays unconditional.

2. (Py, medium) `_run_max_duration_timer` stamped `"max_duration"` before any generation check; since it wins ended-reason ties permanently, a stale watchdog permanently mislabelled the newer live call. The stamp is now gated on `generation == self._call_generation` before the REST await, matching the socket-close gate after it.

Added regressions in both languages: a plain b-leg takeover leaves the newer call's socket open, and a stale watchdog firing after a takeover never mutates the newer call's ended-reason (asserted independently of the socket close). Both fail without the fix; full suite green.

python/scenario/voice/adapters/_twilio_server.py
drewdrewthis: Thanks, this was real against 26d206cd. Fixed on the branch in both languages:

- 9ae45ef5 `fix(voice/twilio): bind media-socket authorization to the call generation`: `place_call()` / `wait_for_call()` (and the TS twins) now call `_reset_connection_state_for_dial()` before origination, which bumps `_call_generation`, clears the stream-connected signal, drops `_stream_ws` / `_stream_sid`, drains the queue and resets the terminal flags. The media loop captures the generation at adoption and closes a stale socket on its next frame (`_twilio_server.py` media loop, `twilio-server.ts` start handler). The wrapper's `finally` is identity-guarded (`if adapter._stream_ws is ws`), so an evicted socket cannot null the live transport.
- a33c6256 drops every frame from a socket that has not been adopted.

Regression tests, both mirroring your scenario (nonceless `start` after `connect()` and before nonce creation, then the genuine socket goes live, then media plus `stop` on the stale socket):
- Python `tests/voice/test_twilio_stream_auth.py::test_a_leg_socket_adopted_before_dialing_is_evicted`
- TypeScript `src/voice/adapters/__tests__/twilio-stream-auth.test.ts` "#762 P1: a socket adopted before dialing is evicted and changes nothing"

Re-ran both suites on 0522b847: 29 TS and 29 Python tests pass, and a throwaway reproduction of your exact sequence through `run_stream_session` shows none of the six symptoms.

javascript/src/voice/adapters/twilio.ts
drewdrewthis: Agreed, and fixed on the branch:

- c3a8efc3 `fix(voice/twilio): recheck the watchdog generation after the hang-up REST call`: `_onMaxDurationExpired` now carries the call generation into the handler and re-checks `stillOurs()` after `await rest.endCall(callSid)` before touching `this._streamWs`. The Python watchdog (`_run_max_duration_timer`) had the same await-then-close window and got the same recheck.
- 24a98990 `fix(voice/twilio): key the watchdog guard on the call generation, not the re-arm counter`, so a re-armed timer for the same call cannot be confused with a new call.

Regression tests:
- TypeScript `src/voice/adapters/__tests__/twilio-call-duration.test.ts` "#762 P2: an expired watchdog in flight does not close a newer call's socket" plus the re-arm follow-ups
- Python `tests/voice/test_twilio_call_duration.py::test_expired_watchdog_in_flight_does_not_close_a_newer_calls_socket`

Both suites pass on 0522b847.

PR #990, opened 2026-09-13
fix(voice): gate Twilio inbound audio on speech so tail-silence can end the callee's turn

drewdrewthis:
## Problem

On a real Twilio a-leg call the simulated caller speaks exactly once. The callee (an ElevenLabs agent) answers, then is heard asking "hello? are you still there? hello?" for the rest of the call while the user simulator never gets its second turn. Reproduced on both a dev call and a production call (langwatch/langwatch#8014 follow-up); traces show every `voice.audio.receive` span terminated by `terminal_chunk` (hang-up) or `hard_ceiling` (60 s), never `tail_silence`.

## Root cause

Twilio Media Streams deliver a `media` frame every 20 ms for the whole life of the call, silence included, and the protocol carries no speech/silence/VAD signal (`mark` is only a playback-completion echo; the endpointing knobs belong to ConversationRelay). `TwilioWebhookServer.mediaStreamLoop` batches those frames every 100 ms and enqueues them unconditionally. `drainAgentResponse` ends a turn on an ARRIVAL GAP of `responseTailSilence` seconds — a gap that can never occur on a phone line, so the callee's turn only ends when the callee hangs up or the drain's hard ceiling fires.

## Fix

A `TwilioSpeechGate` between the media loop and the inbound queue (`voice/adapters/twilio-speech-gate.ts`), on by default:

- each 100 ms chunk's RMS is measured on the int16 scale the SDK already produces; chunks below `rmsThreshold` are dropped so silence becomes a real gap;
- speech is admitted with a `hangoverMs` so plosives/breaths inside an utterance do not split it;
- a bounded `prerollMs` of the quiet before an onset is replayed so the first syllable is not clipped;
- the empty terminal sentinel always passes through.

Defaults (800 / 400 ms / 300 ms) come from measuring a real call against an ElevenLabs callee: between-utterance noise p95 ≈ 740, speech p10 ≈ 940, p50 ≈ 1380 (int16 RMS over 100 ms windows). `twilioAgent({ speechGate })` tunes them; `speechGate: false` restores the pre-fix pass-through. Per-call state resets wherever the inbound queue's buffered audio is cleared; lifetime counters (`voice.twilio.speech_gate.{enabled,dropped_chunks,dropped_ms,onsets}`) are stamped on the `voice.adapter.disconnect` span next to the frame counters. Exports on the `voice` barrel.

No change to the shared drain or to the other transports.

## Tests

- `twilio-speech-gate.test.ts` — the pure admit/drop decision (noise dropped, onset + pre-roll, hangover, sentinel pass-through, resets).
- `twilio-speech-gate-turn-end.test.ts` — drives the REAL `/twilio/stream` route with a real `ws` client and the REAL `adapter.call()` drain while silent frames keep flowing at wall-clock pace: with the gate the turn ends on tail silence within hangover + tail silence; with `speechGate: false` the same line provably never reaches tail silence until frames stop; a second turn gets its own onset.
- Existing suites whose µ-law fixtures are constant bytes (which decode to near-silence) now construct the adapter with `speechGate: false` — they test queue/sentinel/span mechanics, not turn-taking.

`vitest run src/voice`, `tsc --noEmit` (one pre-existing error in `twilio-external-upgrade.test.ts`, untouched here), `eslint` on the touched files.

## Python parity

`python/scenario/voice/adapters/_twilio_server.py` has the same unconditional enqueue; tracked in the follow-up issue linked below.

Python parity issue: #991


drewdrewthis:
Folded into #989 (fast-forwarded its branch to include 9e301e2a); delivering as one PR.

drewdrewthis:
Folded into #989 (same commits, now on that branch) so the two fixes land as one PR.

PR #1000, opened 2026-09-17
fix(python): consolidate pytest config into pyproject.toml (#985)

### LINKED ISSUE #985: python: pytest.ini sets a global 60s timeout that pyproject.toml says does not exist
`python/pyproject.toml` (around lines 106-110) documents that no global pytest timeout is configured and that long-running tests set their own. `python/pytest.ini` however carries:

```
timeout = 60
timeout_method = thread
```

and `pytest.ini` wins over `pyproject.toml`, so every test is capped at 60s with the thread method — which kills the test without running fixture teardown. For the live Twilio e2e tests that means the adapter's own guards (120s stream-connect timeout, 180s max call duration) never fire and the hangup in teardown is skipped; the call is left to Twilio's `TimeLimit`.

`9b4577ce` on PR #982 works around it for `test_twilio_a_leg_external_e2e.py` with `@pytest.mark.timeout(600)`. Proper fix: pick one source of truth (probably move the timeout config into `pyproject.toml`, use `timeout_method = signal` or a per-marker override for live tests), and fix the comment.

https://claude.ai/code/session_01Tc6K2QfkvcWiasy8ane6EG

longxing-alt:
Fixes #985.

## What changed

`python/pytest.ini` and `python/pyproject.toml` both configured pytest, and `pytest.ini` wins. pytest says so on every run:

```text
configfile: pytest.ini (WARNING: ignoring pytest config in pyproject.toml!)
```

So `[tool.pytest.ini_options]` in `pyproject.toml` was dead config, and `pytest.ini`'s `markers` list **replaced** pyproject's rather than adding to it. `pyproject.toml` is now the single source of truth: `pytest.ini` is deleted and its settings move across, with all four markers in one list.

## Why

Two consequences of the shadowing, both observable today:

**1. The comment described behaviour that did not exist.** pyproject.toml said "Per-test timeout deliberately NOT set in ini_options". In fact a 60s global cap with `timeout_method = thread` was in force — the method that kills the whole pytest process instead of the one slow test.

**2. `agent_test` was never registered.** Because `pytest.ini` replaced the marker list:

```text
$ uv run pytest examples/ tests/ --collect-only -q     # before
24 x PytestUnknownMarkWarning: Unknown pytest.mark.agent_test - is this a typo?
1627 tests collected
```

Under `-W error::pytest.PytestUnknownMarkWarning` — a reasonable strictness setting for CI — collection fails outright:

```text
E   pytest.PytestUnknownMarkWarning: Unknown pytest.mark.agent_test - is this a typo?
ERROR examples/test_false_assumptions.py - pytest.PytestUnknownMarkWarning
!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
```

## Behaviour is intentionally unchanged

The 60s cap is load-bearing — the unit suite depends on it and TESTING.md, `python-ci.yml` and `voice-integration.yml` all document it — so it is preserved, not removed. I made `pyproject.toml` authoritative and moved the settings verbatim rather than dropping them.

Reading the resolved config via a `pytest_configure` hook, before → after:

| | before | after |
|---|---|---|
| `inifile` | `pytest.ini` | `pyproject.toml` |
| `timeout` | `60` | `60` |
| `timeout_method` | `thread` | `thread` |
| `asyncio_default_fixture_loop_scope` | `function` | `function` |
| markers registered | 9 | 10 (`+agent_test`) |

## Docs updated

The places that documented the conflict now describe the fixed state, and the still-true operational guidance (give long demos an explicit `--timeout`) is kept:

- `TESTING.md` — "Running the multi-turn demos"
- the stale `pytest-timeout` dependency comment in `pyproject.toml` ("global suite has no limit" was wrong)
- the `pytest.ini` reference in `tests/voice/test_twilio_a_leg_external_e2e.py`

## Verification

```bash
# unit suite, CI-equivalent invocation
uv run pytest tests/ -q -m "not integration"
#   before: 3 failed, 1546 passed, 10 skipped
#   after:  3 failed, 1546 passed, 10 skipped

# collection warnings
uv run pytest examples/ tests/ --collect-only -q
#   PytestUnknownMarkWarning: 24 before -> 0 after; 1627 collected both times

uv run ruff check tests/voice/test_twilio_a_leg_external_e2e.py   # clean
uv run ruff format --check tests/voice/test_twilio_a_leg_external_e2e.py   # already formatted
```

The 3 failures are identical before and after and are unrelated to this change (2 in `tests/test_browser_tab_handoff.py`, 1 in `tests/voice/test_voice_spans_gemini.py`); they reproduce on an untouched `main` checkout.

## Not included

I also wanted to fix the two workflow comments in `.github/workflows/python-ci.yml` and `voice-integration.yml` that name `pytest.ini`, but my token lacks the `workflow` scope so GitHub refuses the push. They are comments only (no behaviour change) — happy to fold them in if a maintainer prefers, otherwise they can be picked up separately.


langwatch-agent (commented):
No actionable findings. Independently parsed current main’s pytest.ini and this head’s TOML as data: all three effective settings and all existing marker definitions are preserved, with agent_test added to the effective marker list. Searched workflow invocations for reliance on the removed config path. No external PR code, pytest plugins, tests, or package scripts executed; visible checks report no failed or pending checks. Residual limitation: runtime pytest discovery was not independently exercised.

LangWatch-Review: verdict=clean sha=56ba717469f0c193fb3fca21fa7290cbbbe6d890 p0=0 p1=0 p2=0 p3=0

longxing-alt:
Following up on @coderabbitai's inline note about the two workflow comments. You're right, and I couldn't include them: my push was rejected with

```
refusing to allow a Personal Access Token to create or update workflow
`.github/workflows/python-ci.yml` without `workflow` scope
```

so the PR is limited to the four non-workflow files. Here are the three comment-only edits, ready to apply. Nothing else in either workflow changes — same commands, same timeouts, same behaviour.

### `.github/workflows/python-ci.yml` (~line 178)

```diff
       - name: Test (Examples)
         if: steps.secrets.outputs.available == 'true'
         # Examples hit real LLMs over the network; they can legitimately take
-        # well over the 60s per-test timeout that pytest.ini sets for the unit
-        # suite. Override to 300s here so slow-but-correct runs don't get
+        # well over the 60s per-test timeout that pyproject.toml sets for the
+        # unit suite. Override to 300s here so slow-but-correct runs don't get
         # pytest-timeout'd.
```

### `.github/workflows/voice-integration.yml` (~line 180)

```diff
-        # `--timeout=540` is NOT optional (issue #491). pytest.ini sets `timeout = 60`
-        # and WINS over pyproject.toml, and `timeout_method = thread` kills the ENTIRE
+        # `--timeout=540` is NOT optional (issue #491). pyproject.toml sets
+        # `timeout = 60` and `timeout_method = thread` kills the ENTIRE
         # pytest process — not just the slow test. In THIS workflow LANGWATCH_API_KEY is
```

### `.github/workflows/voice-integration.yml` (~line 292)

```diff
             # TWO nested caps, and `--timeout=540` is NOT optional:
             #
-            # pytest.ini sets `timeout = 60` for the unit suite, and pytest.ini
-            # WINS over pyproject.toml — pytest says so out loud: "configfile:
-            # pytest.ini (WARNING: ignoring pytest config in pyproject.toml!)".
-            # So pyproject's "@e2e voice tests can legitimately take 5+ min /
-            # per-test timeout deliberately NOT set" note is DEAD CONFIG, and
-            # without an override these demos inherit the 60s cap. They blow
-            # through it structurally — long_hold alone does a 15s wall-clock
+            # pyproject.toml sets `timeout = 60` for the unit suite, and there is
+            # no pytest.ini shadowing it (issue #985), so it is the only source
+            # of that cap. Without an override these demos inherit the 60s cap.
+            # They blow through it structurally — long_hold alone does a 15s
+            # wall-clock
             # scenario.sleep(15) before 8 live LLM+TTS+STT turns — so un-skipping
             # them under a 60s cap would just trade "asserted but never run" for
             # "shot at 60s, red". Same idiom as python-ci's Examples step, which
             # overrides to --timeout=300 for exactly this reason.
```

Note the third hunk's `scenario.sleep(15)` line: I left an awkward short line to keep the diff minimal rather than re-wrap the whole paragraph.

After these three hunks, `grep -rn "pytest\.ini" .github/ TESTING.md python/` is clean — `python/pytest.ini` is gone and nothing but the `[tool.pytest.ini_options]` table name references it.

Happy to push these myself if someone grants a token with `workflow` scope, or a maintainer can apply them directly. Either way it's cosmetic; the functional change is entirely in this PR.

